PR #21, opened 2016-10-28
use Net(bindIp="localhost", ...) to avoid windows firewall popup

julianladisch:
The current code binds the embedded mongo to all IP addresses. This triggers windows firewall popups.
To have the popup only one time UserTempNaming() has been used.

The better solution is to restrict bindIp to localhost.


julianladisch:
This fixes the problem of windows users with locked extract-username-extractmongod.exe in the temp directory.

PR #26, opened 2016-11-10
fix .properties files

julianladisch:
maven: copy .properties using ISO-8859-1
use standard file name schema (APIMessages_en.properties) to help translation toolchain and trigger IDEs to encode them correctly
accept old file name schema during transition time
add German translation


dcrossleyau:
See [DMOD-97](https://issues.folio.org/browse/DMOD-97).

PR #31, opened 2017-01-05
maven-enforcer-plugin: maven 3.3, JDK 1.8.0-111

julianladisch:
Stop build with helpful message if maven or jdk version is too old. See discussion in development channel in slack-folio.


dcrossleyau:
I think that that was a typo "1.8.0-111" as the documentation about the "Let's Encrypt certificate" issue says "1.8.0-101". I have now set it to that in commit b9deff7.

PR #35, opened 2017-01-06
DMOD-141: clean generated .java; use target/generated-sources/

julianladisch:
Might also resolve DMOD-138 (auto-generated code should have a preamble saying so)

GenerateRunner.java cleans the directory before writing the generated sources. ClientGenerator.java doesn't. This code is moved into makeCleanDir(dir) and invoked from both. This fixes DMOD-141.

In addition the output directory for generated sources if changed from a subdirectory of src/ to target//generated-sources/jaxrs. This way a "maven clean" can also clean it. It also makes it obvious for every developer that it is generated code and should not be edited, see DMOD-138.


julianladisch:
commit 16c8a04bc54120ef2ee854fa4d0f463739aa58aa uses makeCleanDir(dir) from this pull request.

Using a separate directory like generated-sources has been dropped.

PR #44, opened 2017-02-09
DMOD-161: BinaryOutStream unit tests

julianladisch:


shale99:
tested and verified on rmb v7

PR #51, opened 2017-03-03
DMOD-170 replace PomReader by ApplicationProperties

julianladisch:
This also merges git.properties into application.properties.


julianladisch:
postponed

PR #52, opened 2017-03-03
DMOD-165 fix db. DB_ environment variables

julianladisch:
Bash doesn't allow environment variables like db.host that contain a dot. Therefore replacing dot by underscore. Converting to the usual uppercase.

allDBConfs() also returns the values from deprecated variables with dot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,14 +166,14 @@ To run the configuration module, navigate to the `/target/` directory in the con
 
 RMB implementing modules expect a set of environment variables to be passed in at module startup. The environment variables expected by RMB modules are:
 
- - db.host 
- - db.port
- - db.username 
- - db.password 
- - db.database 
- - db.queryTimeout
- - db.charset 
- - db.maxPoolSize
+ - DB_HOST
+ - DB_PORT
+ - DB_USERNAME
+ - DB_PASSWORD
+ - DB_DATABASE
+ - DB_QUERYTIMEOUT
+ - DB_CHARSET
+ - DB_MAXPOOLSIZE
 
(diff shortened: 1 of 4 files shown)


shale99:
tested locally

marc-johnson:
It may be a little late for this change, but Bash can executing a command with environment variables that contain dots, for example:

env db.host=localhost \
    db.port=5432 \
    db.database=test \
    db.username=inventory_storage_admin \
    db.password=admin \
    mvn test

julianladisch:
@marc-johnson Bash cannot (easlily) access an environment variable whose name contains a dot:

```bash
> env db.host=localhost   echo $db.host 
.host
> env db.host=localhost   echo ${db.host}
bash: ${db.host}: bad substitution
```

The definition of a Bash `name`: "A `word` consisting solely of letters, numbers, and underscores, and beginning with a letter or underscore. `Name`s are used as shell variable and function names. Also referred to as an `identifier`." Reference: https://www.gnu.org/software/bash/manual/html_node/Definitions.html
This does not contain a dot.

dcrossleyau:
On macOS bash-4.4.12 ...
```
export db.port=5432
-bash: export: `db.port=5432': not a valid identifier
```

marc-johnson:
@julianladisch @dcrossleyau 

I hadn't tried exporting or referring to environment variables in the shell. My goal with the script above was to replicate the behaviour of Okapi when it starts a new process for a module, I didn't have a need to manipulate the environment variables themselves (I was glad they were self contained).

Apologies for the confusion :-(

PR #60, opened 2017-03-22
replace CQLWrapper's IllegalStateException by CQLQueryValidationException

julianladisch:
This is compatible because CQLQueryValidationException extends IllegalStateException

https://issues.folio.org/browse/DMOD-218


shale99:
Julian, can you bump the version by 0.0.1? not sure if other modules may be catching this exception

julianladisch:
Other modules that catch IllegalStateException continue to work because CQLQueryValidationException extends IllegalStateException.

PR #81, opened 2017-05-29
provide UtilityClassTester to other modules

julianladisch:


dcrossleyau:
@julianladisch : I am having trouble building this. I think that it needs "junit" dependency in domain-models-runtime/pom.xml but i am not mvn enough to know "scope".

PR #82, opened 2017-05-30
Use locale C for embedded postgres

adamdickmeiss:
Use locale C on Unix. It should be a dummy sort, byte-wise, and thus give same results everywhere.


julianladisch:
Builds locally and looks good, can be merged.

FOLIO Jenkins Docker Host (localhost)-fbe1a1ab19d4 is offline, the failure report can be ignored.

shale99:
i prefer not to merge this - see jira issue RMB-27

PR #85, opened 2017-06-07
FOLIO-573 RMB-30 filename schema key

dcrossleyau:
Ready to merge.

Use relative pathname as schema key name in schemas and in RAMLs.

Move the raml-util git submodule to beside the rails directory.


dcrossleyau:
Needed to revert that. See comments in RMB-30.

PR #95, opened 2017-06-28
FOLIO-693 HTTP status codes: HttpStatus.java

julianladisch:


shale99:
the rmb uses the vertx which has its own http status enum (io.vertx.core.http.HttpMethod). do we need another?

julianladisch:
@shale99: During the team1 roundup on Monday we decided to add such documentation and enum, see https://issues.folio.org/browse/FOLIO-693 . Please add your concerns to the jira issue FOLIO-693 and explain in more detail.

PR #114, opened 2017-10-24
 RMB-72 Skip body parsing if not necessary, add error handling

sassrobi:


shale99:
tested locally, merging this, will need to add tests at one point

PR #126, opened 2017-12-01
RMB-103 Also dereference raml-util directory

julianladisch:


julianladisch:
I've added msg.raml with x/y/msgs.schema

PR #151, opened 2018-01-11
RMB-113: Ignore missing /ramls/ dir

julianladisch:


shale99:
hi julian, so this adds another property to the maven that the developer needs to declare if they want the ramls in the non default location? why not just use the ramls_file base directory - it has to be relative so it most certainly will be a ${basedir}/my_raml_dir
thoughts?

julianladisch:
This does not require any additional property.
maven automatically creates the `maven.multiModuleProjectDirectory` property when running in a submodule. This is needed for https://github.com/folio-org/mod-configuration where the submodules mod-configuration-client and mod-configuration-server use the ramls dir of the main module.

julianladisch:
Maven's source code for maven.multiModuleProjectDirectory:
https://github.com/apache/maven/search?utf8=%E2%9C%93&q=multiModuleProjectDirectory&type=

shale99:
ok, so this is handling a specific use case , we are still requiring a raml declaration in the pom and on the other hand still forcing the declaration to be in a specific place. i think in such a case - we have two options. 
1. remove the raml file declaration from the pom and indicate it has to be in the ramls or raml-util directories 
2. allow the raml declaration in the pom to decide everything
right now it is half this and half that. 
i will merge but we need to revisit this

PR #183, opened 2018-03-02
RMB-143: Version update: vertx 3.5.1, jackson 2.9.4

julianladisch:
Use vertx-stack-depchain.
mod-users and mod-inventory-storage build with this change.
mod-configuration build with this change if you update vertx-unit to 3.5.1.


shale99:
i can confirm this,
this is potentially a breaking change for other modules, maybe bump the major

PR #198, opened 2018-05-17
runOnContext for SQL queries RMB-166

adamdickmeiss:


shale99:
the change makes sense - i am interested to see that both without and with the change all code was indeed running on the same vertx event loop thread and that only context was lost. i believe that is the case so that this change does not have a negative threading impact in any way

shale99:
tested on mod-config, mod-users, mod-inventory, all tests pass, merging

PR #199, opened 2018-05-21
runOnContext for SQL queries RMB-166

shale99:


shale99:
checked with mod-users, mod-config, mod-inventory-storage

PR #215, opened 2018-07-24
 RMB-194: Run PostgresClientIT.parallel() when not using embedded mode

julianladisch:


adamdickmeiss:
Why do you want to run parallel tests? Has there an issue with doing things in parallel? Looks like this test and many others resly on postgres-runner rather than an on embed_postgres .. And in particular embed_postgres is never true in this test anyway.. So overly compilcated IMHO.

julianladisch:
The tests runs if a running postgres server is configured using DB_ env variables.

This unit test checks that our async works properly and really runs SQL queries in parallel.

adamdickmeiss:
Closing this as embedded postgres is not in use.

PR #233, opened 2018-08-17
RMB-211: FolioLog4j2LogDelegate with error(e) stacktrace logging

julianladisch:


marcjohnson-kint:
@julianladisch Will this be enabled by default in future versions of RAML Module Builder, or will modules need to configure it?

julianladisch:
@marcjohnson-kint I've created https://issues.folio.org/browse/RMB-212 to discuss this question.

util/src/main/java/io/vertx/core/logging/FolioLog4j2LogDelegate.java
```diff
@@ -0,0 +1,81 @@
...
+ * and stacktrace.
+ * <p>
+ * This is in contrast to Log4j2LogDelegate where a logger prints the message only, no stacktrace.
+ * <p>
+ * Example for an Exception e: logger.error(e) is the same as logger.error(e.getMessage, e).
+ * Use logger.error(e.toString()) or logger.error(e.getMessage()) to suppress the stacktrace.
+ */
+public class FolioLog4j2LogDelegate extends Log4j2LogDelegate {
+  public FolioLog4j2LogDelegate(final String name) {
+    // the super constructor is package protected, therefore
```
aclseelen: You probably mean "package private". "Protected" wouldn't be a problem. To me, the choice for a package private constructor suggests that extending the class by developers other than vertx developers is not intended. However, I wouldn't know why. I don't consider using another organization's package name an ideal solution (https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html). Might there be a better workaround? Perhaps without extending and overriding, by just calling the vertx logger from within a new method? Just an idea.
```
private static final Log4jLogDelegate LOGGER = Log4jLogDelegateFactory.createDelegate();

public void error(final Object o) {
  if (o instanceof Throwable) {
    Throwable t = (Throwable) o;
    LOGGER.error(t.getMessage(), t)
  } else {
    LOGGER.fatal(o)
  }
}
```
Or passing this class the original vertx logger with the constructor, especially when using this class is optional, as will be discussed in RMB-212, you'll need a normal logger anyways.

marcjohnson-kint: @aclseelen The approach you describe sounds like a similar approach to the LogWriter class I speculatively introduced for CIRCSTORE-26 (see https://github.com/folio-org/mod-circulation-storage/pull/96) and then dropped for this change.

aclseelen: You mean because the original logger should be hidden to programmers in case the Folio logger is configured?

marcjohnson-kint: @aclseelen Is that question for me? If so, I don't think I understand.

The approach I had taken, prior to awareness of this change by @julianladisch was to introduce a class that used the existing configured logger in the expected way, so it was expressed specifically and in a single place.

As @julianladisch pointed out in that conversation, a trade-off with that approach is that the module developer needs to know to use another class in unison with the logger.

aclseelen: Yes, I meant you, but think I see now. You mean that extending is necessary, since the user doesn't need to know about Julian's subclass when using polymorphism?
```
TheLoggerEveryoneKnows logger = new TheLoggerNobodyKnowsThatExtendsTheOneEveryoneKnows();
```

marcjohnson-kint: I think there are two different approaches to achieved the log output we want:
* Replace the logger that modules are aware of with one that logs exceptions how we expect (by registering this custom delegate behind the scenes)
* Use a class that is performs the expected behaviour, and uses the configured logger as a collaborator

The former is transparent to the module developer once configured, and applies in all situations. The latter requires the module developer to choose to use the additional class.

My understanding of your original message is that you were expressing concerns around the former option and an idea similar to the latter option. Have I understood your message correctly?

aclseelen: @marcjohnson-kint Thanks for the explanation. Yes, you have understood me almost correctly. I agree that we would prefer something "behind the scenes" . To what I understand from what you were saying is that this polymorphism (referring to an adapted subclass) is necessary for this approach. However, I don't like this approach, since this vertx logger class was obviously not intended to be extended outside of their own developer team. Therefore, I was thinking about new ways to log the preferred behavior by fixing it behind the scenes. Does that make sense to you?

aclseelen: Another solution I see there is to ask why the vertx devs made this constructor package private, the file an issue there, hoping they will change it to 'protected' in their next version

julianladisch:
> Use package name "org.folio" instead of "org.vertx".

The package name `io.vertx.core.logging` is required, see the comment in the source code:

    public class FolioLog4j2LogDelegate extends Log4j2LogDelegate {
      public FolioLog4j2LogDelegate(final String name) {
        // the super constructor is package private, therefore
        // FolioLog4j2LogDelegate must be in package io.vertx.core.logging
        super(name);
      }

PR #246, opened 2018-08-27
Added support for pattern and not null in arrays

jeremythuff:
This addresses addresses [RMB-188](https://issues.folio.org/browse/RMB-188) and touches on elements of [RMB-205](https://issues.folio.org/browse/RMB-205)

This PR introduces two custom validation annotations, `@NoNullElements` and `@ElementsPattern` as well as validator tooling to handle these annotations. This is implemented in the generated code through the registration of a custom annotator, `CustomArrayAnnotator` with jsonschema2pojo.

This will prohibit null from being added to any array.

It will also allow for the following entry in jsonschema to result in a pattern constraint against each array element:

```
"myArrayOfIds": {
      "type": "array",
      "items": {
        "type": "string",
        "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
      }
    }
```


domain-models-api-aspects/src/main/java/org/folio/rest/validators/mapper/CustomValidationExceptionMapper.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Provider
+public class CustomValidationExceptionMapper implements ExceptionMapper<ValidationException> {
+
+  private static final Logger log = Logger.getLogger(CustomValidationExceptionMapper.class);
+
+  @Override
+  public Response toResponse(ValidationException exception) {
+    log.info("Handling validation exception: " + exception.getMessage());
+    log.error("Validation Exception", exception.getCause());
```
julianladisch: If loglevel is error then only the stacktrace but not exception.getMessage() gets logged. Why do you want to suppress the message?

Can you join the two log.* lines into a single log.error(...) invokation?

domain-models-api-aspects/src/main/java/org/folio/rest/validators/CollectionValidator.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Annotation;
+import java.util.Collection;
+import java.util.Optional;
+
+import javax.validation.ConstraintValidator;
+
+public interface CollectionValidator<A extends Annotation, E, C extends Collection<E>>
+    extends ConstraintValidator<A, C> {
+
+  public boolean isEntityValid(Optional<E> element);
```
julianladisch: What is "entity"? Or can you rename "isEntityValid" to "isValid"?

domain-models-api-aspects/src/main/java/org/folio/rest/validators/AbstractCollectionValidator.java
```diff
@@ -0,0 +1,30 @@
...
+
+import javax.validation.ConstraintValidatorContext;
+
+public abstract class AbstractCollectionValidator<A extends Annotation, E, C extends Collection<E>>
+    implements CollectionValidator<A, E, C> {
+
+  @Override
+  public boolean isValid(final C collection, final ConstraintValidatorContext context) {
+    Optional<C> items = Optional.ofNullable(collection);
+    boolean valid = true;
```
julianladisch: This variable is not needed. "return false" in the loop and "return true" at the end.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/annotator/CustomArrayAnnotator.java
```diff
@@ -0,0 +1,32 @@
...
+      Optional<JsonNode> itemsNode = Optional.ofNullable(propertyNode.get("items"));
+      if (itemsNode.isPresent()) {
+        Optional<JsonNode> itemsTypeNode = Optional.ofNullable(itemsNode.get().get("type"));
+        Optional<JsonNode> patternNode = Optional.ofNullable(itemsNode.get().get("pattern"));
+        if (itemsTypeNode.isPresent() && itemsTypeNode.get().asText().equals("string") && patternNode.isPresent()) {
+          field.annotate(ElementsPattern.class).param("regexp", patternNode.get().asText());
+        }
+      }
+    }
+  }
```
julianladisch: Doing https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html , dropping Optional and moving the null test into a function gives the code a different look (and more lines): 
```
  private String text(JsonNode node) {
    if (node == null) {
      return null;
    }
    return node.asText();
  }

  @Override
  public void propertyField(JFieldVar field, JDefinedClass clazz, String propertyName, JsonNode propertyNode) {
    super.propertyField(field, clazz, propertyName, propertyNode);
    JsonNode typeNode = propertyNode.get("type");
    if (! "array".equals(text(typeNode))) {
      return;
    }
    field.annotate(NoNullElements.class);
    JsonNode itemsNode = propertyNode.get("items");
    if (itemsNode == null) {
      return;
    }
    JsonNode itemsTypeNode = itemsNode.get("type");
    if (! "string".equals(text(itemsTypeNode))) {
      return;
    }
    String pattern = text(itemsNode.get("pattern"));
    if (pattern == null) {
      return;
    }
    field.annotate(ElementsPattern.class).param("regexp", pattern);
  }
```

domain-models-api-aspects/src/main/java/org/folio/rest/validators/CollectionValidator.java
wwelling: Probably should be isElementValid. The method is being invoked within an iteration upon the elements of an array. Would not want to confuse the validation of the entire collection with the validation of a single element within the collection.

domain-models-api-aspects/src/main/java/org/folio/rest/validators/AbstractCollectionValidator.java
wwelling: http://www.cs.utexas.edu/users/EWD/ewd02xx/EWD215.PDF

wwelling:
Agreed that Optional should not be a method argument. Makes since not to force a method to perform a conditional on an argument. Similar to null checking every argument is not desirable. Perform isPresent in conditional before calling method isEntityValid will potentially save a method call as well. 

More though could be put into the propertyField. It will most likely become more complex with potential checks for other properties. One would be to affor opt out or opt in of the disallowing of nulls in arrays. 

I see the card was to disallow nulls in the arrays, not in the code.

domain-models-api-aspects/src/main/java/org/folio/rest/validators/mapper/CustomValidationExceptionMapper.java
julianladisch: Example invokation:
```
    Exception exception = new RuntimeException("Some detail.");
    log.info("Handling validation exception: " + exception.getMessage());
    log.error("Validation Exception", exception.getCause());
```
Output with loglevel set to info:
```
28 Aug 2018 11:57:25:238 INFO  Log4jTest  Handling validation exception: Some detail.
28 Aug 2018 11:57:25:251 ERROR Log4jTest  Validation Exception
```
Output with loglevel set to warn:
```
28 Aug 2018 11:53:27:354 ERROR Log4jTest  Validation Exception
```
The string "Some detail" is missing from the log.

wwelling: Thanks for the example. To be honest, it was a cut and paste. I would of guessed the cause would include the message. Anyhow, the entire class is not needed for this feature.

jeremythuff:
I have made some changes which make the disallowing of null elements in arrays an opt in feature. If an array needs to disallow null elements it will need to have the following entry in the jsonschema:

 ```
"myArrayOfIds": {
      "type": "array",
      "items": {
        "type": "string",
        "not": {
            "type": "null"
         }
      }
    }
```

File: domain-models-interface-extensions/src/main/java/org/folio/rest/tools/annotator/CustomArrayAnnotator.java [hunk 5]
```diff
@@ -0,0 +1,65 @@
...
+    if(isArray(propertyNode)) {
+      if(isItemsNotNull(propertyNode)) {
+        field.annotate(NoNullElements.class);
+      }
+      Optional<String> pattern = getPattern(propertyNode);
+      if(pattern.isPresent()) {
+        field.annotate(ElementsPattern.class).param(REGEXP, pattern.get());
+      }
+    }
+  }
```
julianladisch: Moving code into the 3 private functions is a good idea!

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/annotator/CustomArrayAnnotator.java
```diff
@@ -0,0 +1,32 @@
+package org.folio.rest.tools.annotator;
+
+import java.util.Optional;
+
+import org.folio.rest.annotations.ElementsPattern;
+import org.folio.rest.annotations.NoNullElements;
+import org.jsonschema2pojo.AbstractAnnotator;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.sun.codemodel.JDefinedClass;
```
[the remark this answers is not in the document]
julianladisch: Append `//NOSONAR` to suppress this warning (false positive).

PR #258, opened 2018-09-02
RMB-206: Add PostgresClient.getById returning POJOs

julianladisch:


marcjohnson-kint:
@julianladisch Thanks for adding this. I've given it a brief look. The signatures look like what I expected and I like the consolidation of implementation by extracting a method that takes a function.

PR #273, opened 2018-09-16
RMB-250: cql2pgjson 2.2.3: Fix fulltext search

julianladisch:
Update to cql2pg-json 2.2.3 that contains these fixes:

* CQLPG-55: Trim trailing space and loose * that breaks fulltext search.
* CQLPG-54: Use the 'simple' dictionary for fulltext instead of 'english', to get around stopwords.


julianladisch:
As discussed on FOLIO's #release-q3-2018 slack channel.

PR #274, opened 2018-09-18
RMB-251: Add unit test emptyCqlQueryValidationException()

julianladisch:


julianladisch:
The unit test proves that both cql2pgjson and CQLWrapper correctly report an empty CQL query.
https://issues.folio.org/browse/RMB-251

PR #277, opened 2018-09-20
 RMB-197: Do not start embedded postgres if PG env variables are set

julianladisch:


heikkilevanto:
I didn't go through every line, but the approach seems reasonable to me. As long as the command line option takes precedence, so we can use that in unit tests, I am happy

PR #305, opened 2018-10-16
Proper URI for generated refs; RMB-275

adamdickmeiss:


domain-models-interface-extensions/src/main/java/org/folio/rest/tools/utils/SchemaDereferencer.java
```diff
@@ -56,7 +58,12 @@ private void fixupRef(Path path, JsonObject jsonObject)
     String file = jsonObject.getString("$ref");
     if (file != null && !hasUriScheme(file)) {
       Path nPath = path.resolveSibling(file);
-      jsonObject.put("$ref", "file:" + nPath.toAbsolutePath().normalize());
+      try {
+      URI u = new URI("file", nPath.toAbsolutePath().normalize().toString(), null);
+      jsonObject.put("$ref", u.toString());
```
julianladisch: Fix indentation

PR #312, opened 2018-10-23
Fixed "langauge" typo

JccoFresh:
This is a common misspelling of language.


dcrossleyau:
Thanks for noticing that and providing the patch.

Could you please follow through on the CLA signing.

dcrossleyau:
@julianladisch Could fixing this have any flow-on effect?

PR #315, opened 2018-11-05
RMB-291: Exclude android, fix jzlib compression bug

julianladisch:


julianladisch:
mod-password-validator works for me with this fix backported to v19.4.1 and serverOptions.setCompressionSupported(true)

PR #324, opened 2018-11-12
RAML and JSON schema multiple interface

wwelling:
Implementation for jsonSchema and raml interface.

https://issues.folio.org/browse/RMB-253


domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
```diff
@@ -0,0 +1,168 @@
...
+    });
+  }
+
+  private List<String> getSchemas() throws IOException {
+    List<String> schemas = new ArrayList<>();
+    File jarFile = new File(getClass().getProtectionDomain().getCodeSource().getLocation().getPath());
+    JarFile jar = new JarFile(jarFile);
+    Enumeration<JarEntry> entries = jar.entries();
+    while(entries.hasMoreElements()) {
+      JarEntry entry = entries.nextElement();
```
julianladisch: This is more easy to understand:

    List<JarEntry> entries = Collections.list(jar.entries());
    for (JarEntry entry : entries) {

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
```diff
@@ -0,0 +1,168 @@
...
+    JsonNode schemaNode = null;
+    File jarFile = new File(getClass().getProtectionDomain().getCodeSource().getLocation().getPath());
+    JarFile jar = new JarFile(jarFile);
+    Enumeration<JarEntry> entries = jar.entries();
+    while(entries.hasMoreElements()) {
+      JarEntry entry = entries.nextElement();
+      String entryName = entry.getName();
+      if (entryName.startsWith("ramls/")) {
+        String schemaName = entryName.substring(entryName.lastIndexOf("/") + 1);
+        if (schemaName.equals(name)) {
```
julianladisch: How do you distinguish ramls/instance.json and ramls/raml-util/schemas/codex/instance.json in https://github.com/folio-org/mod-inventory-storage/tree/master/ramls ?

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
```diff
@@ -0,0 +1,168 @@
...
+    List<String> schemas = new ArrayList<>();
+    File jarFile = new File(getClass().getProtectionDomain().getCodeSource().getLocation().getPath());
+    JarFile jar = new JarFile(jarFile);
+    Enumeration<JarEntry> entries = jar.entries();
+    while(entries.hasMoreElements()) {
+      JarEntry entry = entries.nextElement();
+      String entryName = entry.getName();
+      if (entryName.startsWith("ramls/") && entryName.endsWith(".json")) {
+        String schemaPath = entryName.substring(6);
+        if(!schemaPath.contains("/")) {
```
julianladisch: Does this work for ramls/instance.json?

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
```diff
@@ -0,0 +1,168 @@
...
+
+  private List<String> getSchemas() throws IOException {
+    List<String> schemas = new ArrayList<>();
+    File jarFile = new File(getClass().getProtectionDomain().getCodeSource().getLocation().getPath());
+    JarFile jar = new JarFile(jarFile);
+    Enumeration<JarEntry> entries = jar.entries();
+    while(entries.hasMoreElements()) {
+      JarEntry entry = entries.nextElement();
+      String entryName = entry.getName();
+      if (entryName.startsWith("ramls/") && entryName.endsWith(".json")) {
```
julianladisch: Some json files do not have a .json extension but a .schema extension: https://github.com/folio-org/raml/tree/raml1.0/schemas

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
wwelling: I assumed uniqueness of JSON Schema file names. I will address this.

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
wwelling: It does. The idea is to only check files in the ramls directory and not subdirectories.

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
wwelling: Is there no standardization? I noticed schemas in the schemas directory, which can be retrieved individually. Similarly, there are .sample JSON files. For the list of JSON Schemas a module provides are only directly under the ramls directory.

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
wwelling: Sure.

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
wwelling: http://localhost:9130/jsonSchemas?path=instance.json

http://localhost:9130/jsonSchemas?path=raml-util/schemas/codex/instance.json

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemaAPI.java
wwelling: The JSON schemas that are used to generate POJOs are the only ones intended to be discovered here.

julianladisch:
The jar file is not available during unit tests: https://stackoverflow.com/questions/32829540/how-to-iterate-through-resource-files-regardless-of-execution-from-eclipse-or-ja
My suggestion is to create the list of all files in the https://github.com/folio-org/raml-module-builder/blob/master/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java and write that list into a `target/classes/ramls/ramls.txt` file. GenerateRunner is invoked before the unit test phase.
The implementation can serve all requests by reading one resource file (ramls.txt or some raml file) and no longer needs to read the jar.

wwelling:
Wouldn't you still have to read `ramls.txt` file from the jar? Additionally, that will not be enough as we need to serve up any of the RAML or JSON schema files the module offers either the ones it defines or the ones it references from the raml submodule.

wwelling:
I will give that a try. Hopefully, Class.getResource with the paths added to the ‘ramls.txt’ will be able to resolve the input stream or file weather running from jar or file system. Thanks.

Closing PR

PR #325, opened 2018-11-14
RMB-300: optional leading slash for ResourceUtil::asString

julianladisch:


adamdickmeiss:
Compilation failure in a number of places when merged with master. Seems like the API was changed exception wise.

PR #326, opened 2018-11-15
JSON schema and RAML multiple interface implementation

wwelling:


domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -99,30 +108,33 @@ public static void main(String[] args) throws Exception {
...
-      x1 = x1.getParentFile();
-      if (x1 == null) {
+      temp = temp.getParentFile();
+      if (temp == null) {
         break;
       } else {
-        if (x1.getName().equals(SOURCES_DEFAULT)) {
-          x0 = x1;
+        if (temp.getName().equals(SOURCES_DEFAULT)) {
+          input = temp;
```
julianladisch: Thank you for better variable names!
Can you do us a favor and move this file computation in a method so that only

    File input = rebase(ramlFiles[0]);

remains? (You may come up with some better method name.)

wwelling: Yeah.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -175,4 +187,17 @@ public void generate(String inputDirectory) throws Exception {
     log.info("processed: " + numMatches + " raml files");
   }
 
+  public static void createLookupList(File directory, String name, String...suffixes) throws IOException {
+    File listFile = new File(directory.getAbsolutePath() + File.separator + name);
+    FileOutputStream fos = new FileOutputStream(listFile);
```
julianladisch: This fos assignment should also go into the declaration statement of the try-with-resources statement.

Or you switch to a java.nio.file solution:

    Path path = Paths.get(directory.getAbsolutePath(), name);
    try (BufferedWriter bw = Files.newBufferedWriter(path)) {

wwelling: Good catch. Will use java.nio.

domain-models-runtime/pom.xml
```diff
@@ -357,6 +358,22 @@
...
+            <configuration>
+              <outputDirectory>${basedir}/target/classes/ramls</outputDirectory>
+              <resources>
+                <resource>
+                  <directory>${ramlfiles_path}</directory>
+                  <filtering>true</filtering>
+                </resource>
+              </resources>
+            </configuration>
+          </execution>
```
julianladisch: Thank you for the unit tests!

The raml directory with test.raml and test.schema ends up in the .jar file. But other modules that use RMB do not need them because they are for RMB unit tests only.
My suggestion is to move this raml directory from domain-models-runtime/raml to domain-models-runtime/src/test/resources/raml, and to replace ${basedir}/target/classes/ramls by ${basedir}/target/test-classes/ramls.

domain-models-runtime/src/main/java/org/folio/rest/impl/RamlsAPI.java
```diff
@@ -0,0 +1,138 @@
...
+      String path = matcher.group(0);
+      if (path.contains(RAMLS_PATH)) {
+        path = path.substring(path.lastIndexOf(RAMLS_PATH) + RAMLS_PATH.length());
+      }
+      if (path.endsWith(RAML_EXT)) {
+        matcher.appendReplacement(sb, Matcher.quoteReplacement(okapiUrl + "/_/ramls?path=" + path));
+      } else {
+        matcher.appendReplacement(sb, Matcher.quoteReplacement(okapiUrl + "/_/jsonSchemas?path=" + path));
+      }
+    }
```
julianladisch: Can you add a unit test for the code inside of this while loop?

domain-models-runtime/pom.xml
wwelling: That makes sense.

domain-models-runtime/src/main/java/org/folio/rest/impl/RamlsAPI.java
wwelling: Sure can.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -175,4 +176,33 @@ public void generate(String inputDirectory) throws Exception {
...
   }
 
+  public static void createLookupList(File directory, String name, String...suffixes) throws IOException {
+    File listFile = new File(directory.getAbsolutePath() + File.separator + name);
+    Path path = Paths.get(directory.getAbsolutePath(), name);
+    try (BufferedWriter bw = Files.newBufferedWriter(path)) {
+      for (File file: directory.listFiles((FileFilter) new SuffixFileFilter(suffixes, IOCase.INSENSITIVE))) {
+        log.info("lookup entry: " + file.getName());
+        bw.write(file.getName());
+        bw.newLine();
```
julianladisch: This is \r\n under Windows. Is it intended that operating system of the build server influences the line ending of the .list files in the jar?

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/GenerateRunnerTest.java
```diff
@@ -126,4 +139,38 @@ public void defaultRamlFilesDir() throws Exception {
...
+    GenerateRunner.createLookupList(dest, GenerateRunner.JSON_SCHEMA_LIST, ".json", ".schema");
+    URL jsonSchemaListUrl = getClass().getClassLoader().getResource("ramls/" + GenerateRunner.JSON_SCHEMA_LIST);
+    assertNotNull(jsonSchemaListUrl);
+    List<String> jsonSchemas = Files.readAllLines(Paths.get(jsonSchemaListUrl.toURI()), StandardCharsets.UTF_8);
+    File[] files = src.listFiles((FileFilter) new SuffixFileFilter(new String[] { ".json", ".schema" }, IOCase.INSENSITIVE));
+    assertEquals(files.length, jsonSchemas.size());
+    for (File file: files) {
+      assertTrue(jsonSchemas.contains(file.getName()));
+    }
+  }
```
julianladisch: These two test methods contain a copy of the code to test, instead test against some expected values.
These two test methods are almost the same, the common code should be extracted.

How about something like this:

    @Test
    public void testCreateRamlsLookupList() {
      assertThat(createLookupList(".raml"),
        containsInAnyOrder("jobs.raml\n", "admin.raml\n", "sample.raml\n"));
    }
  
    @Test
    public void testCreateJsonSchemasLookupList() {
      assertThat(createLookupList(".json", ".schema"),
          containsInAnyOrder(
              "book.schema\n",
              "job.schema\n",
              "bulks.schema\n",
              "last_modified.schema\n",
              "jobs.schema\n",
              "bulk.schema\n",
              "parameter.schema\n",
              "jobs_confs.schema\n",
              "jobs_conf.schema\n"));
    }
  
    private List<String> createLookupList(String... suffixes) {
      try {

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
wwelling: https://docs.oracle.com/javase/8/docs/api/java/io/BufferedWriter.html#newLine--

As long as we are aware of the platform dependent line separator, I think this is appropriate.

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/GenerateRunnerTest.java
wwelling: Makes sense to avoid same lines of code that are to be tested. Will update.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
wwelling: We use IOUtils to read the file, which uses a BufferedReader.

https://commons.apache.org/proper/commons-io/javadocs/api-2.5/src-html/org/apache/commons/io/IOUtils.html#line.1288

wwelling: Java Files readAllLines is also safe.

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-java.nio.charset.Charset-

wwelling:
@julianladisch I have wrapped up the testing improvements. Thank you for the help and reviews.

domain-models-runtime/src/main/java/org/folio/rest/impl/JsonSchemasAPI.java
```diff
@@ -0,0 +1,137 @@
...
+      return new ArrayList<>();
+    }
+  }
+
+  private List<String> getSchemas() {
+    return JSON_SCHEMAS;
+  }
+
+  private String getJsonSchemaByPath(String path, String okapiUrl) throws URISyntaxException, IOException {
+    URL jsonSchemaUrl = getClass().getClassLoader().getResource(RAMLS_PATH + path);
```
julianladisch: Under Windows the values are `RAMLS_PATH + path = ramls\test.schema` and `jsonSchemaUrl = file:/C:/Users/me/raml-module-builder/domain-models-runtime/target/test-classes/ramls%5ctest.schema` and this results in a FileNotFoundException.

julianladisch: RMB has ResourceUtil::asString that can handle reading from both jar and classes directory. The latest version from the pull request should work: https://github.com/folio-org/raml-module-builder/pull/325

wwelling: Thanks. Will give that a try.

PR #327, opened 2018-11-17
Simplify starting embedded postgres RMB-299

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3129,43 +3129,23 @@ public void startEmbeddedPostgres() throws IOException {
...
-
-      config.getAdditionalInitDbParams().addAll(Arrays.asList(
-        "-E", "UTF-8",
-        "--locale", locale
-      ));
-
-      postgresProcess = runtime.prepare(config).start();
-
-      log.info("embedded postgress started....");
+      EmbeddedPostgres e = new EmbeddedPostgres(Version.Main.V10);
```
julianladisch: Avoid single letter variable name e, renaming will make the code more easy to read and understand.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3119,6 +3112,13 @@ private void execute(String[] sql, boolean stopOnError,
     });
   }
 
+  static void startPostgresProcess(EmbeddedPostgres e) {
+    Optional<PostgresProcess> p = e.getProcess();
+    if (p.isPresent()) {
+      postgresProcess = p.get();
+    }
+  }
```
marcjohnson-kint: It looks like this method does not start the PostgreSQL process, rather it remembers the process that has just been started (in order to allow for repeated execution of `startEmbeddedPostgres` with only one embedded PostgreSQL started).

Is that a sensible interpretation?

If so, could this instead be named something like `rememberEmbeddedPostgresProcess`? 

(it can also be made private)

adamdickmeiss:
The latest commit is no good. Please do not merge.

PR #332, opened 2018-11-21
RMB-305: incorporate the use of DISTINCT (ON)

craigmcnally:
WIP - need to add unit tests before requesting review.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -21,6 +21,8 @@ import java.util.Map;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.crypto.SecretKey;
 
@@ -47,6 +49,7 @@ import org.folio.rest.tools.messages.Messages;
 import org.folio.rest.tools.monitor.StatsTracker;
 import org.folio.rest.tools.utils.Envs;
 import org.folio.rest.tools.utils.LogUtil;
+import org.folio.rest.tools.utils.NaiveSQLParse;
 import org.folio.rest.tools.utils.ObjectMapperTool;
 import org.folio.rest.tools.utils.ResourceUtils;
 import org.postgresql.copy.CopyManager;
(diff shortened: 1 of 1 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -848,6 +852,10 @@ public void save(AsyncResult<SQLConnection> sqlConnection, String table, String
    * @param replyHandler  result, containing the id field for each inserted element of entities
    */
   public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<ResultSet>> replyHandler) {
+    saveBatch(table, entities, DEFAULT_JSONB_FIELD_NAME, replyHandler);
```
craigmcnally: NOTE I started adding this to facilitate unit testing.  It's not directly related to DISTINCT ON

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1494,7 +1516,16 @@ QueryHelper buildSelectQueryHelper(
...
+    String distinctOnClause = "";
+    if (distinctOn != null && !distinctOn.isEmpty()) {
+      String[] fields = distinctOn.split(",");
+      String distinctOnFields = Stream.of(fields)
+          .map(str -> String.format("lower(f_unaccent(%s))", str))
+          .collect(Collectors.joining(","));
+      distinctOnClause = String.format("DISTINCT ON (%s) ", distinctOnFields);
+    }
+
+    queryHelper.selectQuery = SELECT + distinctOnClause + fieldName + addIdField + FROM + schemaName + DOT + table + SPACE + where;
```
craigmcnally: A lot of string concatenation here... Consider doing this in a more efficient way

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3283,6 +3315,11 @@ static ParsedQuery parseQuery(String query) {
         query = query.replaceAll(notTrue, notTrueReplacement).replaceAll(isTrue, isTrueReplacement);
         net.sf.jsqlparser.statement.Statement statement = CCJSqlParserUtil.parse(query);
         Select selectStatement = (Select) statement;
+
+        Matcher distinctOnMatcher = Pattern.compile("(?i)((DISTINCT) (?i)ON \\((.*)\\) ).*(?i)FROM").matcher(query);
```
craigmcnally: Can probably compile this regex once and store as a static member

craigmcnally:
Added a few comments in case someone else picks this up

craigmcnally:
closing this as it's been duplicated by https://github.com/folio-org/raml-module-builder/pull/354.  The formalization has occurred within that PR.

PR #334, opened 2018-11-26
Add support for application/vnd.api+json content type for POST requests

sduvvuri-ebsco:
In mod-kb-ebsco-java, we use JSON API with content type application/vnd.api+json and while RMB PUT supports all content types, currently POST consumes only a subset of those content types and does not support 'application/vnd.api+json' due  to which we are getting the following error:
```org.folio.rest.RestVerticle
SEVERE: HV000116: The object to be validated must not be null.
java.lang.IllegalArgumentException: HV000116: The object to be validated must not be null.
	at org.hibernate.validator.internal.util.Contracts.assertNotNull(Contracts.java:41)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:190)
	at org.folio.rest.RestVerticle.isValidRequest(RestVerticle.java:1458)```
This PR hopefully should fix the above.


natali-zaitseva:
Thank you @sduvvuri-ebsco for creating PR

PR #336, opened 2018-11-29
Fix broken check for query parameters RMB-307

adamdickmeiss:


julianladisch:
Can you add a comment why valueType.contains("List") works?

julianladisch:
Can you extract this if-else into a method to allow for better unit testing?

julianladisch:
Can you convert the if-else valueType.contains(...) block into a switch-case block?

PR #341, opened 2018-12-11
RMB-293: PostgresClientIT async, postgres-runner config

julianladisch:
* Add missing Async in PostgresClientIT to prevent race conditions resulting in
  "IllegalStateException: Test already completed".
* Close all connections/postgres clients in PostgresClientIT.
* PostgresRunner default configuration and environment configuration - use the
  same as PostgresClient.


adamdickmeiss:
You changed the behavior of Postgres runner. Since almost no modules is using it, it should be safe. As far as I cam concerned, it should go away completely. No need for two "embedded" Postgres systems for RMB.

julianladisch:
The behavior change of postgres runner is backwards compatible. Existing code uses command line parameters. Readme says: "The command line parameters take precedence over DB_* environment variables." This is checked in the configCommandLine() and runnerPortCommandLine() unit tests.

PR #343, opened 2018-12-13
RMB-185: Nicer constructor for the CQLWrapper

julianladisch:


adamdickmeiss:
Looks good. Heikki is still away. Merging

PR #354, opened 2019-01-02
MODORDERS-97 Distinct on

grabsefx:
PURPOSE
* [RMB-305](https://issues.folio.org/browse/RMB-305)
* [MODORDERS-97](https://issues.folio.org/browse/MODORDERS-97) (for acquisitions tracking purposes)

APPROACH
Code refactoring and adding unit test

Actually this PR is a copy of #332


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1527,7 +1547,27 @@ QueryHelper buildSelectQueryHelper(
...
+    String distinctOnClause = "";
+    if (distinctOn != null && !distinctOn.isEmpty()) {
+      String[] fields = distinctOn.split(",");
+      String distinctOnFields = Stream.of(fields)
+          .map(str -> String.format("lower(f_unaccent(%s))", str))
+          .collect(Collectors.joining(","));
+      distinctOnClause = String.format("DISTINCT ON (%s) ", distinctOnFields);
+    }
+
+    queryHelper.selectQuery = new StringBuilder()
```
julianladisch: Use `+` instead of StringBuilder's `append` for increased readability. The compiler automatically uses a StringBuilder:
https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -67,7 +67,7 @@
     System.setProperty(LoggerFactory.LOGGER_DELEGATE_FACTORY_CLASS_NAME, "io.vertx.core.logging.Log4j2LogDelegateFactory");
   }
 
-  @BeforeClass
+ // @BeforeClass
```
julianladisch: This comment-out is not related to this pull request and should go into a separate pull request to be discussed and reviewed separately.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -353,6 +353,19 @@ private void fillA(TestContext context, PostgresClient client, String tenant, in
...
+  private void fillTableWithNumbers(TestContext context, PostgresClient client, String tenant, int ...ints) {
+    String schema = PostgresClient.convertToPsqlStandard(tenant);
+    for (int i: ints) {
+      Async async = context.async();
+      execute(context, "INSERT INTO " + schema + ".a (i) VALUES (" + i + ") ON CONFLICT DO NOTHING;");
+      client.select("SELECT i FROM " + schema + ".a" + " WHERE i =" + i, context.asyncAssertSuccess(get -> {
+        context.assertEquals(i, get.getResults().get(0).getInteger(0));
+        async.complete();
+      }));
+      async.awaitSuccess(5000);
```
julianladisch: The `client.select` check is not needed. `execute` will fail the `TextContext` if the insert fails.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1433,4 +1446,33 @@ public void selectStreamTxFailed(TestContext context) {
...
+        .reduce((n1, n2) -> n1 + ", " + n2)
+        .get();
+      context.assertEquals("4, 5, 6", result);
+      async.complete();
+    });
+    async.awaitSuccess();
+  }
+
+  @Test
+  public void closeAllClientsTest (TestContext testContext) throws Exception {
```
julianladisch: This is not related to this distinct on issue and should go into a separate pull request to be discussed and reviewed separately.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
grabsefx: Sorry, it appeared here by mistake. Fixed

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
grabsefx: reverted to "+" style

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
grabsefx: removed from this PR

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1527,7 +1547,16 @@ QueryHelper buildSelectQueryHelper(
...
     QueryHelper queryHelper = new QueryHelper(transactionMode, table, facets);
 
-    queryHelper.selectQuery = SELECT + fieldName + addIdField + FROM + schemaName + DOT + table + SPACE + where;
+    String distinctOnClause = "";
+    if (distinctOn != null && !distinctOn.isEmpty()) {
+      String[] fields = distinctOn.split(",");
+      String distinctOnFields = Stream.of(fields)
+          .map(str -> String.format("%s", str))
+          .collect(Collectors.joining(","));
+      distinctOnClause = String.format("DISTINCT ON (%s) ", distinctOnFields);
```
julianladisch: What does `str -> String.format("%s", str)` do?
Isn't this sufficient?

    if (distinctOn != null && !distinctOn.isEmpty()) {
      distinctOnClause = "DISTINCT ON (" + distinctOn + ")";
    }

grabsefx: Right, this code is redundant. So it simplified now.

grabsefx:
> Does the distinct-on extension properly work with the facets functionality?
> This distinct-on extension makes the already complicated get code more complicated.
> How many places exists where this distinct-on functionality is needed?
> Can mod-orders use the select method instead?
@craigmcnally could you please take a look at @julianladisch 's questions above?

craigmcnally:
> > Does the distinct-on extension properly work with the facets functionality?
> > This distinct-on extension makes the already complicated get code more complicated.
> > How many places exists where this distinct-on functionality is needed?
> > Can mod-orders use the select method instead?
> > @craigmcnally could you please take a look at @julianladisch 's questions above?

TBH I didn't test whether this works with facets or not during my POC work.  I haven't looked, do unit test cases exist for the facet functionality?  If so, @grabsefx we should extend those test cases to include one or more that incorporates distinct on.

grabsefx:
Hi @julianladisch . I've added unit tests that use distinctOn with the facets functionality. Please, check the latest commits

craigmcnally:
@julianladisch @marc-johnson Any chance one or both of you guys can take another look at this and approve?

PR #356, opened 2019-01-14
Make /_/jsonSchemas to return schemas from subfolders

andrii-paias:
## Purpose
As per comments in https://issues.folio.org/browse/RMB-253, it would be useful to allow /_/jsonSchemas endpoint to return schemas from subfolders.  

## Approach
Add optional property schema_paths. 
If this property is set then /\_/jsonSchemas endpoint will return all schemas from specified folders. Wildcard characters can be used to recursively search for schemas in subfolders: 
```
<systemProperty>
  <key>schema_paths</key>
  <value>schemas/**,raml-util/**</value>
</systemProperty>
```
If schema_paths is not set then /\_/jsonSchemas will return only
schemas from root folder, same as before


andrii-paias:
Thanks for review, @julianladisch. Updated PR to allow specifying recursive/non-recursive search in property.

julianladisch:
@andrii-paias Thanks for the changes.

Can you extend the unit tests?
One of the two directories test1 or test2
* should contain more than one json file to check that multiple files can be found
* should contain another subdirectory (for example `test/test2/sub`) with a json file to check that the search is recursive
* should contain a file that is filtered because of wrong or missing file extension.
Can you also test these cases?
`Arrays.asList("test")` not returning anything
`Arrays.asList("test/**")` returning all

andrii-paias:
Updated PR with additional unit tests

adamdickmeiss:
Merged. Next time: please make an issue in Jira -- pretty much with the description you had in this PR.

dmtkachenko:
> 
> 
> Merged. Next time: please make an issue in Jira -- pretty much with the description you had in this PR.

Thank you, Adam.. I'll make sure a Jira is created next time we're going to propose some changes in RMB

PR #358, opened 2019-01-15
RMB-320 populate metadata when user id information is missing in the request

dmytrotur:
## Purpose
RMB-320
Populate metadata when user id information is missing in the request
Resolves: [RMB-320](https://issues.folio.org/browse/RMB-320)

## Approach
When a record is updated in anonymous request (no information about user id in request) metadata is not populated.
This causes error if record has been previously saved with metadata.
Proposed solution: populate metadata with 'createdDate' and 'updatedDate' regardless token is anonymous.
Also 'createdByUserId' and 'updatedByUserId' should be set as database trigger sets 'undefined' if case they are empty. 
Proposed value for this field is '00000000-0000-0000-0000-000000000000' as marker for update made by a 'system'.


adamdickmeiss:
Great. I don't see tests.. You have tested it I suppose.

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -109,6 +109,7 @@
   public static final String        OKAPI_REQUESTID_HEADER          = "X-Okapi-Request-Id";
   public static final String        STREAM_ID                       =  "STREAMED_ID";
   public static final String        STREAM_COMPLETE                 =  "COMPLETE";
+  public static final String SYSTEM_USER_ID = "00000000-0000-0000-0000-000000000000";
```
julianladisch: The system user should not have the nil UUID.
https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID

adamdickmeiss: So just create some UUID for the system instead?

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -1542,6 +1543,11 @@ private void populateMetaData(Object entity, Map<String, String> okapiHeaders, S
     //the created date and by fields are stored in the db in separate columns on insert trigger so that even if
     //we overwrite them here, the correct value will be set in the db level via a trigger on update
     String json;
+    Metadata md = new Metadata();
+    md.setUpdatedDate(new Date());
+    md.setCreatedDate(new Date());
+    md.setCreatedByUserId(SYSTEM_USER_ID);
+    md.setUpdatedByUserId(SYSTEM_USER_ID);
```
OleksiiKuzminov: Should we fill updatedByUsername and createdByUsername fields?

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
OleksiiKuzminov: @adamdickmeiss 
> The system user should not have the nil UUID
What kind of problems we can find using UUID nil value? 
But we will be sure that the system will not randomly generate a user with the same UUID as the system user. Although the chance is extremely small for this case

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
dmytrotur: The case is, that we do not have information about user name from the request.
Also there is quite stage behavior of data base trigger for record with metadata ([trigger](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl)).
It erases createdByUsername and updatedByUsername on update

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
marcjohnson-kint: If we are intending to use a known nil value, or make up a well known but meaningless value, what does that provide over omitting the property entirely?

The UI attempts to resolve these IDs as references to users, so both of these will fail.

PR #362, opened 2019-01-22
Rmb 326 upload complete state set before end of stream

adamdickmeiss:


sample2/src/test/java/org/folio/rest/impl/APITests.java
```diff
@@ -108,4 +112,34 @@ public void testEchoWithBadLanguage2(TestContext context) {
         .contentType("text/plain")
         .statusCode(400);
   } 
+
+  @Test
+  public void testUpload1(TestContext context) {
```
julianladisch: Can this renamed to `testChunkedUpload`?

sample2/src/test/java/org/folio/rest/impl/APITests.java
```diff
@@ -108,4 +112,34 @@ public void testEchoWithBadLanguage2(TestContext context) {
...
+    Async async = context.async();
+    final int sz = 100000;
```
julianladisch: Can this renamed to `chunkSize`?

sample2/src/test/java/org/folio/rest/impl/APITests.java
```diff
@@ -108,4 +112,34 @@ public void testEchoWithBadLanguage2(TestContext context) {
...
+    final int cnt = 100;
```
julianladisch: Can this renamed to chunkCount?

adamdickmeiss:
I did the renames.. PR failure(s) that seems to be unrelated to THIS work..

PR #370, opened 2019-02-20
RMB-334 Add a way to pass properties to jsonschema2pojo

andrii-paias:
## Purpose
[RMB-334](https://issues.folio.org/browse/RMB-334)
Make it possible to generate toString, hashCode and equals methods in generated classes.
## Approach
transfer all properties that start with "jsonschema2pojo.config" to JsonMapperConfiguration map.
For example module can set following system property to generate toString method
```
<systemProperty>
  <key>jsonschema2pojo.config.includeToString</key>
  <value>true</value>
</systemProperty>
```


domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -259,4 +262,18 @@ private static File rebase(String path) {
...
   }
 
+  /**
+   * Copies properties that start with prefix SCHEMA_CONFIG_PROPERTY_PREFIX into specified map
+   * @param properties Properties to copy
+   * @param config target map
+   */
+  private void copyConfigProperties(Properties properties, Map<String, String> config) {
+    properties.stringPropertyNames().stream()
+      .filter(property -> property.startsWith(SCHEMA_CONFIG_PROPERTY_PREFIX))
```
julianladisch: Can you rename `property` to `name`?

PR #381, opened 2019-03-05
[RMB-320] Updating records containing metadata causes exceptions in database trigger

opochepam:
- Make metadata update trigger work regardless any metadata field presence
- Metadata update trigger - set date format with zone offset, perform conversion with UTC timezone
- RestVertice - fill in metadata regardless userId presence

Issue link: [RMB-320](https://issues.folio.org/browse/RMB-320)

### DIFF
--- a/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
+++ b/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
@@ -21,8 +21,6 @@ CREATE TRIGGER set_${table.tableName}_md_trigger BEFORE INSERT ON ${myuniversity
 CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_${table.tableName}_md_json()
 RETURNS TRIGGER AS $$
 DECLARE
-  createdDate timestamp WITH TIME ZONE;
-  createdBy text;
   updatedDate timestamp WITH TIME ZONE;
   updatedBy text;
 
@@ -30,26 +28,26 @@ DECLARE
   injectedMetadata jsonb;
 
 BEGIN
+  if NEW.creation_date IS NULL then
+  	RETURN NEW;
+  end if;
+
   SET TIME ZONE 'UTC';
 
-  createdBy = NEW.created_by;
-  createdDate = NEW.creation_date;
   updatedDate = NEW.jsonb->'metadata'->>'updatedDate';
   updatedBy = NEW.jsonb->'metadata'->>'updatedByUserId';
 
-  if createdDate IS NOT NULL then
-
-    injectedMetadata = jsonb_build_object(
-      'createdByUserId', createdBy,
-      'createdDate', to_char(createdDate, dateFormat),
+  injectedMetadata = jsonb_build_object(
+      'createdByUserId', NEW.created_by,
+      'createdDate', to_char(NEW.creation_date, dateFormat),
       'updatedDate', to_char(updatedDate, dateFormat)
     );
-    if updatedBy IS NOT NULL then
-      injectedMetadata = jsonb_set(injectedMetadata, '{updatedByUserId}', to_jsonb(updatedBy));
-    end if;
 
-    NEW.jsonb = jsonb_set(NEW.jsonb, '{metadata}', injectedMetadata);
+  if updatedBy IS NOT NULL then
+    injectedMetadata = jsonb_set(injectedMetadata, '{updatedByUserId}', to_jsonb(updatedBy));
   end if;
+
+  NEW.jsonb = jsonb_set(NEW.jsonb, '{metadata}', injectedMetadata);
 
 RETURN NEW;
 END;
--- a/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
+++ b/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
@@ -22,33 +22,36 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_${table.tableName}_md
 RETURNS TRIGGER AS $$
 DECLARE
   createdDate timestamp WITH TIME ZONE;
-  createdBy text ;
+  createdBy text;
   updatedDate timestamp WITH TIME ZONE;
-  updatedBy text ;
-  injectedId text;
+  updatedBy text;
+
+  dateFormat text := 'YYYY-MM-DD"T"HH24:MI:SS.MSOF';
+  injectedMetadata jsonb;
+
 BEGIN
+  SET TIME ZONE 'UTC';
+
   createdBy = NEW.created_by;
   createdDate = NEW.creation_date;
   updatedDate = NEW.jsonb->'metadata'->>'updatedDate';
   updatedBy = NEW.jsonb->'metadata'->>'updatedByUserId';
 
-  if createdBy ISNULL then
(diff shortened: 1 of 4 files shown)


domain-models-runtime/src/test/java/org/folio/rest/persist/MetadataTriggerIT.java
```diff
@@ -0,0 +1,183 @@
...
+  private static final String TENANT_ID = "test_tenant";
+  private static final String TABLE = "test_md_trigger_table";
+
+  private static Vertx vertx;
+
+  @BeforeClass
+  public static void setUp(TestContext context) {
+    vertx = VertxUtils.getVertxWithExceptionHandler();
+    try {
+      PostgresClient.getInstance(vertx).startEmbeddedPostgres();
```
julianladisch: There is no need to unconditionally start embedded postgres.
Postgres is started in the pre-integration-test phase: https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/pom.xml#L252
This is an example how to setup a *IT.java test where embedded postgres is started only when needed, for example when running a single *IT.java only: https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTransactionsIT.java#L33

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
```diff
@@ -22,33 +22,36 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_${table.tableName}_md
...
-  injectedId text;
+  updatedBy text;
+
+  dateFormat text := 'YYYY-MM-DD"T"HH24:MI:SS.MSOF';
+  injectedMetadata jsonb;
+
 BEGIN
+  SET TIME ZONE 'UTC';
+
   createdBy = NEW.created_by;
```
julianladisch: Can `createBy` been removed? It is used one time only where we can directly write `NEW.created_by`

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
```diff
@@ -22,33 +22,36 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_${table.tableName}_md
...
   createdDate = NEW.creation_date;
```
julianladisch: Can `createdDate` been removed? If is used one time only where we can directly write `NEW.creation_date`

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
```diff
@@ -22,33 +22,36 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_${table.tableName}_md
...
   updatedDate = NEW.jsonb->'metadata'->>'updatedDate';
   updatedBy = NEW.jsonb->'metadata'->>'updatedByUserId';
 
-  if createdBy ISNULL then
-    createdBy = 'undefined';
-  end if;
-  if updatedBy ISNULL then
-    updatedBy = 'undefined';
-  end if;
   if createdDate IS NOT NULL then
```
julianladisch: Can `createdDate` been removed? If is used two times only where we can directly write `NEW.creation_date`

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
julianladisch: Can `createdDate` been removed? If is used two times only where we can directly write `NEW.creation_date`

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
julianladisch: Can this long if clause been avoided by having this guard clause at the begin of the function?

    if NEW.creation_date IS NULL then
      -- no metadata object
      RETURN NEW;
    end if;

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
```diff
@@ -22,33 +22,36 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_${table.tableName}_md
...
   updatedDate timestamp WITH TIME ZONE;
-  updatedBy text ;
-  injectedId text;
+  updatedBy text;
+
+  dateFormat text := 'YYYY-MM-DD"T"HH24:MI:SS.MSOF';
+  injectedMetadata jsonb;
+
 BEGIN
+  SET TIME ZONE 'UTC';
```
julianladisch: This is used in a trigger, changing the time zone may have side effects for all SQL commands that run afterwards. Can you avoid setting the time zone?

domain-models-runtime/src/test/java/org/folio/rest/persist/MetadataTriggerIT.java
opochepam: done

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
opochepam: done

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
opochepam: done

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
opochepam: done

domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl
opochepam: Currently it is the only way I found how to transform `timestamp` to `text` with UTC timezone. Actually it applies only to the current transaction. Or do you mean it can affect some SQL commands which is run in the current transaction after trigger execution?

julianladisch: `to_char(updatedDate at time zone '+0000', dateFormat)` should do the trick.

Regarding `SET TIME ZONE`: "Once the surrounding transaction is committed, the effects will persist until the end of the session". https://www.postgresql.org/docs/current/sql-set.html

opochepam: It should not work in this case, `to_char` returns a string formatted with timezone set in configuration parameters, no matter which timezone the timestamp has.

As workaround timezone can be set to UTC before timestamp formatting and set back after.

Or it can be left at it is, but in this case timestamp is converted to string with default timezone offset like this `2019-01-01T02:00:00.000+02`

opochepam: @julianladisch also please take a loook at ffc6251f5c311b33f964e543fb3852f7eaa37807, it prevents potential issues if application and database have different timezones

julianladisch:
See discussion in
https://issues.folio.org/browse/RMB-320
https://issues.folio.org/browse/RMB-353
why we decided to use a different solution:
https://github.com/folio-org/raml-module-builder/pull/414

PR #388, opened 2019-03-13
Log CQL and generated WHERE clause together RMB-341

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -167,6 +169,7 @@ public String toString() {
     }
     spaceAppend(sb, limit.toString());
     spaceAppend(sb, offset.toString());
+    log.info("CQL >>> SQL " + this.query + " >>>" + sb.toString());
     return sb.toString();
   }
```
julianladisch: The log entry

    CQL >>> SQL name==x >>> WHERE lower(f_unaccent(jsonb->>'name')) LIKE lower(f_unaccent('x'))

is more easy to understand with a colon:

    CQL >>> SQL: name==x >>> WHERE lower(f_unaccent(jsonb->>'name')) LIKE lower(f_unaccent('x'))

julianladisch: This invokes sb.toString() two times:

    log.info("CQL >>> SQL: " + this.query + " >>>" + sb.toString());
    return sb.toString();

Can it be wrapped?

    String sql = sb.toString();
    if (log.isInfoEnabled()) {
      log.info("CQL >>> SQL: " + this.query + " >>>" + sql);
    }
    return sql;

PR #397, opened 2019-03-21
Do not spawn postgres-runner when using -Dmaven.test.skip=true RMB-349

adamdickmeiss:


adamdickmeiss:
Julian .. You expanded the problem.. I have now pushed a naive extension .. This works for -DskipITs , but I can not get it to work for -DskipTests (still spawning). If we can't get it to work, let's close with what we have.. Of course, you have an idea on how to do it, even better.

adamdickmeiss:
OK. Then we're back to sq1. I'll change the issue to its original. If somebody wants to deal with the 2 other cases, fine. Not me. Not that important anyway.

PR #401, opened 2019-03-25
Add PostClient.saveBatch variant with SQLConnection RMB-354

adamdickmeiss:
Sorry.. some formatting also part of this..

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -884,7 +884,9 @@ public class PostgresClient {
     saveBatch(table, entities, DEFAULT_JSONB_FIELD_NAME, replyHandler);
   }
 
-  public void saveBatch(String table, JsonArray entities, String column, Handler<AsyncResult<ResultSet>> replyHandler) {
+  public void saveBatch(String table, JsonArray entities, String column,
+    Handler<AsyncResult<ResultSet>> replyHandler) {
+
     client.getConnection(res -> {
       if (res.failed()) {
         replyHandler.handle(Future.failedFuture(res.cause()));
@@ -897,7 +899,9 @@ public class PostgresClient {
     });
   }
 
-  public void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table, JsonArray entities, String column, Handler<AsyncResult<ResultSet>> replyHandler) {
+  private void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table,
+    JsonArray entities, String column, Handler<AsyncResult<ResultSet>> replyHandler) {
+
     long start = System.nanoTime();
     if (entities == null || entities.isEmpty()) {
       // return empty result
@@ -937,15 +941,9 @@ public class PostgresClient {
     });
   }
 
-  /***
-   * Save a list of POJOs.
-   * POJOs are converted to json and saved in a single sql call. The generated IDs of the
-   * inserted records are returned in the ResultSet.
-   * @param table  destination table to insert into
-   * @param entities  each list element is a POJO
-   * @param replyHandler  result, containing the id field for each inserted POJO
-   */
-  public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<ResultSet>> replyHandler) {
+  private JsonArray batchEntitiesToJsonArray(List<Object> entities,
+    Handler<AsyncResult<ResultSet>> replyHandler) {
+
     JsonArray jsonArray = new JsonArray();
     try {
       for (Object entity : entities) {
@@ -954,9 +952,35 @@ public class PostgresClient {
       }
     } catch (Exception e) {
       replyHandler.handle(Future.failedFuture(e));
-      return;
+      return null;
     }
-    saveBatch(table, jsonArray, replyHandler);
+    return jsonArray;
+  }
+
+  /***
+   * Save a list of POJOs.
+   * POJOs are converted to json and saved in a single sql call. The generated IDs of the
+   * inserted records are returned in the ResultSet.
+   * @param table  destination table to insert into
+   * @param entities  each list element is a POJO
+   * @param replyHandler result, containing the id field for each inserted POJO
+   */
+  public void saveBatch(String table, List<Object> entities,
+    Handler<AsyncResult<ResultSet>> replyHandler) {
+
+    JsonArray jsonArray = batchEntitiesToJsonArray(entities, replyHandler);
+    if (jsonArray != null) {
+      saveBatch(table, jsonArray, replyHandler);
+    }
+  }
+
+  public void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table,
+    List<Object> entities, Handler<AsyncResult<ResultSet>> replyHandler) {
+
+    JsonArray jsonArray = batchEntitiesToJsonArray(entities, replyHandler);
+    if (jsonArray != null) {
+      saveBatch(sqlConnection, table, jsonArray, DEFAULT_JSONB_FIELD_NAME, replyHandler);
+    }
   }
 
   /**
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -885,50 +885,55 @@ public class PostgresClient {
   }
 
   public void saveBatch(String table, JsonArray entities, String column, Handler<AsyncResult<ResultSet>> replyHandler) {
-    long start = System.nanoTime();
-    if (entities == null || entities.isEmpty()) {
-      // return empty result
-      ResultSet resultSet = new ResultSet(
-          Collections.singletonList(idField), Collections.emptyList(), null);
-      replyHandler.handle(Future.succeededFuture(resultSet));
-      return;
-    }
     client.getConnection(res -> {
       if (res.failed()) {
         replyHandler.handle(Future.failedFuture(res.cause()));
         return;
       }
-
-      if (log.isInfoEnabled()) {
-        log.info("starting: saveBatch size=" + entities.size());
-      }
-      SQLConnection connection = res.result();
-      StringBuilder sql = new StringBuilder()
-          .append(INSERT_CLAUSE)
-          .append(schemaName).append(DOT).append(table)
-          .append(" (").append(column).append(") VALUES (?)");
-      for (int i=1; i<entities.size(); i++) {
-        sql.append(",(?)");
-      }
-      sql.append(" RETURNING ").append(idField);
-
-      connection.queryWithParams(sql.toString(), entities, queryRes -> {
-        connection.close();
-        if (queryRes.failed()) {
-          log.error("saveBatch size=" + entities.size()
-            + SPACE +
-              queryRes.cause().getMessage(),
-              queryRes.cause());
-          statsTracker("saveBatchFailed", table, start);
-          replyHandler.handle(Future.failedFuture(queryRes.cause()));
-          return;
-        }
-        if (log.isInfoEnabled()) {
-          log.info("success: saveBatch size=" + entities.size());
-        }
-        statsTracker("saveBatch", table, start);
-        replyHandler.handle(Future.succeededFuture(queryRes.result()));
+      saveBatch(res, table, entities, column, x -> {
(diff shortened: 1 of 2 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -885,50 +885,55 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
+        return;
+      }
+      saveBatch(res, table, entities, column, x -> {
+        res.result().close();
+        replyHandler.handle(x);
+      });
+    });
+  }
+
+  public void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table, JsonArray entities, String column, Handler<AsyncResult<ResultSet>> replyHandler) {
```
OleksiiMaksymov: Plz add another wrapper public method, it is more convenient for work with List<Object>.
as:
`public void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table, List<Object> entities, Handler<AsyncResult<ResultSet>> replyHandler) {....}`

In most cases, we operate on entities, and I think it is more convenient to have two methods with List of Objects and an JsonArray

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -885,50 +885,55 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
-    client.getConnection(res -> {
-      if (res.failed()) {
-        replyHandler.handle(Future.failedFuture(res.cause()));
+    if (log.isInfoEnabled()) {
+      log.info("starting: saveBatch size=" + entities.size());
+    }
+    StringBuilder sql = new StringBuilder()
+      .append(INSERT_CLAUSE)
+      .append(schemaName).append(DOT).append(table)
+      .append(" (").append(column).append(") VALUES (?)");
```
OleksiiMaksymov: ` String column` - I assume that this parameter is redundant to save objects in batch

adamdickmeiss: Redundant in the "inner" function?? Are you saying that the inner function do not need a column OR that  you just don't need it for YOUR purpose.. Just like the previous one.. WOuld like a version that takes List<Object> and no column parameter?

OleksiiMaksymov: In our case, the `column `parameter is not needed but may be needed for the future. Currently, we are using in the `mod-calendar` a list of objects without a  `column` parameter

adamdickmeiss:
Note.. I need a test for this.. But want to be sure, this kind of method is what you prefer.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -884,63 +884,66 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
+      if (res.failed()) {
+        replyHandler.handle(Future.failedFuture(res.cause()));
+        return;
+      }
+      saveBatch(res, table, entities, column, x -> {
+        res.result().close();
+        replyHandler.handle(x);
+      });
+    });
+  }
```
julianladisch: a) This method consists mostly of duplicated code.
b) The res.failed() clause is not tested.

Please use the one liner we use for all new code:

        client.getConnection(conn -> saveBatch(conn, table, entities, column, closeAndHandleResult(conn, replyHandler)));

Move handling the res.failed() code into the saveBatch(res, ...) method.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -949,9 +952,35 @@ public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<R
...
+   * @param replyHandler result, containing the id field for each inserted POJO
+   */
+  public void saveBatch(String table, List<Object> entities,
+    Handler<AsyncResult<ResultSet>> replyHandler) {
+
+    JsonArray jsonArray = batchEntitiesToJsonArray(entities, replyHandler);
+    if (jsonArray != null) {
+      saveBatch(table, jsonArray, replyHandler);
+    }
+  }
```
julianladisch: This is duplicated code of the following saveBatch method.

Please use the usual one-liner to invoke the other saveBatch method instead:

    client.getConnection(conn -> saveBatch(conn, entities, closeAndHandleResult(conn, replyHandler)));

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -949,9 +952,35 @@ public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<R
...
+
+  public void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table,
+    List<Object> entities, Handler<AsyncResult<ResultSet>> replyHandler) {
+
+    JsonArray jsonArray = batchEntitiesToJsonArray(entities, replyHandler);
+    if (jsonArray != null) {
+      saveBatch(sqlConnection, table, jsonArray, DEFAULT_JSONB_FIELD_NAME, replyHandler);
     }
-    saveBatch(table, jsonArray, replyHandler);
   }
```
julianladisch: Please add a unit test for jsonArray == null. After that please resubmit for code review!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -884,63 +884,66 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
-  /***
-   * Save a list of POJOs.
-   * POJOs are converted to json and saved in a single sql call. The generated IDs of the
-   * inserted records are returned in the ResultSet.
-   * @param table  destination table to insert into
-   * @param entities  each list element is a POJO
-   * @param replyHandler  result, containing the id field for each inserted POJO
-   */
-  public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<ResultSet>> replyHandler) {
+  private JsonArray batchEntitiesToJsonArray(List<Object> entities,
```
julianladisch: There is no docum,entation when batchEntitiesToJsonArray invokes the replyHandler.
I don't think that using the replyHandler in this method is a good idea. Simply throw the exception and let the caller handle it, that is more easy to understand and the Exception must be created anyhow.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: For these RMB methods with callbacks we want to make sure that all Exceptions are caught and reported via the replyHandler.
Please wrap into a try-catch clause.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -884,63 +884,66 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
+        return;
+      }
+      saveBatch(res, table, entities, column, x -> {
+        res.result().close();
+        replyHandler.handle(x);
+      });
+    });
+  }
+
+  private void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table,
```
julianladisch: This should be public.
The corresponding saveBatch method without sqlConnection is already public.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -884,63 +884,66 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
+    StringBuilder sql = new StringBuilder()
+      .append(INSERT_CLAUSE)
+      .append(schemaName).append(DOT).append(table)
+      .append(" (").append(column).append(") VALUES (?)");
+    for (int i = 1; i < entities.size(); i++) {
+      sql.append(",(?)");
+    }
+    sql.append(" RETURNING ").append(idField);
+
+    SQLConnection connection = sqlConnection.result();
```
julianladisch: Defensive programming: replyHandler needs to be called with a failedFuture if  sqlConnection==null or sqlConnection.failed().

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -507,6 +509,22 @@ public void saveBatchX(TestContext context) {
...
+  @Test
+  public void saveBatchXTrans(TestContext context) {
+    List<Object> list = Collections.singletonList(xPojo);
+    postgresClient = createFoo(context);
+    postgresClient.startTx(asyncAssertTx(context, trans -> {
+      postgresClient.saveBatch(trans, FOO, list, context.asyncAssertSuccess(save -> {
+        final String id = save.getResults().get(0).getString(0);
+          postgresClient.endTx(trans, context.asyncAssertSuccess(end -> {
+            postgresClient.getById(FOO, id, context.asyncAssertSuccess(get -> {
+            context.assertEquals("x", get.getString("key"));
```
julianladisch: fix indentation

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Use generics for the List type.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -949,9 +952,35 @@ public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<R
...
+
+  /***
+   * Save a list of POJOs.
+   * POJOs are converted to json and saved in a single sql call. The generated IDs of the
+   * inserted records are returned in the ResultSet.
+   * @param table  destination table to insert into
+   * @param entities  each list element is a POJO
+   * @param replyHandler result, containing the id field for each inserted POJO
+   */
+  public void saveBatch(String table, List<Object> entities,
```
julianladisch: Use generics for the List type.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -949,9 +952,35 @@ public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<R
...
+    Handler<AsyncResult<ResultSet>> replyHandler) {
+
+    JsonArray jsonArray = batchEntitiesToJsonArray(entities, replyHandler);
+    if (jsonArray != null) {
+      saveBatch(table, jsonArray, replyHandler);
+    }
+  }
+
+  public void saveBatch(AsyncResult<SQLConnection> sqlConnection, String table,
+    List<Object> entities, Handler<AsyncResult<ResultSet>> replyHandler) {
```
julianladisch: Use generics for the List type.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -949,9 +952,35 @@ public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<R
       }
     } catch (Exception e) {
       replyHandler.handle(Future.failedFuture(e));
-      return;
+      return null;
```
julianladisch: Unit test for null missing.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -949,9 +952,35 @@ public void saveBatch(String table, List<Object> entities, Handler<AsyncResult<R
...
+   * @param entities  each list element is a POJO
+   * @param replyHandler result, containing the id field for each inserted POJO
+   */
+  public void saveBatch(String table, List<Object> entities,
+    Handler<AsyncResult<ResultSet>> replyHandler) {
+
+    JsonArray jsonArray = batchEntitiesToJsonArray(entities, replyHandler);
+    if (jsonArray != null) {
+      saveBatch(table, jsonArray, replyHandler);
+    }
```
julianladisch: Please add unit test for jsonArray == null.

PR #402, opened 2019-03-26
Rmb 355 facets 2

adamdickmeiss:


adamdickmeiss:
I'm not too happy with getDistinctOnWithFacets.. If I could get real facets back and NOT get error 
 "column \"_id\" does not exist"
that would be better. To me it looks like _id is a column in the table definition.. Any ideas ?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1497,6 +1501,8 @@ public void streamGetDistinctOn(TestContext context) throws IOException {
     async.awaitSuccess();
   }
 
+  // While facets are passed, this does NOT seem to deal with facets
+  // in fact it, quite possibly streamGet do not support facets at all
```
julianladisch: Can you fix the wording?

adamdickmeiss: I don't know how. I don't think streamGet even deal with facets.. Is that something I should investigate? (There's no regression.-- streamGet possibly never supported facts)

PR #408, opened 2019-04-09
Update to cql2pgjson 4.0.0 RMB-368

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -1,232 +1,138 @@
...
- *         c.field = "'price' -> 'po_currency' ->> 'value'";
- *         c.operation = "LIKE";
- *         c.value = "USD";
+ * Criteria c = new Criteria(); c.field = "'price' -> 'po_currency' ->>
+ * 'value'"; c.operation = "LIKE"; c.value = "USD";
  *
- *         c.field = "'rush'";
- *         c.operation = "IS TRUE";
- *         c.value = null;
+ * c.field = "'rush'"; c.operation = "IS TRUE"; c.value = null;
```
julianladisch: The code example should get reverted to the previous formatting (line wrapping) and be wrapped within some javadoc tags, see https://stackoverflow.com/questions/541920/multiple-line-code-example-in-javadoc-comment

adamdickmeiss: Like this?
```
/**
 *
 * Criteria c = new Criteria(); c.field = "'price' -> 'po_currency' ->>
 * 'value'"; c.operation = "LIKE"; c.value = "USD";
 *
 * <code>
 * c.field = "'rush'";
 * c.operation = "IS TRUE";
 * c.value = null;
 * </code>
 */
```

??

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -235,164 +141,73 @@ private String addPrefix() {
...
     for (int i = 0; i < size; i++) {
-      if(isArray){
+      if (isArray) {
         //skip first field if array - included in FROM clause with an AS
-        if(i==0){
+        if (i == 0) {
           i++;
         }
       }
       String fVal = field.get(i);
```
julianladisch: Can you remove `fVal` and use `field.get(i)` directly?

adamdickmeiss: Sure

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: At least with code tag, the editor leaves it alone on reformat (which is what happened before)

julianladisch: This is better, but the rendered javadoc is still a single line with all commands concatenated. Your IDE should show you the rendered javadoc if you move your mouse over "Criteria" of "public class Criteria".

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -23,210 +17,127 @@
...
+  private static final String ARRAY_FROM_CLAUSE = "jsonb_array_elements";//("jsonb->'fund_distributions'[]->'amount'->>'sum'")
+  private static final String GET_JSON_FIELD = "->";
+  private static final String GET_VALUE_FIELD = "->>";
+  private static final Pattern JSON_OPS = Pattern.compile("@>|#>|#>>|<@", Pattern.CASE_INSENSITIVE);
+
+  String alias = null;
+
+  boolean joinON = false;
+
+  String forceCast = null;
```
julianladisch: Can this variable removed?

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -23,210 +17,127 @@
...
+
+  int valueType = 1;
```
julianladisch: Can this variable removed?

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -23,210 +17,127 @@
...
 
   String column;
```
julianladisch: Can this variable removed?

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -431,60 +242,46 @@ public Criteria setOperation(String operation) {
...
 
-  public boolean isJSONB() {
-    return isJSONB;
-  }
-
   /**
-   * set this to false if not a jsonb field criteria
-   * For example: the _id is not in the jsonb object hence would be false if criteria based on _id
+   * set this to false if not a jsonb field criteria For example: the _id is not
+   * in the jsonb object hence would be false if criteria based on _id
```
julianladisch: The new line wrapping should be accompanied by polishing punctuation and capitalisation.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -431,60 +242,46 @@ public Criteria setOperation(String operation) {
...
    */
   public Criteria setJSONB(boolean isJSONB) {
     this.isJSONB = isJSONB;
     return this;
   }
 
-  public boolean isNotQuery() {
-    return isNotQuery;
-  }
-
```
julianladisch: As long as there is the setter setNotQuery we can have the getter isNotQuery. Possible usage: Criterion.java should use the getter instread of directly reading the variable.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -431,60 +242,46 @@ public Criteria setOperation(String operation) {
...
   public void setNotQuery(boolean isNotQuery) {
     this.isNotQuery = isNotQuery;
   }
 
-  public boolean isArray() {
-    return isArray;
-  }
-
```
julianladisch: As long as there is the setter setArray we can have the getter isArray. Even if it not used yet, but it could.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -431,60 +242,46 @@ public Criteria setOperation(String operation) {
...
   public void setArray(boolean isArray) {
     this.isArray = isArray;
   }
 
-  public String getArrayField() {
-    return arrayField;
-  }
-
```
julianladisch: As long as there is the setter setArrayField we can have the getter isArrayField. Even if it not used yet, but it could.

julianladisch:
Can you move CriteriaTest into the expected package/directory org.folio.rest.persist.criteria?

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: yes

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: yes

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: yes

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: sure

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
adamdickmeiss: Will remove this and its sibling setArrayField(sring field).

adamdickmeiss:
Yes. About the move.

PR #410, opened 2019-04-11
Update SchemaDereferencer.java

openSupporter237:
I modified the fixupRef fuction to generate a correct absolute path of $ref. The path will have a slash before the drive letter and it will go well when we use URI's getPath() function  in windows environment.


adamdickmeiss:
Thanks

PR #411, opened 2019-04-22
Rmb 347 remove gen random UUID

evaluk:


julianladisch:
Can you remove generateId, isGenerateId, setGenerateId from
https://github.com/folio-org/raml-module-builder/blob/v23.12.0/domain-models-runtime/src/main/java/org/folio/rest/persist/ddlgen/Table.java ?

evaluk:
> Can you remove generateId, isGenerateId, setGenerateId from
> https://github.com/folio-org/raml-module-builder/blob/v23.12.0/domain-models-runtime/src/main/java/org/folio/rest/persist/ddlgen/Table.java ?

will this trigger a serialization error in schema files with the parameter?

julianladisch:
Yes, you should remove generateId, isGenerateId, setGenerateId from Table.java and try whether the exception error message is good enough when generateId exists in schema.json. If the exception error message is not good enough you can catch the exception and create a better message.

evaluk:
I have completed the updates to remove all of uses of gen random uuid.  Please take a look and let me know what you think.  My branch passes CI so hopefully that means something as far as functionality of the new values..

domain-models-runtime/src/main/java/org/folio/rest/persist/ddlgen/SchemaMaker.java
```diff
@@ -114,15 +114,13 @@ public String generateDDL(boolean recreateIndexMode) throws IOException, Templat
       int size = tables.size();
       for (int i = 0; i < size; i++) {
         Table t = tables.get(i);
-
-        if(t.getMode() == null){
+        if(t.getMode() == null){
```
julianladisch: A space was added after the {
The editor/IDE should be configured to automatically remove trailing whitespace in java files, see https://dev.folio.org/guides/developer-setup/#coding-style

evaluk: I looked into why my ide didnt catch this, and came to the conclusion that the eclipse version of the editor config plug in does not support many key features of editor config.  This is less then ideal.  I may look into switching to VS Code or something which I use in other cases but was not able to get junit/maven working to my satisfaction when i started working with RMB.  May give it another shot.

PR #417, opened 2019-05-16
RMB-346: Always sync _id and jsonb->>'id' using a trigger, drop 'populateJsonWithId: false' option

julianladisch:


evaluk:
The User story mentions adding text to the readme.md regarding the trigger addition?

PR #420, opened 2019-05-20
RMB-376: merge CQL2PG into RMB

julianladisch:
* Merge CQL2PG into RMB
* Create dbschema submodule with ddlgen classes to avoid this circular dependency:
  [ERROR] The projects in the reactor contain a cyclic reference: Edge between
  'Vertex{label='org.folio:domain-models-runtime:24.1.0-SNAPSHOT'}' and
  'Vertex{label='org.folio:cql2pgjson:24.1.0-SNAPSHOT'}' introduces to cycle in the
  graph org.folio:cql2pgjson:24.1.0-SNAPSHOT
  --> org.folio:domain-models-runtime:24.1.0-SNAPSHOT
  --> org.folio:cql2pgjson:24.1.0-SNAPSHOT
* Replace drools ComparableVersion by copy of apache maven ComparableVersion to
  avoid drools dependency.
* Codestyle: disable AvoidInlineConditionals


hjiebsco:
Just noticed that git history was not preserved after the merging. Is this OK regarding Apache license requirement?

PR #423, opened 2019-05-21
RMB-379: wrong masking of Criteria.setValue(Object) - SQL Injection

julianladisch:


julianladisch:
These are the unit tests only. The failing unit tests are commented out.

julianladisch:
There are three ways to call Criteria.setValue(Object):

    setValue("'" + value.replace("'", "''") + "'");
    setValue("'" + value + "'");
    setValue(value);

The first one is always correct and will never result in any SQL injection.

The second and the third one are prone to SQL injection if value may contain single quotes.

The documentation does not mention the first way.

My proposal is to replace setValue(Object) by setVal(String) where setVal(String) will only accept the third way. Internally it will always wrap the value in single quotes and mask any containing single quotes by duplicating them.

This prevents SQL injection and is also the most easy way to invoke. The new method name will trigger compile errors. This forces developers
* to replace setValue by setVal and
* to remove any single quote adding.

I don't see a reason why one would like to pass in an Object. We may add setValObject(Object) or similar methods lateron if needed.

PR #428, opened 2019-05-27
Try with non-static hash arrays RMB-384

adamdickmeiss:


adamdickmeiss:
This change is known to fix the issue that mod-invoice-storage was seen.

PR #430, opened 2019-05-28
Cqlpg 99 implement foreign key support for cqlpg json parser

evaluk:
Pull request opening for discussion


adamdickmeiss:
I can confirm that the branch fails in (unit,it) testing. Clearly needs resolving before merge

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,126 @@
...
+  @Test
+  public void ForeignKeySearchWithMissingFK() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKeyMissingFK.json");
+    try {
+      cql2pgJson.toSql("tableb.tableb_data == 11111111-1111-1111-1111-111111111111").getWhere();
+      fail();
+    } catch(QueryValidationException q) {
+      //left empty on purpose
+    }
```
julianladisch: Can you add a check that the exception message contains both "foreignKeys object missing" and the table name?

You may use `Exception exception = assertThrows` from JUnit5.
Or you may use JUnit4's ExpectedException as in https://github.com/folio-org/raml-module-builder/blob/fc7e14ea972077456fb015e03010f5cc4b670656/domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerTest.java#L48-L66

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+  @Test
+  public void ForeignKeySearchWithMalformedFK() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKeyMalformedFK.json");
+    try {
+      cql2pgJson.toSql("tableb.tableb_data == 11111111-1111-1111-1111-111111111111").getWhere();
+      fail();
+    } catch(QueryValidationException q) {
+      //left empty on purpose
```
julianladisch: Can you add a check that the exception message contains both "foreignKeys.targetTable missing" and the current table name?

You may use assertThrows from JUnit5 or ExpectedException from JUnit4.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,126 @@
...
+  public void ForeignKeySearchFailureDueToTable() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.jsonb" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
+    try {
+      String sql = cql2pgJson.toSql("tablec.tablec_data == 11111111-1111-1111-1111-111111111111").getWhere();
+      assertEquals("lower(f_unaccent(tablea.jsonb->'tablec'->>'tablec_data')) LIKE lower(f_unaccent('11111111-1111-1111-1111-111111111111'))",sql);
+      sql = cql2pgJson.toSql("ardgsdfgdsfg.tableb_data == 11111111-1111-1111-1111-111111111111").getWhere();
+      assertEquals("lower(f_unaccent(tablea.jsonb->'ardgsdfgdsfg'->>'tableb_data')) LIKE lower(f_unaccent('11111111-1111-1111-1111-111111111111'))",sql);
+    } catch(Exception e) {
+      e.printStackTrace();
```
julianladisch: What do you expect to catch here?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+    // default pkColumnName is id without underscore
+    //assertEquals(" ( SELECT TableA.jsonb->>'tableb_data' from TableB WHERE (lower(f_unaccent(TableA.jsonb->>'id'::text)) = lower(f_unaccent(TableB.jsonb->>'tableb_data'::text)))) LIKE lower(f_unaccent(11111111-1111-1111-1111-111111111111))", sql);
```
julianladisch: Can this comment been removed?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -60,6 +60,9 @@
   /** name of the primary key column */
   private static final String PK_COLUMN_NAME = "id";
 
+  private final Pattern uuidPattern = Pattern.compile("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
+  private final Pattern tableNamePattern = Pattern.compile("^[0-9a-zA-Z_-]+\\.[0-9a-zA-Z_-]+$");
```
julianladisch: The tableNamePattern matches 1.5 but this is not a valid table name.
Minus is an invalid character for a table name.
See https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS for the spec.
My suggestion is "^[a-zA-Z_][0-9a-zA-Z_]*\\.[0-9a-zA-Z_]+$"

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -491,6 +507,120 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+      }
+    }
+    return null;
+  }
+
+  private String subQuery(String index,CQLTermNode node, Table correlation) throws QueryValidationException {
+    String[] idxParts = index.split("\\.");
+    String[] termParts = node.getTerm().split("\\.");
+
+    String [] foreignTarget ;
```
julianladisch: String [] foreignTarget = (idxParts.length > termParts.length) ? idxParts : termParts;

improves readability.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -491,6 +507,120 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+      throw new QueryValidationException(msg);
+    }
+    if (fkey.getTargetTable() == null) {
+      String msg = "subQuery: Missing target table for foreignKey field " + fkey.getFieldName();
+      logger.log(Level.SEVERE, msg);
+      throw new QueryValidationException(msg);
+    }
+    try {
+
+      CQL2PgJSON c = new CQL2PgJSON(foreignTarget[0] + ".jsonb");
```
julianladisch: You can avoid
- to create a new CQL2PgJSON instance
- the catch block (and the try line)
- repeatedly using `foreignTarget[0] + ".jsonb"`

if you replace this line with `String foreignJsonb = foreignTarget[0] + ".jsonb";`

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -491,6 +507,120 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+
+      String indexString = "";
+      String selectString = "";
+      if (isTermConstant) {
+        String termString = "";
+        if(isTermUUID) {
+          termString = "(" + term + ")::UUID";
+          indexString =  "(" + whereField + ")::UUID";
+        } else {
+          termString = wrapInLowerUnaccent("'" + Cql2SqlUtil.cql2like(term) + "'", modifiers);
```
julianladisch: This only works for the LIKE operator but not for the = operator.

julianladisch: You may want to add a Cql2SqlUtil.cql2quotedLike(String) method.

julianladisch: You cannot use modifiers. We dropped supporting CQL modifiers in https://issues.folio.org/browse/CQLPG-81 Note 3.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -579,7 +579,7 @@ private String subQuery(String index,CQLTermNode node, Table correlation) throws
       if (isTermConstant) {
         String termString = "";
         if(isTermUUID) {
-          termString = "(" + term + ")::UUID";
+          termString = "('" + Cql2SqlUtil.cql2like(term) + "')::UUID";
```
julianladisch: You can you use Cql2SqlUtil.cql2like only after a LIKE operator.
For quoting a SQL string you need to create a new method Cql2SqlUtil.quotedString(String)

evaluk:
I am looking into reducing duplication here to pass sonarcloud.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -24,16 +24,28 @@
...
-   * CQL escapes * ? ^ \ and SQL LIKE escapes \ % _.
+   * Signal the underlying function that this is not a quoted String case.
    *
    * @param s  CQL string without leading or trailing double quote
    * @return SQL LIKE string without leading or trailing single quote
    */
   @SuppressWarnings("squid:S3776")  // suppress "Cognitive Complexity of methods should not be too high"
   public static String cql2like(String s) {
+    return cql2String(s,false);
+
```
julianladisch: Can you remove the empty line?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -24,16 +24,28 @@
...
+   * Signal the underlying function that this is not a quoted String case.
    *
    * @param s  CQL string without leading or trailing double quote
    * @return SQL LIKE string without leading or trailing single quote
    */
   @SuppressWarnings("squid:S3776")  // suppress "Cognitive Complexity of methods should not be too high"
   public static String cql2like(String s) {
+    return cql2String(s,false);
+  
+  }
```
julianladisch: Can you insert an empty line after the method?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -24,16 +24,28 @@
...
+  /**
+   * Convert a CQL string to an SQL string.
+   * CQL escapes * ? ^ \ and SQL LIKE escapes \ % _.
+   *
+   * @param s  CQL string without leading or trailing double quote
+   * @param quotedString  Whether this is a quoted String following an = sign
+   * @return SQL string without leading or trailing single quote
+   */
+  @SuppressWarnings("squid:S3776")  // suppress "Cognitive Complexity of methods should not be too high"
+  public static String cql2String(String s, boolean quotedString) {
```
julianladisch: There should be a cql2string method similar to the cql2like method:

    public static String cql2string(String s) {
      return cql2string(s, true);
    }

julianladisch: Can you replace the capital S by a small s to match the other method names?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -43,41 +55,38 @@ public static String cql2like(String s) {
...
         }
         break;
       case '*':
         if (backslash) {
           like.append("\\*");
-          backslash = false;
         } else {
           like.append('%');
         }
         break;
```
julianladisch: If the CQL string contains an unescaped ? or an unescaped * then an exception should be thrown.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
julianladisch: The joined method with the quotedString boolean parameter is less readable than two separate methods.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/Cql2SqlUtilTest.java
```diff
@@ -76,10 +76,29 @@ public Object cql2likeParams() {
...
+        "\\?   \\?  ",
+        "\\'   ''   ",
+        "\\\\  \\\\ "
+        );
+  }
 
+  @Test
+  @Parameters(method = "quotedStringParams")
+  public void cql2like(String cql, String sql) {
+    assertThat(Cql2SqlUtil.cql2String(cql, true), is(sql));
```
julianladisch: The name of the unit test method is wrong, should be quotedString.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/Cql2SqlUtilTest.java
```diff
@@ -76,10 +76,29 @@ public Object cql2likeParams() {
...
+  public void quotedString(String cql, String sql) {
     assertThat(Cql2SqlUtil.cql2like(cql), is(sql));
   }
+  public Object quotedStringParams() {
+    return params(
+        "           ",
+        "'     ''   ",
+        "a     a    ",
+        "*     %    ",
+        "?     _    ",
```
julianladisch: Both * and ? test cases are wrong because such an unmasked CQL wildcard cannot be used in an SQL string constant. Should throw an exception.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/Cql2SqlUtilTest.java
```diff
@@ -76,10 +76,29 @@ public Object cql2likeParams() {
...
+        "\\    \\\\ ",
```
julianladisch: A single backslash in CQL (is invalid but) should result in a single backslash in SQL. In an SQL string constant there is no escaping character other than the double '

    SELECT * FROM (VALUES ('\'), ('\\'), ('\''\')) t;
     column1 
    ---------
     \
     \\
     \'\

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/Cql2SqlUtilTest.java
```diff
@@ -76,10 +76,29 @@ public Object cql2likeParams() {
...
+        "\\*   \\*  ",
+        "\\?   \\?  ",
```
julianladisch: There is no backslash escape in an SQL string.

Correct:

    "\\?   ?  ",

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/Cql2SqlUtilTest.java
```diff
@@ -76,10 +76,29 @@ public Object cql2likeParams() {
...
+        "\\'   ''   ",
+        "\\\\  \\\\ "
```
julianladisch: There is no backslash escape in an SQL string.

Correct:

    "\\\\  \\ ",

The CQL string "\\" represents a single backslash. In CQL it needs to be prepended by a masking backslash.
The SQL string '\' represents a single backslash.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -0,0 +1,87 @@
...
+  public void ForeignKeySearchWithInjection1() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+
+    assertThat(cql("tableb.prefix == \"x0')));((('DROP tableb\""), containsInAnyOrder("test3") );
+  }
+  @Test
+  public void ForeignKeySearchWithInjection2() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+
+    assertThat(cql("tableb.prefix == \"x0')  or 1=1)--\""), is(empty()) );
+  }
+}
```
julianladisch: Can you remove all throws declarations that are not strictly needed?
Can you insert a blank line between all methods?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+  @Test
+  public void ForeignKeySearchWithInjection1() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+
+    assertThat(cql("tableb.prefix == \"x0')));((('DROP tableb\""), containsInAnyOrder("test3") );
+  }
+  @Test
+  public void ForeignKeySearchWithInjection2() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+
+    assertThat(cql("tableb.prefix == \"x0')  or 1=1)--\""), is(empty()) );
```
julianladisch: Both injection tests don't fail if there is some injection.
Can you rewrite them?
For example `tableb.prefix == '`

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,122 @@
...
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE f_unaccent(tableb.jsonb->>'otherindex') = f_unaccent('y0'))", sql);
+  }
+  @Test
+  public void ForeignKeySearchWithMissingFK() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKeyMissingFK.json");
+
+    thrown.expect(QueryValidationException.class);
+    thrown.expectMessage("No foreignKey for table tableb found");
+    cql2pgJson.toSql("tableb.prefix == 11111111-1111-1111-1111-111111111111").getWhere();
```
julianladisch: Can this line been removed?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,122 @@
...
+@RunWith(JUnitParamsRunner.class)
+public class ForeignKeyGenerationTest  {
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+  @Test
+  public void ForeignKeySearchNumeric() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
+    String sql = cql2pgJson.toSql("tableb.blah == /number 123452").getWhere();
+    // default pkColumnName is id without underscore
```
julianladisch: Can this comment been removed from all test methods?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,122 @@
...
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE (tableb.jsonb->>'blah')::NUMERIC = ('123452')::NUMERIC)", sql);
+  }
+  @Test
+  public void ForeignKeySearch() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
```
julianladisch: Can we use a better method name, for example searchWithLower?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,122 @@
...
+  @Test
+  public void ForeignKeySearchWithConstant() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
+    String sql = cql2pgJson.toSql("tableb.prefix == x0").getWhere();
+    // default pkColumnName is id without underscore
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE lower(tableb.jsonb->>'prefix') = lower('x0'))", sql);
+  }
+  @Test
+  public void ForeignKeySearchWithConstantOther() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
```
julianladisch: Can we use a better method name, for example searchWithFUnaccent?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -0,0 +1,122 @@
...
+  @Test
+  public void ForeignKeySearch() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
+    String sql = cql2pgJson.toSql("tableb.prefix == 11111111-1111-1111-1111-111111111111").getWhere();
+    // default pkColumnName is id without underscore
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE lower(tableb.jsonb->>'prefix') = lower('11111111-1111-1111-1111-111111111111'))", sql);
+  }
+  @Test
+  public void ForeignKeySearchWithConstant() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
```
julianladisch: Can we use a better method name, for example searchWithLower?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -0,0 +1,87 @@
+package org.z3950.zing.cql.cql2pgjson;
+
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.assertEquals;
```
julianladisch: Can this line been removed?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -24,16 +24,28 @@
...
    * @return SQL LIKE string without leading or trailing single quote
    */
   @SuppressWarnings("squid:S3776")  // suppress "Cognitive Complexity of methods should not be too high"
   public static String cql2like(String s) {
+    return cql2String(s,false);
+  
+  }
+  /**
+   * Convert a CQL string to an SQL string.
+   * CQL escapes * ? ^ \ and SQL LIKE escapes \ % _.
```
julianladisch: The comment about SQL LIKE should be moved to the sql2like method.

Here the comment should be "and SQL string constant escapes only the single quote ' by doubling it."

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
evaluk: I clarified my injection tests a bit, hopefully they make sense. Long story short is the tests are trying to achieve execution of code and they would fail if they succeeded.  Since they dont fail, they arent successfully injecting.  Thats the idea anyways.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
evaluk: I attempted to disable the duplication error in sonar cloud with suppress warning with no success.  I would be open to options.  I do not understand the readability issues as they exist here.  In this case the cql2like is a more restrictive version of the cql2string so it follows logically to call the functions in this way.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -75,13 +76,16 @@ public void uuidConstant() throws Exception {
     assertThat(cql("tableb.name == 33333333-3333-3333-3333-333333333333"), is(empty()));
   }
   @Test
-  public void ForeignKeySearchWithInjection1() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
-
+  public void ForeignKeySearchWithInjection1() {
+    //check to see if we can execute some arbitrary sql
     assertThat(cql("tableb.prefix == \"x0')));((('DROP tableb\""), containsInAnyOrder("test3") );
+    //then check to see if the drop table worked by checking to see if there is anything there
+    assertThat(cql("tableb.prefix == \"x0')));((('DROP tableb\"").size() > 0, is(true));
```
julianladisch: If there is some injection resulting in executing both the select statement and the drop statement then the select statement returns 0 results and the drop statement returns 0 results.

A more easy way to check for injection is one (= unbalanced) single quote that triggers an SQL parse error that will raise a database exception in java:

    assertThat(cql("tableb.prefix == '"), is(empty()));

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -97,67 +87,113 @@ public static String cql2string(String s, boolean quotedString) {
...
+  /**
+   * Convert a CQL string to an SQL string. and SQL string constant escapes only
+   * the single quote ' by doubling it.
+   *
+   * @param s CQL string without leading or trailing double quote
+   * @return SQL string without leading or trailing single quote
+   */
+  public static String cql2string(String s) throws QueryValidationException {
+    StringBuilder like = new StringBuilder();
+    boolean backslash = false; // previous character is backslash
```
julianladisch: Can you remove the misleading comment and add the same javadoc used in the other methods?

    /**
     * true if the previous character is an escaping backslash
     */

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -97,67 +87,113 @@ public static String cql2string(String s, boolean quotedString) {
...
 
+  /**
+   * Convert a CQL string to an SQL string. and SQL string constant escapes only
+   * the single quote ' by doubling it.
+   *
+   * @param s CQL string without leading or trailing double quote
+   * @return SQL string without leading or trailing single quote
+   */
+  public static String cql2string(String s) throws QueryValidationException {
+    StringBuilder like = new StringBuilder();
```
julianladisch: Can you rename the variable? For example rename `s` to `cql` and `like` to `s`.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -37,10 +37,7 @@ private Cql2SqlUtil() {
   @SuppressWarnings("squid:S3776")  // suppress "Cognitive Complexity of methods should not be too high"
   public static String cql2like(String s) {
     StringBuilder like = new StringBuilder();
-    /**
-     * true if the previous character is an escaping backslash
-     */
-    boolean backslash = false;
+    boolean backslash = false;  // previous character is backslash
```
julianladisch: Can you re-insert "escaping" before backslash?
Can you re-insert "true if" at the begin of the comment to show that the comment is not about "false".

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java
```diff
@@ -140,10 +137,7 @@ public static String cql2string(String s) throws QueryValidationException {
   @SuppressWarnings("squid:S3776")  // suppress "Cognitive Complexity of methods should not be too high"
   public static String cql2regexp(String s) {
     StringBuilder regexp = new StringBuilder();
-    /**
-     * true if the previous character is an escaping backslash
-     */
-    boolean backslash = false;
+    boolean backslash = false;  // previous character is backslash
```
julianladisch: Can you re-insert "escaping" before backslash?
Can you re-insert "true if" at the begin of the comment to show that the comment is not about "false".

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
julianladisch: There is no need to run the same CQL query twice.

PR #433, opened 2019-06-03
Rmb 380 relation modifiers array

adamdickmeiss:


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -756,6 +764,30 @@ private String pgId(CQLTermNode node) throws QueryValidationException {
...
+    vals.setIndexText(index2sqlText("t.c", relationModifier.getSubfield()));
+
+    final String table = this.jsonField.split("\\.")[0];
+    final String field = this.jsonField.split("\\.")[1];
+
+    res.append("id in (select t.id from (select id as id, jsonb_array_elements(");
+    res.append(field + "->'" + index + "') as c from " + table + ") as t where t.c");
+    res.append(" @> '{\"");
+    // pass the name from the schema rather than CQL because case is lost
+    res.append(Cql2SqlUtil.cql2string(relationModifier.getModifierName()));
```
julianladisch: This is a JSON string embedded in an SQL string. For the JSON string " needs to be masked.
We propably need Cql2SqlUtil.cql2JsonSqlString(String)

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -756,6 +764,30 @@ private String pgId(CQLTermNode node) throws QueryValidationException {
...
+    res.append("\": \"");
+    res.append(Cql2SqlUtil.cql2string(modifierValue));
```
julianladisch: This is a JSON string embedded in an SQL string. For the JSON string " needs to be masked.
We propably need Cql2SqlUtil.cql2JsonSqlString(String)

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -167,6 +168,7 @@ public void cql2pgJsonException(CQL2PgJSON cql2pgJson, String cql,
 
   @Test
   @Parameters({
+    "\"field'))@@to_tsquery(('english\"=x # ",
```
julianladisch: This is the basic(String) method, but this is not a basic test. Please move it into some other test method and add documentation if needed so that it is clear what class of input this tests (SQL injection in field name?)

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -766,16 +799,45 @@ private String pgId(CQLTermNode node) throws QueryValidationException {
...
+    DbIndex dbIndex = DbSchemaUtils.getDbIndex(dbTable, index);
+
+    CqlModifiers cqlModifiers = new CqlModifiers(node);
+    List<Modifier> relationModifiers = cqlModifiers.getRelationModifiers();
+    if (relationModifiers.isEmpty()) {
+      return indexNode(index, node, vals, dbIndex, cqlModifiers);
+    } else if (relationModifiers.size() > 1) {
+      throw new QueryValidationException("CQL: Only one @-relation modifier supported");
+    } else {
+      Modifier relationModifier = relationModifiers.get(0);
```
julianladisch: Using guard clauses reduces the indentation of the else-clause and the complexity, and increases readability:

    if (elationModifiers.isEmpty()) {
      return indexNode(index, node, vals, dbIndex, cqlModifiers);
    }
    if (relationModifiers.size() > 1) {
      throw new QueryValidationException("CQL: Only one @-relation modifier supported");
    }
    Modifier relationModifier = relationModifiers.get(0);

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: Not necessary. It's a quoted string so while one can make bad JSON it's still quoted (with single quotes) and it's therefore impossible to unquote that .

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: No

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
adamdickmeiss: Sure . Can make a separate test for that.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: OK. But I'll try to generalize it instead.. Was not really happy with only one relation modifier.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
julianladisch: or create

    @> json_build_object('someName', 'sameValue')

using

    res.append(" @> json_build_object('");
    res.append(Cql2SqlUtil.cql2string(relationModifier.getModifierName()));
    res.append("', '");
    res.append(Cql2SqlUtil.cql2string(modifierValue));

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -756,6 +765,30 @@ private String pgId(CQLTermNode node) throws QueryValidationException {
...
+
+    IndexTextAndJsonValues vals = new IndexTextAndJsonValues();
+    vals.setIndexText(index2sqlText("t.c", relationModifier.getSubfield()));
+
+    final String table = this.jsonField.split("\\.")[0];
+    final String field = this.jsonField.split("\\.")[1];
+
+    res.append("id in (select t.id from (select id as id, jsonb_array_elements(");
+    res.append(field + "->'" + index + "') as c from " + table + ") as t where t.c");
+    res.append(" @> '{\"");
```
julianladisch: This SQL is hard to read and understand.

How about something like this with indentation that follows the structure?

    String modifierName = Cql2SqlUtil.cql2string(relationModifier.getModifierName());
    ...
    String res =
        "id in (select t.id "
      + "  from (select id as id,"
      + "               jsonb_array_elements(" + index + ") as c"
      + "        from " + table ") as t"
      + "  where t.c @> '{\"" + modifierName + "\": \"" + modifierValue + "\"}'"
      + "        and " + indexNode
      + ")"
    );

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -756,6 +765,30 @@ private String pgId(CQLTermNode node) throws QueryValidationException {
     }
   }
 
+  private String arrayNode(String index, CQLTermNode node, org.folio.rest.persist.ddlgen.Modifier relationModifier,
+    String modifierValue, DbIndex dbIndex, CqlModifiers modifiers) throws QueryValidationException {
+
+    StringBuilder res = new StringBuilder();
```
julianladisch: The StringBuilder is not needed. Use String instead. The compiler will automatically use a StringBuilder in the background. Using String concatenation is much more easy to read than StringBuilder concatenation. https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -902,22 +904,42 @@ public void allAnyAdjFT(String testcase)
...
+    "contributors all \"contributorNametypeid 2b94c631-fca9-4892-a730-03ee529ffe2a\" # a",
+    "contributors all \"contributorNametypeid 2b94c631\" # a",
+    "contributors all \"contributorNametypeid 4892\" # a",
+    "contributors = \"contributorNameTypeId 2b94c631\" #",
+    "identifiers = \"value 0552142352\" # a",
+    "contributors = name           # a; b",
+    "name = terry                  #",
+    "contributors.name = terry     #",
+    "contributors.name = terry     #",
+    })
```
julianladisch: Can you add test cases for 1 level and 2 level nested array like personal.addresses and contactInformation.phone.mobile?

julianladisch: Can you add test cases for exact match comparison using ==, for example

    "contributors ==/@contributornametypeid==e8b311a6-3b21-43f2-a269-dd9310cb2d0a creator "B. Maya"

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -765,27 +765,58 @@ private String pgId(CQLTermNode node) throws QueryValidationException {
...
+              no++;
+              found = true;
+            }
+          }
+        }
+      }
+      if (!found) {
+        throw new QueryValidationException("CQL: Unsupported relation modifier " + relationModifier.getType());
+      }
+    }
```
julianladisch: The code for the right hand side of `@>` should be moved into a subfunction.

adamdickmeiss: It could and SonarQube will warn if the function is too complex.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
adamdickmeiss: Sure.. Something added in this.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: Tried it, but couldn't get it to work.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: ok. Something changed.

PR #440, opened 2019-06-11
Normalize schema ref path RMB-397

adamdickmeiss:


adamdickmeiss:
Before RMB 25 the ref handling also used normalize ..

julianladisch:
See https://github.com/folio-org/raml-module-builder/compare/rmb-397-regression-with-ref-paths for a list of unit tests.

adamdickmeiss:
Closing now that there are tests for separate branch

PR #441, opened 2019-06-11
Rmb 397 regression with ref paths

adamdickmeiss:


adamdickmeiss:
Removing a comment made the check passes. Illustrates that we have a testing problem .. RMB-398

PR #443, opened 2019-06-13
RMB-400: PgUtil should return 400 (not 500) on invalid UUID id

julianladisch:


julianladisch:
The only issue sonarcloud finds is this: 3.7% Duplicated Lines on New Code (is greater than 3%)
This makes the sonarcloud analysis fail. However, rewriting the code make it more difficult to read. I consider this a false positive.
Note that there is 100% line coverage and no other issues.

util/src/test/java/org/folio/util/UuidUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+    assertThat(UuidUtil.isLooseUuid(s), is(true));
+    assertThat(UuidUtil.isUuid(s), is(true));
+  }
+
+  @ParameterizedTest
+  @CsvSource({
+    "00000000-0000-0000-0000-000000000000, true",
+    "11111111-1111-1111-1111-111111111111, true",
+    "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa, true",
+    "ffffffff-ffff-ffff-ffff-ffffffffffff, true",
```
adamdickmeiss: why ", true" string? Isn't that redundant?

julianladisch: Yes, it is no longer used.
Thanks, removed!

PR #450, opened 2019-06-25
Rmb 387 implement parent child foreign key

evaluk:
update to allow parent-> child joins


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -542,7 +542,9 @@ private String subQuery(String index,CQLTermNode node, Table correlation) throws
     String[] termParts = node.getTerm().split("\\.");
 
     String [] foreignTarget = (idxParts.length > termParts.length) ? idxParts : termParts;
-    ForeignKeys fkey = findForeignKey(dbTable.getPkColumnName(),correlation);
+    boolean indexInTable = findForeignKey(dbTable.getPkColumnName(),correlation, dbTable) == null;
+    ForeignKeys fkey = indexInTable ? findForeignKey(PK_COLUMN_NAME,dbTable, correlation) : findForeignKey(dbTable.getPkColumnName(),correlation, dbTable);
```
julianladisch: This is called two times in the two lines, can this be avoided?

    findForeignKey(dbTable.getPkColumnName(),correlation, dbTable)

julianladisch: Can you add a space after each comma?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -577,12 +579,12 @@ private String subQuery(String index,CQLTermNode node, Table correlation) throws
...
+      selectString =  targetField ;
       inKeyword = myField + " IN ";
       whereClause = " WHERE " + indexString + " = " + termString;
     } else {
       inKeyword = myField + " IN ";
-      selectString = "Cast ( " + index2sqlText(foreignTableJsonb, foreignTarget[1]) + "as UUID)";
+      selectString = indexInTable ?  PK_COLUMN_NAME : "Cast ( " + index2sqlText(foreignTableJsonb,  foreignTarget[1]) + " as UUID)";
     }
 
     return inKeyword + " ( SELECT " + selectString + " from " + foreignTarget[0] + whereClause + ")";
```
julianladisch: The 9 string variables make it hard to follow how the SQL string is constructed and what the SQL query makes.

Can you at least extract the most easy case, for example like this?

    
    if (tableNamePattern.matcher(term).matches()) {
      if (indexInTable) {
        return myField + " IN ( SELECT id FROM " + foreignTarget[0] + ")";
      }
      return myField + " IN ( SELECT Cast ( " + index2sqlText(foreignTableJsonb,  foreignTarget[1]) + " as UUID)" 
            + " FROM " + foreignTarget[0] + ")";
    }

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -559,12 +561,12 @@ private String subQuery(String index,CQLTermNode node, Table correlation) throws
 
     boolean isTermConstant = !tableNamePattern.matcher(term).matches();
 
-    String myField = dbTable.getTableName() + ".id";
-    String targetField = index2sqlText(foreignTableJsonb, fkey.getFieldName());
+    String myField = !indexInTable ?  dbTable.getTableName() + "." + PK_COLUMN_NAME : "Cast ( " + index2sqlText(dbTable.getTableName() + ".jsonb",fkey.getFieldName()) + " as UUID)";
```
julianladisch: Using `Cast ( x as UUID )` is more difficult to read than `x::UUID`, can you change it here and in the following lines?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -612,11 +614,13 @@ private String getWrapTemplateWithSchemaDetection(String whereField,Table target
...
    * @return the ForeignKeys if found, null otherwise
    */
-  private ForeignKeys findForeignKey(String field, Table targetTable) {
-    String fieldName = dbTable.getTableName() + field;
-    for (ForeignKeys key : targetTable.getForeignKeys()) {
-      if (fieldName.equalsIgnoreCase(key.getFieldName())) {
-        return key;
+  private ForeignKeys findForeignKey(String field, Table targetTable, Table currentTable) {
+    String fieldName = currentTable.getTableName() + field;
+      if(targetTable.getForeignKeys() != null) {
```
julianladisch: Can you fix the indentation?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -612,11 +614,13 @@ private String getWrapTemplateWithSchemaDetection(String whereField,Table target
    * @param targetTable  where to search
    * @return the ForeignKeys if found, null otherwise
    */
-  private ForeignKeys findForeignKey(String field, Table targetTable) {
-    String fieldName = dbTable.getTableName() + field;
-    for (ForeignKeys key : targetTable.getForeignKeys()) {
-      if (fieldName.equalsIgnoreCase(key.getFieldName())) {
-        return key;
+  private ForeignKeys findForeignKey(String field, Table targetTable, Table currentTable) {
```
julianladisch: Can you change to natural order of the parameters: Table currentTable, String field, Table targetTable?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
julianladisch: Can you make it a guard clause that reduces the indentation of the following block and therefore improves readability?

    if (targetTable.getForeignKeys() == null) {
      return null;
    }
    for (ForeignKeys key : …

julianladisch: Can you add a space after `if`?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -22,28 +22,35 @@ public void ForeignKeySearchNumeric() throws FieldException, QueryValidationExce
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
     cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
     String sql = cql2pgJson.toSql("tableb.blah == /number 123452").getWhere();
-    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE (tableb.jsonb->>'blah')::NUMERIC = ('123452')::NUMERIC)", sql);
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId' as UUID) from tableb WHERE (tableb.jsonb->>'blah')::NUMERIC = ('123452')::NUMERIC)", sql);
```
julianladisch: This space change shows that this kind of unit tests is hard to maintain.

Instead all unit tests in ForeignKeyGenerationTest.java that do not check thrown exceptions should be converted into 
ForeignKeyGenerationIT.java tests that check the functionality and not how the SQL in constructed.

julianladisch:
The title and git branch name say that this implements parent->child direction. Is this true?

cql2pgjson/src/test/resources/foreignKey.sql
```diff
@@ -1,15 +1,18 @@
-DROP TABLE IF EXISTS tablea, tableb, tablec;
+DROP TABLE IF EXISTS tablea, tableb, tablec, tabled;
 CREATE EXTENSION IF NOT EXISTS unaccent WITH SCHEMA public;
 CREATE OR REPLACE FUNCTION f_unaccent(text) RETURNS text AS $$
   SELECT public.unaccent('public.unaccent', $1);
 $$ LANGUAGE sql IMMUTABLE;
-CREATE TABLE tablea (id UUID PRIMARY KEY, jsonb JSONB NOT NULL);
+CREATE TABLE tabled (id UUID PRIMARY KEY, jsonb JSONB NOT NULL);
+CREATE TABLE tablea (id UUID PRIMARY KEY, jsonb JSONB NOT NULL, tabledId UUID references tabled);
```
julianladisch: This test setup intendes to have foreign key pointers from tablec -> tableb -> tablea.

For testing child -> parent foreign key we should use tableb and make a CQL query using tablea.*. Can you remove tablea.tabledId and tabled?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -67,17 +73,24 @@ public void foreignKeyFilter1() throws Exception {
     assertThat(cql("id == tableb.tableaId"), containsInAnyOrder("test1", "test2", "test3"));
   }
 
+  @Test
+  public void foreignKeyFilterParentChild() throws Exception {
+    assertThat(cql("id ==  tabled.tabledId"), containsInAnyOrder("test0", "test1","test2", "test3"));
```
julianladisch: `tabled.tabledId` does not exists. What does this intend to test?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -50,6 +51,11 @@ public void foreignKeySearch0() throws Exception {
     assertThat(cql("tableb.prefix == x0"), is(empty()));
   }
 
+  @Test
+  public void foreignKeySearchParentChild() throws Exception {
+    assertThat(cql("tabled.prefix == a1"), containsInAnyOrder("test0", "test1"));
```
julianladisch: You may want to rename cql(String) to cqla(String) and create a new cqlb(String):

    private List<String> cqlb(String cql) {
      try {
        String sql = cql2pgJson.toSql(cql).toString();
        return firstColumn("select jsonb->>'name' from tableb " + sql);
      } catch (QueryValidationException e) {
        throw new RuntimeException(e);
      }
    }

Then you can query the other direction using tableb as the current table and use tableb.* in CQL.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -22,28 +22,35 @@ public void ForeignKeySearchNumeric() throws FieldException, QueryValidationExce
...
   }
   @Test
```
julianladisch: Can you insert a blank line before `@Test`?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -22,28 +22,35 @@ public void ForeignKeySearchNumeric() throws FieldException, QueryValidationExce
...
   public void ForeignKeySearch() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
     cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
     String sql = cql2pgJson.toSql("tableb.prefix == 11111111-1111-1111-1111-111111111111").getWhere();
-    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE lower(tableb.jsonb->>'prefix') = lower('11111111-1111-1111-1111-111111111111'))", sql);
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId' as UUID) from tableb WHERE lower(tableb.jsonb->>'prefix') = lower('11111111-1111-1111-1111-111111111111'))", sql);
+  }
+  @Test
```
julianladisch: Can you insert a blank line before `@Test`?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -22,28 +22,35 @@ public void ForeignKeySearchNumeric() throws FieldException, QueryValidationExce
...
+  public void ForeignKeySearchParentChild() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
+    String sql = cql2pgJson.toSql("tabled.prefix == 11111111-1111-1111-1111-111111111111").getWhere();
+    assertEquals("Cast ( tablea.jsonb->>'tabledId' as UUID) IN  ( SELECT id from tabled WHERE lower(tabled.jsonb->>'prefix') = lower('11111111-1111-1111-1111-111111111111'))", sql);
   }
   @Test
```
julianladisch: Can you insert a blank line before `@Test`?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -22,28 +22,35 @@ public void ForeignKeySearchNumeric() throws FieldException, QueryValidationExce
...
   public void ForeignKeySearchWithLowerConstant() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
     cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
     String sql = cql2pgJson.toSql("tableb.prefix == x0").getWhere();
-    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE lower(tableb.jsonb->>'prefix') = lower('x0'))", sql);
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId' as UUID) from tableb WHERE lower(tableb.jsonb->>'prefix') = lower('x0'))", sql);
   }
   @Test
```
julianladisch: Can you insert a blank line before `@Test`?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -22,28 +22,35 @@ public void ForeignKeySearchNumeric() throws FieldException, QueryValidationExce
...
   public void ForeignKeySearchWithFUnaccent() throws FieldException, QueryValidationException, ServerChoiceIndexesException {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea.json" );
     cql2pgJson.setDbSchemaPath("templates/db_scripts/foreignKey.json");
     String sql = cql2pgJson.toSql("tableb.otherindex == y0").getWhere();
-    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId'as UUID) from tableb WHERE f_unaccent(tableb.jsonb->>'otherindex') = f_unaccent('y0'))", sql);
+    assertEquals("tablea.id IN  ( SELECT Cast ( tableb.jsonb->>'tableaId' as UUID) from tableb WHERE f_unaccent(tableb.jsonb->>'otherindex') = f_unaccent('y0'))", sql);
   }
   @Test
```
julianladisch: Can you insert a blank line before `@Test`?

cql2pgjson/src/test/resources/foreignKey.sql
```diff
@@ -35,16 +39,31 @@ END; $$ language 'plpgsql';
...
+  NEW.id       = NEW.jsonb->>'id';
+  
+  RETURN NEW;
+END; $$ language 'plpgsql';
+CREATE TRIGGER update_tabled_references BEFORE INSERT OR UPDATE ON tabled
+  FOR EACH ROW EXECUTE PROCEDURE update_tabled_references();
+INSERT INTO tabled (jsonb) VALUES
+('{"id": "D1111111-1111-1111-1111-111111111111", "prefix": "a1", "otherindex": "z1"}'),
+('{"id": "D2222222-2222-2222-2222-222222222222", "prefix": "a2", "otherindex": "z2"}'),
+('{"id": "D3333333-3333-3333-3333-333333333333", "prefix": "a2", "otherindex": "z3"}');
```
julianladisch: tabled INSERT should go after tablec INSERT.

cql2pgjson/src/test/resources/templates/db_scripts/foreignKey.json
```diff
@@ -1,7 +1,20 @@
...
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
+        }
+      ], "foreignKeys": [
+        {
+          "fieldName": "tabledId",
+          "targetTable": "tabled"
+        }
+      ]
```
julianladisch: Can you avoid having a foreign key in tablea and instead use the existing foreign key from tableb to tablea? Reusing the same foreign key relation makes it more easy to follow the child->parent and parent->child tests.

cql2pgjson/src/test/resources/foreignKey.sql
```diff
@@ -35,16 +39,31 @@ END; $$ language 'plpgsql';
...
+('{"id": "D1111111-1111-1111-1111-111111111111", "prefix": "a1", "otherindex": "z1"}'),
+('{"id": "D2222222-2222-2222-2222-222222222222", "prefix": "a2", "otherindex": "z2"}'),
+('{"id": "D3333333-3333-3333-3333-333333333333", "prefix": "a2", "otherindex": "z3"}');  
+
 INSERT INTO tablea (jsonb) VALUES
-('{"id": "A0000000-0000-0000-0000-000000000000", "name": "test0"}'),
-('{"id": "A1111111-1111-1111-1111-111111111111", "name": "test1"}'),
-('{"id": "A2222222-2222-2222-2222-222222222222", "name": "test2"}'),
-('{"id": "A3333333-3333-3333-3333-333333333333", "name": "test3"}');
-INSERT INTO tablea (jsonb) VALUES (jsonb_build_object('id', md5('a' || generate_series(1, 2000)::text)));
```
julianladisch: Using EXPLAIN ANALYSE shows what indexes postgresql uses. For that the additional records are needed, if there are only a few records then postgresql uses a full table scan.

cql2pgjson/src/test/resources/foreignKey.sql
```diff
@@ -35,16 +39,31 @@ END; $$ language 'plpgsql';
...
+('{"id": "A2222222-2222-2222-2222-222222222222", "name": "test2","tabledId": "D2222222-2222-2222-2222-222222222222"}'),
+('{"id": "A3333333-3333-3333-3333-333333333333", "name": "test3","tabledId": "D3333333-3333-3333-3333-333333333333"}'),
+('{"id": "A4444444-4444-4444-4444-444444444444", "name": "test4"}');
 
 INSERT INTO tableb (jsonb) VALUES
 ('{"id": "B1111111-1111-1111-1111-111111111111", "prefix": "x1", "otherindex": "y1","tableaId": "A1111111-1111-1111-1111-111111111111"}'),
 ('{"id": "B2222222-2222-2222-2222-222222222222", "prefix": "x2", "otherindex": "y2", "tableaId": "A2222222-2222-2222-2222-222222222222"}'),
 ('{"id": "B3333333-3333-3333-3333-333333333333", "prefix": "x2", "otherindex": "y3","tableaId": "A2222222-2222-2222-2222-222222222222"}'),
 ('{"id": "B4444444-4444-4444-4444-444444444444", "prefix" : "x0'')));(((''DROP tableb","otherindex": "y4", "tableaId": "A3333333-3333-3333-3333-333333333333"}');
-INSERT INTO tableb (jsonb) VALUES (jsonb_build_object('id', md5('b' || generate_series(1, 2000)::text)));
```
julianladisch: Using EXPLAIN ANALYSE shows what indexes postgresql uses. For that the additional records are needed, if there are only a few records then postgresql uses a full table scan.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
evaluk: this is intended to test the filter functionality, though I do agree that it looks weird.  This may need to look like this instead but it fails my unit tests:
tablea.tableaId == id.  But I need to think about it more

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
evaluk: I am uncertain about this.  I still feel like these unit tests are useful to guide development and specify what sql is used.  Otherwise we would be debugging changes directly from the sql exceptions and syntax.  This seems like not that great a situation.  I used and am using these extensively to test my code, so why would they not be useful in the future?

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationIT.java
evaluk: This unit test is no longer valid.  It will need to be removed in a future update

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
evaluk: I removed a lot of duplication in this code block.  I am not sure this is needed anymore.

cql2pgjson/src/test/java/org/z3950/zing/cql/cql2pgjson/ForeignKeyGenerationTest.java
julianladisch: I understand. You need these test because subQuery method is too difficult to understand because you cannot see the complete SQL string, you can only see the complete SQL string in these tests.
However, then we need to fix subQuery method to be more easy to understand.
In this case most duplicate code was removed by extracting them into variables but readability suffered. Readability is more important than avoiding code duplication.

Two days ago I proposed this:

    if (tableNamePattern.matcher(term).matches()) {
      if (indexInTable) {
        return myField + " IN ( SELECT id FROM " + foreignTarget[0] + ")";
      }
      return myField + " IN ( SELECT Cast ( " + index2sqlText(foreignTableJsonb,  foreignTarget[1]) + " as UUID)" 
            + " FROM " + foreignTarget[0] + ")";
    }

This can be combined with splitting into a parentChild and a childParent method that handle the two indexInTable cases.

PR #453, opened 2019-07-02
Full text match for array modifiers RMB-416

adamdickmeiss:
Array relation modifiers without value (and comparator) are now
allowed and will override modifiersSubfield setting. Multiple
relation modifiers without value are OR'ed together.
Multiple relation modifiers with value are ANDed together and are using
full-text queries. There were using array contains before (@>).


julianladisch:
The syntax `/@type1=value1` should be changed to `/@type1==value1` because it is an exact match. For consistency.

cql2pgjson/README.md
```diff
@@ -183,20 +183,33 @@ The type1, type2 and subfield must all be defined in schema.json, because
...
+      "fieldName": "identifiers",
+      "tOps": "ADD",
+      "arraySubfield": "value",
+      "arrayModifiers": ["identifierTypeId"]
+    }
+
+This will allow you to perform searches, such as:
+
+    identifiers = /@identifierTypeId=7e591197-f335-4afb-bc6d-a6d76ca3bace 6316800312
+
```
julianladisch: Can you reorganize the "Matching array elements" section? First show the identifiers example, then explain the `/@` modifier, and at the end present how to match the complete array as a string.

adamdickmeiss:
> The syntax `/@type1=value1` should be changed to `/@type1==value1` because it is an exact match. For consistency.

The matching is fulltext , not exact anymore. So that's why = is appropriate. If we really want exact match - which I doubt - then we can add == operator later.

cql2pgjson/README.md
adamdickmeiss: I'm not sure what you mean

adamdickmeiss:
> The syntax `/@type1=value1` should be changed to `/@type1==value1` because it is an exact match. For consistency.

I can that the documentation is not right with this, I'll fix that.

PR #454, opened 2019-07-02
RMB-414: audit table jsonb, rename item field to $auditingFieldName

julianladisch:
* Add `auditingTableName` and `auditingFieldName` properties to schema.json's table section.
* Add maven submodule `domain-models-runtime-it` for domain-models-runtime integration tests.
* Remove `dbTemplate.schema` that is outdated and not used.


julianladisch:
Squashed: #455

PR #461, opened 2019-07-03
RMB-419: Rename table user -> users, pass table from PgUtil.get to CQL2PgJSON

julianladisch:


julianladisch:
Yes, thanks for finding this omission!

PR #464, opened 2019-07-05
Adjust count query RMB-422

adamdickmeiss:
This fixes the code to dig out the query for "Counting". The original query returned by CQL2pgJSON is corrupted. Actually, re-parsing SQL is a very bad idea. Eventually, the Cql2PG should return a "SQL" object where the where-clause (WITHOUT "where") and sorting are separate entities so they can be shuffled around for later usage.


adamdickmeiss:
2 ouf of 6 lines not covered in PostgresClient (and that's a SQL message that I just adjusted!!)

PR #470, opened 2019-07-16
RMB-430: Generate audit id for concurrent queries.

julianladisch:


domain-models-runtime/src/main/resources/templates/db_scripts/audit.ftl
```diff
@@ -10,29 +10,28 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.audit_${table.tableName}_
...
-      -- UUID variant byte
-      seed = overlay(seed placing '8' from 17);
-      maxid = seed::uuid;
-    ELSE
-      maxid = ${myuniversity}_${mymodule}.next_uuid(maxid);
-    END IF;
+    jsonb = CASE WHEN TG_OP = 'DELETE' THEN OLD.jsonb ELSE NEW.jsonb END;
+
+    -- create uuid based on the jsonb value so that concurrent updates of different records are possible.
+    uuidtext = md5(jsonb::text);
```
hjiebsco: Any performance concern to run md5 for every record?

domain-models-runtime/src/main/resources/templates/db_scripts/audit.ftl
```diff
@@ -10,29 +10,28 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.audit_${table.tableName}_
...
+    -- UUID version byte
+    uuidtext = overlay(uuidtext placing '4' from 13);
+    -- UUID variant byte
+    uuidtext = overlay(uuidtext placing '8' from 17);
+    uuid = uuidtext::uuid;
+    -- If uuid is already in use increment until an unused is found. This can only happen if the jsonb content
+    -- is exactly the same. This should be very rare when it includes a timestamp.
+    WHILE EXISTS (SELECT 1 FROM ${myuniversity}_${mymodule}.${table.auditingTableName} WHERE id = uuid) LOOP
+      uuid = ${myuniversity}_${mymodule}.next_uuid(uuid);
```
hjiebsco: If concurrency happens here, will they get the same uuid as well (the new uuid will not be saved/committed yet)?

domain-models-runtime/src/main/resources/templates/db_scripts/audit.ftl
julianladisch: create table t (a int);

    insert into t values (generate_series(1, 1000000));
    INSERT 0 1000000
    Time: 1157.428 ms (00:01.157)

    insert into t values (length(md5('' || generate_series(1, 1000000))));
    INSERT 0 1000000
    Time: 1667.770 ms (00:01.668)

About half a nanosecond. This is fast enough.

domain-models-runtime/src/main/resources/templates/db_scripts/audit.ftl
julianladisch: This can only happen if the two transactions try to alter the same record. Then one of them will rollback even if there is no audit table.
RMB-430 is to fix the case where the two concurrent transactions alter different records.

PR #487, opened 2019-08-06
RMB-395: support nested sub queries across multiple tables

hjiebsco:
The implemented approach uses foreignKeys info to construct subqueries across multiple tables (one-way, either child -> parent or parent -> child, not mixed). Added optional tableAlias and targetTableAlias attributes to foreignKeys to address potential naming collision.

API call to **`/instance?query=item.barcode == 7834324634`** will generate nested subquery like below. A quick test in perf env shows the execution time is **under 1 ms**.
```
WHERE instance.id IN  ( SELECT (holdings_record.jsonb->>'instanceId')::UUID from holdings_record WHERE holdings_record.id IN  ( SELECT (item.jsonb->>'holdingsRecordId')::UUID from item WHERE lower(f_unaccent(item.jsonb->>'barcode')) LIKE lower(f_unaccent('7834324634'))))
```
```
folio=# explain analyze select * from instance WHERE instance.id IN  ( SELECT (holdings_record.jsonb->>'instanceId')::UUID from holdings_record WHERE holdings_record.id IN  ( SELECT (item.jsonb->>'holdingsRecordId')::UUID from item WHERE lower(f_unaccent(item.jsonb->>'barcode')) LIKE lower(f_unaccent('7834324634'))));
                                                                        QUERY PLAN                                                                        
----------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=17.61..20.02 rows=1 width=882) (actual time=0.114..0.115 rows=1 loops=1)
   ->  HashAggregate  (cost=17.18..17.19 rows=1 width=350) (actual time=0.103..0.103 rows=1 loops=1)
         Group Key: ((holdings_record.jsonb ->> 'instanceId'::text))::uuid
         ->  Nested Loop  (cost=9.14..17.17 rows=1 width=350) (actual time=0.101..0.102 rows=1 loops=1)
               ->  HashAggregate  (cost=8.71..8.72 rows=1 width=592) (actual time=0.084..0.084 rows=1 loops=1)
                     Group Key: ((item.jsonb ->> 'holdingsRecordId'::text))::uuid
                     ->  Index Scan using item_barcode_idx on item  (cost=0.43..8.71 rows=1 width=592) (actual time=0.081..0.081 rows=1 loops=1)
                           Index Cond: (lower(public.f_unaccent((jsonb ->> 'barcode'::text))) = '7834324634'::text)
                           Filter: (lower(public.f_unaccent((jsonb ->> 'barcode'::text))) ~~ '7834324634'::text)
               ->  Index Scan using holdings_record_pkey on holdings_record  (cost=0.44..8.46 rows=1 width=366) (actual time=0.010..0.010 rows=1 loops=1)
                     Index Cond: (id = ((item.jsonb ->> 'holdingsRecordId'::text))::uuid)
   ->  Index Scan using instance_pkey on instance  (cost=0.44..2.83 rows=1 width=882) (actual time=0.008..0.008 rows=1 loops=1)
         Index Cond: (id = ((holdings_record.jsonb ->> 'instanceId'::text))::uuid)
 Planning time: 3.439 ms
 Execution time: 0.229 ms
(15 rows)
```
Note, I based off this PR on another branch RMB-411_Implement_poLine_equals_star
that is supposed to be merged soon. **The actual code change for this PR is in the last commit**.


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -539,14 +540,33 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
-      return subQuery(node.getIndex(), node, indexTable);
+
+//  //determine if this table is real by checking in schema
+//  //determine right here whether this node deals with a foreign term on right or left side
+//  Table indexTable = checkForForeignLocationOfIndex(node);
+//  if (indexTable != null) {
+//    //we are doing a foreign key search
+//    //determine foreign key linking tables
+//    return subQuery(node.getIndex(), node, indexTable);
+//  }
```
julianladisch: Can you remove this commented-out code? We have git history if we want to restore.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -123,4 +126,24 @@ public void invalidCurrentTable() throws Exception {
...
   }
 
+  @Test
+  public void testSearchInstanceByItemBarcode() throws Exception {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("instance");
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/subquery.json");
+    String sql = cql2pgJson.toSql("item.barcode == 7834324634").toString();
+    String expected = "WHERE instance.id IN  ( SELECT (holdings_record.jsonb->>'instanceId')::UUID from holdings_record WHERE holdings_record.id IN  ( SELECT (item.jsonb->>'holdingsRecordId')::UUID from item WHERE lower(f_unaccent(item.jsonb->>'barcode')) LIKE lower(f_unaccent('7834324634'))))";
+    assertEquals(expected, sql);
+    System.out.println(sql);
```
julianladisch: Can you remove the println? This is redundant because it either matched expected or is reported by assertEquals.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -123,4 +126,24 @@ public void invalidCurrentTable() throws Exception {
...
+  }
+
+  @Test
+  public void testSearchItemByInstanceTitle() throws Exception {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("item");
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/subquery.json");
+    String sql = cql2pgJson.toSql("instance.title = 'Olmsted in Chicago'").toString();
+    String expected = "WHERE (item.jsonb->>'holdingsRecordId')::UUID IN  ( SELECT id from holdings_record WHERE (holdings_record.jsonb->>'instanceId')::UUID IN  ( SELECT id from instance WHERE to_tsvector('simple', lower(f_unaccent(instance.jsonb->>'title'))) @@ to_tsquery('simple', lower(f_unaccent('Olmsted<->in<->Chicago''')))))";
+    assertEquals(expected, sql);
+    System.out.println(sql);
```
julianladisch: Can you remove the println? This is redundant because it either matched expected or is reported by assertEquals.

cql2pgjson/src/main/java/org/folio/cql2pgjson/model/DbFkInfo.java
```diff
@@ -0,0 +1,33 @@
+package org.folio.cql2pgjson.model;
+
+public class DbFkInfo {
```
julianladisch: Can you rename it to a self-explanatory name, for example ForeignKey, avoiding the abbreviation Fk?
Is Db redundant because foreign keys are always in the db?
Is Info redundant because all classes with class variables contain info?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -43,6 +47,13 @@ private static boolean checkDbIndex(String cqlIndex, List<Index> indexes) {
     return getIndex(cqlIndex, indexes) != null;
   }
 
+  /**
+   * Get {@link Index} for given cqlIndex name.
+   *
+   * @param cqlIndex
+   * @param indexes
+   * @return
```
julianladisch: Can you document what is returned if not found?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -560,6 +580,55 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+    vals.setIndexJson(index2sqlJson(foreignTableJsonb, foreignTarget[1]));
+    vals.setIndexText(index2sqlText(foreignTableJsonb, foreignTarget[1]));
+
+    CqlModifiers cqlModifiers = new CqlModifiers(node);
+    String indexField = foreignTarget[1];
+    return indexNode(indexField, targetTable, node, vals, cqlModifiers);
+  }
+
   //method to determine if the right side of the node is a foreign term and returns the table it is attached to is so
   private Table checkForForeignLocationOfIndex(CQLTermNode node) {
```
julianladisch: Can you remove this unused method?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -588,7 +657,7 @@ private String subQuery(String index, CQLTermNode node, Table correlation) throw
     boolean indexInTable =  childParentForeignKey == null;
     ForeignKeys fkey = indexInTable ? findForeignKey(correlation, PK_COLUMN_NAME, dbTable) : childParentForeignKey;
     Table indexTable = indexInTable ? dbTable : correlation;
-    
+
```
julianladisch: Can you remove this unused method `subQuery`?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -136,7 +137,7 @@ public CQL2PgJSON(List<String> fields) throws FieldException {
    * @param fields Field names of the JSON fields, may include schema and table name (e.g. tenant1.user_table.json).
    *  Must conform to SQL identifier requirements (characters, not a keyword), or properly quoted using double quotes.
    *  The first field name on the list will be the default field for terms in queries that don't specify a json field.
-   * @param viewTarget the table that this is actually targetting since views hide the fact of the original table.  This is used to determine how to find indexes and other table specific info thats lost. 
+   * @param viewTarget the table that this is actually targetting since views hide the fact of the original table.  This is used to determine how to find indexes and other table specific info thats lost.
```
julianladisch: Can you also wrap the line to avoid excessive line length?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -539,14 +540,33 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+//    //we are doing a foreign key search
+//    //determine foreign key linking tables
+//    return subQuery(node.getIndex(), node, indexTable);
+//  }
+
+    // determine if index is in a foreign table
+    if (dbTable != null) {
+      String srcTabName = dbTable.getTableName();
+      String targetTabAlias = node.getIndex().split("\\.")[0];
+      List<DbFkInfo> fks = DbSchemaUtils.findForeignKeysFromSourceTableToTargetAlias(dbSchema, srcTabName, targetTabAlias);
```
julianladisch: Can you rename `fks` to a self-explanatory name, for example `foreignKeys`?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -560,6 +580,55 @@ private String pg(CQLTermNode node) throws QueryValidationException {
     return index2sql(node.getIndex(), node);
   }
 
+  private String pgSubQuery(CQLTermNode node, List<DbFkInfo> fks, boolean childToParent) throws QueryValidationException {
```
julianladisch: Can you rename fks to a self-explanatory name, for example foreignKeys?
Can you add a javadoc comment about childToParent including the meaning of false?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -560,6 +580,55 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+    String [] foreignTarget = node.getIndex().split("\\.", 2);
+    sb.append(indexNodeForForeignTable(node, targetTable, foreignTarget));
+    for (int i = 0; i < fks.size(); i++) {
+      sb.append(")");
+    }
+    return sb.toString();
+  }
+
+  private String indexNodeForForeignTable(CQLTermNode node, Table targetTable, String[] foreignTarget)
+      throws QueryValidationException {
```
julianladisch: Can you add a test to ForeignKeyGenerationIT that ensures that this also works for nested fields?
For example
* 1 nesting level like in user: personal.lastName https://s3.amazonaws.com/foliodocs/api/mod-users/users.html#users_post
* 2 nesting levels like in organisations: edi.ediFtp.username https://s3.amazonaws.com/foliodocs/api/mod-organizations-storage/organization.html#organizations_storage_organizations_post
Do not create user and organisations schemas, create such fields in one of our existing unit test schemas.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
     return null;
   }
 
+  /**
+   * Get {@link Table} for given table name in {@link Schema}
+   *
+   * @param dbSchema
+   * @param tableName
+   * @return
```
julianladisch: Can you document what is returned if not found?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+  }
+
+  /**
+   * Find a list of {@link ForeignKeys} from source table to target table alias
+   *
+   * @param dbSchema
+   * @param srcTabName
+   * @param targetTabAlias
+   *
+   * @return a list of {@link ForeignKeys}
```
julianladisch: Can you document what is returned if not found?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+    }
+  }
+
+  /**
+   * Find a list of {@link ForeignKeys} from source table alias and target table.
+   *
+   * @param dbSchema
+   * @param srcTabAlias
+   * @param targetTabName
+   * @return
```
julianladisch: Can you document what is returned if nothing found?

cql2pgjson/src/test/java/org/folio/cql2pgjson/util/DbSchemaUtilsTest.java
```diff
@@ -70,4 +75,38 @@ public void testGetDbIndexForNullTable() {
...
+    assertEquals("e", list.get(0).getTable());
+    assertEquals("d", list.get(1).getTable());
+    assertEquals("c", list.get(2).getTable());
+
+    // source alias 2
+    list = DbSchemaUtils.findForeignKeysFromSourceAliasToTargetTable(dbSchema, "e2Alias", "b");
+    list.forEach(fk -> System.out.println(fk));
+    assertEquals("e", list.get(0).getTable());
+    assertEquals("c", list.get(1).getTable());
+  }
```
julianladisch: Can you make all the list asserts more easy to read?

    List<String> list = tableNames(DbSchemaUtils.findForeignKeysFromSourceTableToTargetAlias(dbSchema, "f", "a"));
    assertThat(list, contains("f", "e", "c", "b"));

You only need to write tableNames and add

    import static org.hamcrest.collection.IsIterableContainingInOrder.contains;

cql2pgjson/src/test/java/org/folio/cql2pgjson/util/DbSchemaUtilsTest.java
```diff
@@ -70,4 +75,38 @@ public void testGetDbIndexForNullTable() {
...
+
+    // source alias
+    list = DbSchemaUtils.findForeignKeysFromSourceAliasToTargetTable(dbSchema, "eAlias", "b");
+    assertEquals("e", list.get(0).getTable());
+    assertEquals("d", list.get(1).getTable());
+    assertEquals("c", list.get(2).getTable());
+
+    // source alias 2
+    list = DbSchemaUtils.findForeignKeysFromSourceAliasToTargetTable(dbSchema, "e2Alias", "b");
+    list.forEach(fk -> System.out.println(fk));
```
julianladisch: Can you remove this line? assert will report any mismatch.

cql2pgjson/src/test/resources/templates/db_scripts/subquery.json
```diff
@@ -0,0 +1,27 @@
+{
```
julianladisch: Can you rename this file from subquery.json to foreignKeyMultiTable.json to be self-explanatory and to better match the other .json file names?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: I'll have no object to short local names

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+   */
+  public static List<DbFkInfo> findForeignKeysFromSourceAliasToTargetTable(Schema dbSchema, String srcTabAlias,
```
julianladisch: Can you rename the long method name to findSourceAliasForeignKeys and put the details into the javadoc?
Can you extend the javadoc explaining which is child and which is parent table?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+  /**
+   * Find a list of {@link ForeignKeys} from source table to target table alias
+   *
+   * @param dbSchema
+   * @param srcTabName
+   * @param targetTabAlias
+   *
+   * @return a list of {@link ForeignKeys}
+   */
+  public static List<DbFkInfo> findForeignKeysFromSourceTableToTargetAlias(Schema dbSchema, String srcTabName,
```
julianladisch: Can you rename the long method name to findTargetAliasForeignKeys and put the details into the javadoc?
Can you extend the javadoc explaining which is child and which is parent table?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+   * @param tableName
+   * @return
+   */
+  public static Table getTable(Schema dbSchema, String tableName) {
+    for (Table table : dbSchema.getTables()) {
+      if (tableName.equals(table.getTableName())) {
+        return table;
+      }
+    }
+    return null;
```
julianladisch: Please add a unit test that checks the `return null` line.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -560,6 +580,55 @@ private String pg(CQLTermNode node) throws QueryValidationException {
...
+
+    String foreignTableJsonb = targetTable.getTableName() + ".jsonb";
+
+    IndexTextAndJsonValues vals = new IndexTextAndJsonValues();
+    if (jsonField == null) {
+      vals = multiFieldProcessing(foreignTarget[1]);
```
julianladisch: Can you add a unit test for this line?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+   */
+  public static List<DbFkInfo> findForeignKeysFromSourceTableToTargetAlias(Schema dbSchema, String srcTabName,
+      String targetTabAlias) {
+    return findForeignKeys(dbSchema, srcTabName, targetTabAlias, true);
+  }
+
+  private static List<DbFkInfo> findForeignKeys(Schema dbSchema, String srcTabName, String targetTabName,
+      boolean useTargetAlias) {
+    List<DbFkInfo> list = new ArrayList<>();
+    Table srcTab = getTable(dbSchema, srcTabName);
```
julianladisch: Can you add a unit test for findForeignKeys where getTable returns null?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: done

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
adamdickmeiss: done

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
adamdickmeiss: done

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: done

cql2pgjson/src/main/java/org/folio/cql2pgjson/model/DbFkInfo.java
adamdickmeiss: I'll leave it as is.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: done

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: done

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: leaving as is.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: done

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: done

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: done

cql2pgjson/src/test/java/org/folio/cql2pgjson/util/DbSchemaUtilsTest.java
adamdickmeiss: leaving as is

cql2pgjson/src/test/java/org/folio/cql2pgjson/util/DbSchemaUtilsTest.java
adamdickmeiss: done

cql2pgjson/src/test/resources/templates/db_scripts/subquery.json
adamdickmeiss: leaving as is.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: leaving this as is

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: leaving as is for now

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
adamdickmeiss: I think it's worse than that.. If a test was added we'd see that vals contents is modified in following lines anyway.. So this is clearly a bug. jsonField == null when in view mode and we aren't using a view anyway when we are targeting a table AFAIK; so I think the whole if+body should be removed.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: Added a guard for this for now

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+    }
+    // direct FK
+    for (ForeignKeys fk : srcTab.getForeignKeys()) {
+      String targetName = useTargetAlias ? fk.getTargetTableAlias() : fk.getTargetTable();
+      if (targetTabName.equals(targetName)) {
+        list.add(new DbFkInfo(srcTab.getTableName(), fk.getFieldName(), fk.getTargetTable()));
+        return list;
+      }
+    }
+    // find the shortest path
```
julianladisch: Please explain which one is chosen when two paths have the same length:
tableA -> tableX -> tableB
tableA -> tableY -> tableB

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+      }
+      for (ForeignKeys fk : table.getForeignKeys()) {
+        String tabAlias = fk.getTableAlias() == null ? table.getTableName() : fk.getTableAlias();
+        if (srcTabAlias.equals(tabAlias)) {
+          // direct FK
+          if (targetTabName.equals(fk.getTargetTable())) {
+            list.add(new DbFkInfo(table.getTableName(), fk.getFieldName(), fk.getTargetTable()));
+            return list;
+          } else {
+            // find the shortest path
```
julianladisch: Please explain which one is chosen when two paths have the same length:
tableA -> tableX -> tableB
tableA -> tableY -> tableB

cql2pgjson/src/main/java/org/folio/cql2pgjson/model/DbFkInfo.java
julianladisch: org.folio.rest.persist.ddlgen.Index parallels
org.folio.cql2pgjson.model.DbIndex

org.folio.rest.persist.ddlgen.ForeignKeys
parallels
org.folio.cql2pgjson.model.DbForeignKey
but not
org.folio.cql2pgjson.model.DbFkInfo

Can you explain why we want this inconsistency?

README.md
```diff
@@ -1076,14 +1071,12 @@ If the search term is a number then a numeric mode is used for "==", "<>", "<",
...
-* the join desired index must be only 1 table deep
+* The join desired index must be only 1 table deep
   - e.g.  table1 -> table2 not table1 -> table2 -> table3
 * precede the index you want to search with the table name in Camel Case. There is no change with table1 fields, use them the regular way without table name prefix.  
   - e.g. someTableName.indexYouWantToSearch = value
 * The table2 fields to be used must have an index declared in schema.json, it must be of type "index", not "likeIndex", "uniqueIndex", "ginIndex", or "fullTextIndex".
-* currently no other operators are supported.  They will be added at a later date. Attention: Using other operators (e.g. <>) will silently use the = operator.
+* Use table.index = * in order to do a filter query with no condition
 * Example Schema: 
 {
```
craigmcnally: This example is hard to read as it's rendered as a single long line.  Maybe put in a code block?

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
```diff
@@ -54,4 +65,99 @@ public static Index getIndex(String cqlIndex, List<Index> indexes) {
...
+   */
+  public static List<DbFkInfo> findForeignKeysFromSourceAliasToTargetTable(Schema dbSchema, String srcTabAlias,
+      String targetTabName) {
+    List<DbFkInfo> list = new ArrayList<>();
+    for (Table table : dbSchema.getTables()) {
+      if (table.getForeignKeys() == null) {
+        continue;
+      }
+      for (ForeignKeys fk : table.getForeignKeys()) {
+        String tabAlias = fk.getTableAlias() == null ? table.getTableName() : fk.getTableAlias();
```
julianladisch: This is contrary to what we decided in RMB-395.
We decided that if tableAlias is null then there will be no sub query at all and that we don't want tableName as a fall back because that will lead to unwanted results:
a) The table name holdings_records is not suitable, we want camel case.
b) There may be a name clash in that there exists a field that is that same as the table name.

README.md
adamdickmeiss: Right. Thanks

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: I don't see that being decided at all. Primarily an optional feature to avoid name clashes.

adamdickmeiss:
There are some unit tests in ForeignKeyGenerationTest.java for subquery.json / instance

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: I'm pretty sure the table that is listed first in schema.json is used (array list). So tableX is chosen before tableY if that is listed first in schema.json.

cql2pgjson/src/main/java/org/folio/cql2pgjson/util/DbSchemaUtils.java
adamdickmeiss: Same here.

PR #489, opened 2019-08-08
RMB-432: RMB-438: fulltext: word splitting, punctuation, minus

julianladisch:


cql2pgjson/src/test/java/org/folio/cql2pgjson/util/Cql2SqlUtilTest.java
```diff
@@ -153,6 +171,214 @@ public void cql2regexp(String cql, String sql) {
...
+        "abc* def* ghi*",  // 'abc':* & 'def':* & 'ghi':*
+                           // minus in words, RMB-438
+        "abc-def",         // 'abc-def' & 'abc' & 'def'
+        "abc--def",        // 'abc' & 'def'
+        "abc---def",       // 'abc' & 'def'
+        "abc-def-ghi",     // 'abc-def-ghi' & 'abc' & 'def' & 'ghi'
+        "abc-def-ghi*",    // 'abc-def-ghi':* & 'abc':* & 'def':* & 'ghi':*
+        "abc-def ghi-jkl", // 'abc-def' & 'abc' & 'def' & 'ghi-jkl' & 'ghi' & 'jkl'
+        "0cc175b9-c0f1-b6a8-31c3-99e269772661",
+        "12345678-9999-0000-1234-012345678901",
```
adamdickmeiss: Now that UUIDs are better matched (was & before).. I think it would be nice with a test using
a commutation of those ones.. Eg expect f(alse) for 0cc175b9-c0f1-31c3-b6a8-99e269772661

PR #492, opened 2019-08-13
RMB-450: Updated Documentation for Cross Table Query

evaluk:
Updated docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1067,18 +1067,24 @@ Correct number matching must result in 3.4 == 3.400 == 0.34e1 and correct number
 If the search term is a number then a numeric mode is used for "==", "<>", "<", "<=", ">", and ">=" if the actual JSONB type of the stored value is `number`
 (JSONB has no `integer` type).
 
-### CQL2PgJSON: Cross index searches
+### CQL2PgJSON: Cross table index queries
 
-Limited cross table searches are supported.  If you desire a join across tables the following conditions must be met:
+CQL2PgJSON supports cross table joins via subquery.  This allows arbitrary depth relationships in parent->child and child->parent relationships.  Foreign keys must be uni-directional(parent->child and child->parent).
 
-* The join desired index must be only 1 table deep
-  - e.g.  table1 -> table2 not table1 -> table2 -> table3
+Example relationship:
+```
+table1 -> table2 -> table3
+```
+or
+```
+table1 <- table2 <- table3
+```
+* Start at table 3 and query table 1 or table 1 to query table 3 or anywhere in between.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1082,18 +1082,24 @@ Correct number matching must result in 3.4 == 3.400 == 0.34e1 and correct number
 If the search term is a number then a numeric mode is used for "==", "<>", "<", "<=", ">", and ">=" if the actual JSONB type of the stored value is `number`
 (JSONB has no `integer` type).
 
-### CQL2PgJSON: Cross index searches
+### CQL2PgJSON: Cross table index queries
 
-Limited cross table searches are supported.  If you desire a join across tables the following conditions must be met:
+CQL2PgJSON supports cross table joins via subquery.  This allows arbitrary depth relationships in parent->child and child->parent relationships.  Foreign keys must be uni-directional(parent->child and child->parent).
```
julianladisch: Foreign keys must be uni-directional, the direction is always from the foreign key field in the child table to the primary key id field of the parent table.

README.md
```diff
@@ -1082,18 +1082,24 @@ Correct number matching must result in 3.4 == 3.400 == 0.34e1 and correct number
...
 
-* The join desired index must be only 1 table deep
-  - e.g.  table1 -> table2 not table1 -> table2 -> table3
+Example relationship:
+```
+table1 -> table2 -> table3
+```
+or
+```
+table1 <- table2 <- table3
```
julianladisch:
```
table1 -> table2 -> table3
```
or
```
table3 <- table2 <- table1
```

README.md
```diff
@@ -1116,10 +1122,60 @@ Limited cross table searches are supported.  If you desire a join across tables
           "targetTable": "table1"
         }
       ]
+    },
+    {
+      "tableName": "table2",
```
julianladisch: This is the second "table2" section, there cannot be a duplicate.

README.md
```diff
@@ -1116,10 +1122,60 @@ Limited cross table searches are supported.  If you desire a join across tables
...
+          "fieldName": "table2Id",
+          "targetTable": "table2"
+        }
+      ]
     }
   ]
 }
 ```
+## Cross Table query ambiguity
+There will be some cases where two foreign keys refer to the same table.  This ambiguity has the possibility of leading the join logic down a dead end if the foreign key does not lead to the intended table.  To resolve this there are new properties:
```
julianladisch: The second and third sentence should be reduced to:

    This ambiguity is resolved by these properties:

README.md
evaluk: this was meant to say table3, updated

README.md
evaluk: I am not sure what should be changed here so I removed the line.

PR #494, opened 2019-08-14
Fix array search example typo (reported by Jakub)

adamdickmeiss:


skoczko:
I am still unclear why "subfield" is an object (and not a string) in this example. I would expect this:

```
"property" : [ 
  {
    "type1" : "value1",
    "type2" : "value2",
    "subfield": "value"
  },
 ...
]
```

Since the query you give as an example is `property =/@type1=value1 value`

adamdickmeiss:
You got it

PR #495, opened 2019-08-15
Explain and analyze queries RMB-457 (part 1)

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -96,6 +96,8 @@
   public static final String     DEFAULT_SCHEMA           = "public";
   public static final String     DEFAULT_JSONB_FIELD_NAME = "jsonb";
 
+  static final long              EXPLAIN_QUERY_THRESHOLD_DEFAULT = 1000;
```
julianladisch: Can you add some javadoc explaining the unit (seconds?) and whether there is a magic value (-1?) to disable it.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -157,15 +159,19 @@
 
   private static final Logger log = LoggerFactory.getLogger(PostgresClient.class);
 
+
   private static int embeddedPort            = -1;
 
+  private static long explainQueryThreshold = EXPLAIN_QUERY_THRESHOLD_DEFAULT;
```
julianladisch: Can you add some javadoc about the unit (seconds?) and whether there is a magic value (-1?) to disable it.

README.md
```diff
@@ -279,9 +279,15 @@ RMB implementing modules expect a set of environment variables to be passed in a
...
  - DB_CHARSET
  - DB_MAXPOOLSIZE
+ - RMB_EXPLAIN_QUERY_THRESHOLD
 
 Environment variables with periods/dots in their names are deprecated in RMB because a period is [not POSIX compliant](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) and therefore some shells, notably, the BusyBox /bin/sh included in Alpine Linux, strip them (reference: [warning in OpenJDK docs](https://hub.docker.com/_/openjdk/)).
 
+The environment variable `RMB_EXPLAIN_QUERY_THRESHOLD` is not observed by Postgres itself, but
+is a value - in milliseconds - that triggers query execution analysis. If a single query exceeds this
+threshold, it will be analyzed by using `EXPLAIN ANALYZE`. Note that this in turn further adds time
+to the query, so this should only be executed for slow queries that needs further attention.
```
julianladisch: Can you add whether this can be set per RMB module, per PostgresClient, per connection or per query?

adamdickmeiss: Per RMB module like the rest

julianladisch: To avoid any confustion: Can you add that this is not [auto_explain](https://www.postgresql.org/docs/current/auto-explain.html) but a reimplementation?

README.md
```diff
@@ -279,9 +279,15 @@ RMB implementing modules expect a set of environment variables to be passed in a
...
  - DB_CHARSET
  - DB_MAXPOOLSIZE
+ - RMB_EXPLAIN_QUERY_THRESHOLD
 
 Environment variables with periods/dots in their names are deprecated in RMB because a period is [not POSIX compliant](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) and therefore some shells, notably, the BusyBox /bin/sh included in Alpine Linux, strip them (reference: [warning in OpenJDK docs](https://hub.docker.com/_/openjdk/)).
 
+The environment variable `RMB_EXPLAIN_QUERY_THRESHOLD` is not observed by Postgres itself, but
+is a value - in milliseconds - that triggers query execution analysis. If a single query exceeds this
+threshold, it will be analyzed by using `EXPLAIN ANALYZE`. Note that this in turn further adds time
+to the query, so this should only be executed for slow queries that needs further attention. Like the DB-environment variables this pertains per RMB-module (process).
```
julianladisch: This tells me: "this should only be executed for slow queries" How can I enable this per query if this is per RMB-module?
Or should it say: "so this should only be executed in a separate performance test environment."?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2537,6 +2538,41 @@ public void select(String sql, Handler<AsyncResult<ResultSet>> replyHandler) {
...
+    conn.query(sql, res -> {
+      long queryTime = (System.nanoTime() - start);
+      StatsTracker.addStatElement(STATS_KEY + statMethod, queryTime);
+      if (res.failed()) {
+        log.error("queryAndAnalyze: " + res.cause().getMessage() + " - "
+          + sql, res.cause());
+        replyHandler.handle(Future.failedFuture(res.cause()));
+        return;
+      }
+      if (queryTime >= explainQueryThreshold * 1000000) {
```
julianladisch: Can you reduce indentation and complexity and improve readability by using a guard clause?

    if (queryTime < explainQueryThreshold * 1000000) {
      replyHandler.handle(res);
      return;
    }
    final String explainQuery = "EXPLAIN ANALYZE " + sql;
    …

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2537,6 +2538,41 @@ public void select(String sql, Handler<AsyncResult<ResultSet>> replyHandler) {
...
+        final String explainQuery = "EXPLAIN ANALYZE " + sql;
+        conn.query(explainQuery, explain -> {
+          replyHandler.handle(res); // not before, so we have conn if it gets closed
+          if (explain.failed()) {
+            log.warn(explainQuery + " failed", explain.cause().getMessage());
```
julianladisch: This swallows the stack trace. Can you pass the Exception to the logger?

    log.warn(explain.cause().getMessage() + ": " + explainQuery, explain.cause());

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2537,6 +2538,41 @@ public void select(String sql, Handler<AsyncResult<ResultSet>> replyHandler) {
...
+            return;
+          }
+          StringBuilder e = new StringBuilder();
+          for (JsonArray ar : explain.result().getResults()) {
+            if (ar.getValue(0) instanceof String) {
+              e.append("\n" + ar.getString(0));
```
julianladisch: e.append('\n').append(ar.getString(0));

avoids creating a temporary String.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2537,6 +2538,41 @@ public void select(String sql, Handler<AsyncResult<ResultSet>> replyHandler) {
...
+            }
+          }
+          log.warn(explainQuery + e.toString());
```
julianladisch: explainQuery should be appended to e a few lines before. This avoids creating a temporary String needed for this + operation.

julianladisch:
This currently only affects PostgresClient.get(…) and PostgresClient.join(…) methods, but no other methods like PostgresClient.getById(…) or PostgresClient.streamGet(…).
This should be documented.

README.md
adamdickmeiss: No. We want it to be enabled by default with NO modification to any environment so we get information about unacceptable slow queries.

adamdickmeiss:
> There should be a documented way how to disable it.
> It should be disabled by default.

Jakub and I agreed it should be enabled by default for slow queries.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: I'm going to say no to that one.. Why? SQ did not complain and guard clauses IMHO are mostly for exceptional cases/errors.. which this is NOT.

julianladisch: SQ doesn't have any guard clause rules, this is checked by human code review only.
Guard clauses are *not* mostly for exceptional cases/errors:
* https://refactoring.guru/replace-nested-conditional-with-guard-clauses
* http://wiki.c2.com/?GuardClause
They are also for checking the state of the object or checking for trivial cases.
Can you elaborate whether that guard clause reduces complexity and improves readability?

julianladisch: Using the proposed guard clause reduces the Cognitive Complexity (as calculated by SQ) by 4.

README.md
julianladisch: Thank you, can you add to README that it is enabled by default and the default threshold value?

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/utils/Envs.java
```diff
@@ -12,7 +12,8 @@
   DB_DATABASE,
   DB_QUERYTIMEOUT,
   DB_CHARSET,
-  DB_MAXPOOLSIZE;
+  DB_MAXPOOLSIZE,
+  RMB_EXPLAIN_QUERY_THRESHOLD;
```
julianladisch: To make it work in Envs it must start with `DB_`

adamdickmeiss: It was having prefix NOT DB_ to avoid it passing passed to the Mysql/postgres client.. (which would probably choke if we passed an unknown setting ). I don't have tests / or it-tests with the env-thingy set (since I don't know how to set System env in tests), but manual testing suggests it works fine.

julianladisch:
This doesn't read the option from the configuration file: https://github.com/folio-org/raml-module-builder/blob/v27.0.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L482

My suggestion:
Rename the configuration option to start with DB_.

At the end of getPostgreSQLClientConfig https://github.com/folio-org/raml-module-builder/blob/v27.0.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L511
check if the explain option is set. If yes remove it from the config JsonObject and set explainQueryThreshold accordingsly.
Extend Envs to properly parse it as Long.

adamdickmeiss:
> This doesn't read the option from the configuration file: https://github.com/folio-org/raml-module-builder/blob/v27.0.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L482
> 

Exact and intended.. This option should not be passed to the DB driver.


> My suggestion:
> Rename the configuration option to start with DB_.
> 
> At the end of getPostgreSQLClientConfig https://github.com/folio-org/raml-module-builder/blob/v27.0.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L511
> check if the explain option is set. If yes remove it from the config JsonObject and set explainQueryThreshold accordingsly.
> Extend Envs to properly parse it as Long.

skoczko:
Guys, can we merge this? @julianladisch did you have any other reservations?

PR #497, opened 2019-08-15
added tables, data entry and tests for integration for multi table index query

evaluk:
I added some test set up changes and clarifications.  Added test cases for the desired functionality


cql2pgjson/src/test/resources/templates/db_scripts/foreignKey.json
```diff
@@ -57,10 +57,16 @@
...
         }
+      ],
+      "index": [
+
+        {
+          "fieldName": "cindex",
+          "tOps": "ADD",
+          "caseSensitive": true,
+          "removeAccents": true
+        }
```
julianladisch: Removing this new "index" for "cindex" does not make the tests fail.
Either remove this new "index" section or add a test that uses it.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -64,6 +65,20 @@ public void foreignKeySearchParentChild() throws Exception {
...
+  @Test
+  public void foreignKeySearchMulti2() throws Exception {
+    assertThat(cql("tablec.cindex == z2"), containsInAnyOrder("test2"));
+  }
+
+  @Test
+  public void foreignKeySearchMultiWild() throws Exception {
+    assertThat(cql("tablec.cindex == *"), containsInAnyOrder("test1", "test2"));
+  }
+
```
julianladisch: These are tests from parent (table a) to child (table c).
Can you also add at least one test for the other direction?

    cqlc("tablea.name == test1")

julianladisch:
Can you add a blank line between all methods to match the existing code style?

PR #501, opened 2019-08-16
RMB-200: Single quote SQL Injection in PostgresClient.update with updateSection

julianladisch:


evaluk:
Everything seems fine with this change.

PR #517, opened 2019-09-03
Okapi 457 db explain query threshold

adamdickmeiss:


domain-models-interface-extensions/src/test/java/org/folio/rest/tools/utils/EnvsTest.java
```diff
@@ -26,6 +26,7 @@ public void setUp() {
     map.put("DB_MAXPOOLSIZE", "5");
     // we dropped support for dot form. check that it is ignored
     map.put("db.username", "superwoman");
+    map.put("DB_EXPLAIN_QUERY_THRESHOLD", "100");
```
julianladisch: This is misplaced and should go before the line comment.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -475,6 +492,11 @@ static JsonObject getPostgreSQLClientConfig(String tenantId, String schemaName,
       throws Exception {
     // static function for easy unit testing
     JsonObject config = environmentVariables;
+
+    Object v = config.remove(Envs.DB_EXPLAIN_QUERY_THRESHOLD.name());
+    if (v instanceof Long) {
+      PostgresClient.setExplainQueryThreshold((Long) v);
+    }
```
julianladisch: This should go after the `if (config == null)` block. This ensures that config is not null, and the the config can be read from the config file at `getConfigFilePath()`.

PR #519, opened 2019-09-12
RMB-421: 422 JSON error message for PgUtil

julianladisch:


domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -193,6 +225,147 @@ static String message(Throwable t) {
...
+        String value = matcher.group(2);
+        String refTable = matcher.group(3);
+        String message = "Cannot set " + table + "." + field + " = " + value
+            + " because it does not exist in " + refTable + ".id.";
+        if (response422method == null) {
+          return response(message, valueMethod, failResponseMethod);
+        }
+        return respond422(response422method, table + "." + field, value, message);
+      }
+
```
adamdickmeiss: A lot of work for little gain ..

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -193,6 +225,147 @@ static String message(Throwable t) {
...
+      String message = key + " value already exists in table " + table + ": " + value;
+      if (response422method == null) {
+        return response(message, valueMethod, failResponseMethod);
+      }
+      return respond422(response422method, key, value, message);
+    } catch (Exception e) {
+      return response(e, failResponseMethod);
+    }
+  }
+
```
adamdickmeiss: And two more identical looking functions here.. Why not just having a simple translation table of the Postgres IDs  and a map to the status?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -57,6 +62,49 @@
...
+   * <p>
+   * 1 = <code>id</code>
+   * <p>
+   * 2 = <code>64f55fa2-50f4-40e5-978a-bbad17dc644d</code>
+   * <p>
+   * 3 = <code>referencing</code>
+   */
+  private static final Pattern KEY_STILL_REFERENCED_PATTERN = Pattern.compile(
+      "^Key \\(([^=]+)\\)=\\((.*)\\) is still referenced from table \"(.*)\".$");
   /** Number of records to read from the sort index in getWithOptimizedSql and generateOptimizedSql method */
```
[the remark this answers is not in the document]
julianladisch: The message-Id 23503 is the same for KEY_NOT_PRESENT and KEY_STILL_REFERENCED.
The new message contains the "table" variable that Postgres doesn't mention in the message; this helps to see which of several SQL statements caused the error.
There were complaints that the error message text from Postgres is not human readable and should be improved: https://issues.folio.org/browse/RMB-403 https://issues.folio.org/browse/MODTAG-18
Wordings like "Key (id)=(64f55fa2-50f4-40e5-978a-bbad17dc644d) is still referenced from table ..." do not describe in plain English what happened, they are too technical and therefore violate the accessibility requirements: https://www.w3.org/TR/WCAG21/#error-identification

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
julianladisch: One Postgres ID can have several status messages, for example 23503 is the same for KEY_NOT_PRESENT and KEY_STILL_REFERENCED.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
julianladisch: The gain is small if only a single module will use this. However this is a library method to be used by many modules and many, many API endpoints.

adamdickmeiss: OK.. If you say so..

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: Of course... And we'll have to maintain that mapping from Postgres to human readable messages or some of them for the future.. which is just one extra thing we'll have to update if Postgres changes behavior on this etc.. I really hope the users will enjoy this.. If we have a condition that a UI can react to .. A multilingual UI will have to map yet again.. Which is why an error code (NOT HTTP error code), is better for this.

adamdickmeiss: I think you should just merge this anyway.

PR #530, opened 2019-10-03
Vert.x must be updated too for RMB 27.1

adamdickmeiss:


adamdickmeiss:
I noticed this when working with upgrading mod-inventory-storage to RMB 27.1.1

PR #533, opened 2019-10-14
RMB-474: Support 'field[*].subfield' syntax in multiFieldNames

julianladisch:


cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -65,6 +68,41 @@ public void multiFieldnamesMultipartFT() throws Exception {
     String expected = "WHERE to_tsvector('simple', concat_space_sql(tabled.jsonb->'proxy'->'personal'->>'city' , tabled.jsonb->'proxy'->'personal'->>'state')) @@ replace((to_tsquery('simple', ('''Boston''')) && to_tsquery('simple', ('''MA''')))::text, '&', '<->')::tsquery";
     assertEquals(expected, sql);
   }
+
+  @Test
+  public void multiFieldnamesFTStar() throws Exception {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
+    String sql = cql2pgJson.toSql("ftfieldstar = \"Boston MA\"").toString();
+    String expected = "WHERE to_tsvector('simple', concat_space_sql(concat_array_object_values(tablea.jsonb->>'field1','city') , concat_array_object_values(tablea.jsonb->>'field2','state'))) @@ replace((to_tsquery('simple', ('''Boston''')) && to_tsquery('simple', ('''MA''')))::text, '&', '<->')::tsquery";
```
julianladisch: Can you change

* `tablea.jsonb->>'field1'` to `tablea.jsonb->'field1'` and
* `tablea.jsonb->>'field2'` to `tablea.jsonb->'field2'`

to avoid converting the jsonb object to a string?

cql2pgjson/src/test/resources/compoundIndex.sql
```diff
@@ -18,7 +18,20 @@ $func$  LANGUAGE sql IMMUTABLE;
 create or replace function concat_space_sql(VARIADIC text[])
 RETURNS text AS $$ select concat_ws(' ', VARIADIC $1); $$ LANGUAGE SQL IMMUTABLE;
 
+create or replace function concat_array_object_values(jsonb_data text, field text) RETURNS text AS $$
```
julianladisch: Can you change `jsonb_data text` to `jsonb_data jsonb` to avoid converting the jsonb object array to a string?

cql2pgjson/src/test/resources/compoundIndex.sql
```diff
@@ -18,7 +18,20 @@ $func$  LANGUAGE sql IMMUTABLE;
...
+DECLARE
 
+  res text;
+  i jsonb;
+BEGIN
+  
+  FOR i IN SELECT * FROM json_array_elements(($1)::json)
```
julianladisch: Can you use `jsonb_array_elements($1)`? This avoids converting the JSONB to a string and converting the string into a JSON.

domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerTest.java
```diff
@@ -82,6 +82,11 @@ public void canCreateCompoundIndex() throws IOException, TemplateException {
 
     assertThat(result,containsString("CREATE INDEX IF NOT EXISTS tablea_ftfield_idx_ft"));
     assertThat(result, containsString("concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName')"));
+
+    assertThat(result,containsString("((concat_space_sql(concat_array_object_values(tablea.jsonb->>'user','firstName') , concat_array_object_values(tablea.jsonb->>'user','lastName'))) gin_trgm_ops)"));
+    assertThat(result,containsString("((concat_space_sql(concat_array_object_values(tablea.jsonb->'user'->>'info','firstName') , concat_array_object_values(tablea.jsonb->'user'->>'info','lastName'))) gin_trgm_ops)"));
+    assertThat(result,containsString("( to_tsvector('english', concat_space_sql(concat_array_object_values(tablea.jsonb->>'field1','firstName') , concat_array_object_values(tablea.jsonb->>'field2','lastName'))) )"));
+    assertThat(result,containsString("( to_tsvector('english', concat_space_sql(concat_array_object_values(tablea.jsonb->'field1'->>'info','firstName') , concat_array_object_values(tablea.jsonb->'field2'->>'info','lastName'))) )"));
```
julianladisch: Can you replace 'english' by 'simple'? See https://issues.folio.org/browse/RMB-462

cql2pgjson/src/test/resources/compoundIndex.sql
```diff
@@ -18,7 +18,20 @@ $func$  LANGUAGE sql IMMUTABLE;
...
+   LOOP
+    res =  concat_ws(', ',res, i->$2);
```
julianladisch: Can you use `i->>$2` to get the plain string without quotes and not a JSONB string including quotes?

julianladisch: Can you use ' ' instead of ', '? We don't need commas in the full text matching string.

cql2pgjson/src/test/resources/compoundIndex.sql
```diff
@@ -18,7 +18,20 @@ $func$  LANGUAGE sql IMMUTABLE;
...
+  END LOOP;
+  RETURN res;
+END; 
+$$ LANGUAGE plpgsql IMMUTABLE;
```
julianladisch: Can you use this function that runs 9% faster?

    CREATE OR REPLACE FUNCTION concat_array_object_values(jsonb, text) RETURNS text LANGUAGE sql IMMUTABLE AS
      $$ SELECT string_agg(value->>$2, ' ') FROM jsonb_array_elements($1) $$;

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -65,6 +68,41 @@ public void multiFieldnamesMultipartFT() throws Exception {
...
+    assertEquals(expected, sql);
+  }
+  @Test
+  public void multiFieldnamesGINDotStar() throws Exception {
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
+    cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
+    String sql = cql2pgJson.toSql("ginfielddotstar == \"Boston MA\"").toString();
+    String expected = "WHERE lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field3'->>'info','city') , concat_array_object_values(tablea.jsonb->'field3'->>'info','state'))) LIKE lower('Boston MA')";
+    assertEquals(expected, sql);
+  }
```
julianladisch: Can you add support for string arrays as described in the `languages[*]` example in RMB-474?

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,24 +84,45 @@ public String getFinalSqlExpression(String tableLoc) {
           result .append(" , ");
         }
         String [] rawExpandedTerm = splitIndex[i].split("\\.");
-        StringBuilder expandedTerm = formatExpandedTerm(rawExpandedTerm);
-        result.append(tableLoc).append(".jsonb").append(expandedTerm);
+        StringBuilder expandedTerm = formatExpandedTerm(tableLoc + ".jsonb",rawExpandedTerm);
+        result.append(expandedTerm);
```
julianladisch: Can you change these three lines to appendExpandedTerm(result, tableLoc, splitIndex[i])? This will move all expanding code into the method and improve readability.

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,24 +84,45 @@ public String getFinalSqlExpression(String tableLoc) {
...
       }
       result.append(")");
       return result.toString();
     }
 
   }
-  private StringBuilder formatExpandedTerm(String[] rawExpandedTerm) {
-    StringBuilder expandedTerm = new StringBuilder();
 
+  private StringBuilder formatExpandedTerm(String table, String[] rawExpandedTerm) {
```
julianladisch: It is very difficult to understand this method. The two cases (with and witout [*]) are joined but should completely been split for readability. For example:

    void appendExpandedTerm(StringBuilder result, String tableLoc, String term) {
      if (term.contains("[*]") {
        appendExpandedArrayTerm(result, tableLoc, term);
      } else {
        appendExpandedSimpleTerm(result, tableLoc, term);
      }
    }

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,24 +84,45 @@ public String getFinalSqlExpression(String tableLoc) {
...
+    StringBuilder expandedTerm = new StringBuilder();
+    StringBuilder result = new StringBuilder();
```
julianladisch: Creating additional StringBuilders for substrings should be avoided. The idea of a StringBuilder is to try to append to a single result StringBuilder.

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
julianladisch: Can you add unit tests for this method? The most easy way is using https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests , for example:

    @ParameterizedTest
    @CsvSource({
      "a    , ->>a",
      "a.b  , ->'a'->>'b'",
      "a.b.c, ->'a'->'b'->>'c'",
    })
    void appendExpandedSimpleTerm(String term, String expectedSql) {
      StringBuilder result = new StringBuilder();
      Index.appendExpandedTerm(result, "table", term);
      assertThat(result.toString(), is("table.jsonb" + expecedSql));
    }

The same for array terms.

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
evaluk: I understand what you are saying regarding this stringbuilder, however the information needed to do this wrapping comes naturally while iterating through the list of terms.  But the actual function call itself needs to wrap the term.  So I feel like my options are to use a sub builder like I did, or add a bunch of logic to detect the start and stop of a term with the brackets in it.

PR #534, opened 2019-10-15
RMB-482: Use command line to enable metrics

hjiebsco:


domain-models-runtime/src/test/java/org/folio/rest/RestLauncherTest.java
```diff
@@ -0,0 +1,57 @@
...
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
+
+public class RestLauncherTest {
+
+  private static final String DISABLE_METRICS = "-Dvertx.metrics.options.enabled=false";
+  private static final String ENABLE_METRICS = "-Dvertx.metrics.options.enabled=true";
+
+  @Test
+  public void testMetricsOptions01() {
```
julianladisch: Can you avoid the number and use a descriptive method name for readability, for example `metricsIsDisabledByDefault`?

domain-models-runtime/src/test/java/org/folio/rest/RestLauncherTest.java
```diff
@@ -0,0 +1,57 @@
...
+    DummyLauncher launcher = new DummyLauncher();
+    launcher.dispatch(new String[] { "run", TestVerticle.class.getName() });
+    assertFalse(launcher.enabled);
+  }
+
+  @Test
+  public void testMetricsOptions02() {
```
julianladisch: Can you avoid the number and use a descriptive method name for readability, for example `canEnableMetrics`?

domain-models-runtime/src/test/java/org/folio/rest/RestLauncherTest.java
```diff
@@ -0,0 +1,57 @@
...
+    DummyLauncher launcher = new DummyLauncher();
+    launcher.dispatch(new String[] { "run", TestVerticle.class.getName(), ENABLE_METRICS });
+    assertTrue(launcher.enabled);
+  }
+
+  @Test
+  public void testMetricsOptions03() {
```
julianladisch: Can you avoid the number and use a descriptive method name for readability, for example `canDisableMetrics`?

domain-models-runtime/src/test/java/org/folio/rest/RestLauncherTest.java
```diff
@@ -0,0 +1,57 @@
+package org.folio.rest;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
```
julianladisch: Can you document the reason why JUnit 4 is used and not JUnit 5?

PR #535, opened 2019-10-21
RMB-433: saveBatch must not overwrite existing id

julianladisch:


julianladisch:
The saveBatch with column parameter was introduced by [RMB-305](https://issues.folio.org/browse/RMB-305) = [MODORDERS-97](https://issues.folio.org/browse/MODORDERS-97):
https://github.com/folio-org/raml-module-builder/pull/354/files
But this was unrelated to that DISTINCT ON change.
The saveBatch with column parameter is not in use: https://github.com/search?q=org%3Afolio-org+savebatch&type=Code
There is not use case for it because we only have `jsonb` as column name in the database table.

PR #541, opened 2019-10-24
RMB-496: connectionReleaseDelay: Use 1 minute as default, add config

julianladisch:
I forgot to enable connectionReleaseDelay in RMB-485.
* Use 1 minute delay as default (the same default we had before: https://github.com/folio-org/vertx-mysql-postgresql-client/blob/release-connection-pool-3-5-1-folio-1/src/main/java/io/vertx/ext/asyncsql/impl/pool/AsyncConnectionPool.java#L52 ).
* Allow DB_CONNECTIONRELEASEDELAY environment variable to overwrite the default.
* Document this in RMB README.


julianladisch:
Yes

PR #544, opened 2019-10-25
Rmb 497 replace jsql parser with sqlselect

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -150,25 +150,56 @@ private void spaceAppend(StringBuilder sb, String text) {
...
       }
-      append(sb, wrap.wrapper.getQuery(), wrap.wrapper.getField());
+      appendWhere(sb, wrap.wrapper.getQuery(), wrap.wrapper.getField());
     }
-    if (sb.length() > 0) {
-      sb.insert(0, " WHERE ");
+    return sb.toString();
+  }
+
+  public String getWhereFull() {
```
julianladisch: Can you add javadoc?

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -150,25 +150,56 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+    String s = getWhere();
+    if (!s.isEmpty()) {
+      return " WHERE " + s;
+    }
+    return "";
```
julianladisch: More easy to read and understand and avoids the negation:

    if (s.isEmpty()) {
      return "";
    }
    return " WHERE " + s;

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -150,25 +150,56 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+  }
+
+  public String getOrderBy() {
```
julianladisch: Can you add javadoc?

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -150,25 +150,56 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+      return "";
+    }
+    try {
+      return field.toSql(query).getOrderBy();
+    } catch (QueryValidationException e) {
+      throw new CQLQueryValidationException(e);
     }
+  }
+
+  public String getOrderByFull() {
```
julianladisch: Can you add javadoc?

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -150,25 +150,56 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+    String s = getOrderBy();
+    if (!s.isEmpty()) {
+      return "ORDER BY " + s;
+    }
+    return "";
```
julianladisch: More easy to read and understand and avoids the negation:

    if (s.isEmpty()) {
      return "";
    }
    return "ORDER BY " + s;

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -150,25 +150,56 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+      return field.toSql(query).getOrderBy();
+    } catch (QueryValidationException e) {
+      throw new CQLQueryValidationException(e);
     }
+  }
+
+  public String getOrderByFull() {
+    String s = getOrderBy();
+    if (!s.isEmpty()) {
+      return "ORDER BY " + s;
```
julianladisch: Why doesn't this return a space as first character like getWhereFull() does?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2271,6 +2271,184 @@ public void getDistinctOnWithFacets(TestContext context) throws IOException {
...
+        "cql.allRecords="); // syntax error
+      Async async = context.async();
+      postgresClient.get(MOCK_POLINES_TABLE, Object.class, "*",
+        cqlWrapper, true, true, false, null, null/*facets*/, handler -> {
+          context.assertTrue(handler.failed());
+          async.complete();
+        });
+      async.awaitSuccess();
+    }
+  }
```
julianladisch: This can be reduced to

    @Test
    public void getCQLWrapperFailure(TestContext context) throws Exception {
      CQL2PgJSON cql2pgJson = new CQL2PgJSON("jsonb");
      CQLWrapper cqlWrapper = new CQLWrapper(cql2pgJson,
          "cql.allRecords="); // syntax error
      postgresClient().get(MOCK_POLINES_TABLE, Object.class, "*",
          cqlWrapper, true, true, false, null, null/*facets*/, context.asyncAssertFailure());
    }

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1503,7 +1532,7 @@ public TotaledResults(ResultSet set, Integer total) {
    * @param facets
    * @param replyHandler
    */
-  private <T> void doGet(
+  private <T> void doGetObsolete(
```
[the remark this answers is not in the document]
adamdickmeiss: I'll revert the name. It's an internal function, so deprecating is not of much use.. I do want to remove it eventually internally.

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: Sure

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: I'll change and not return a leading space for that

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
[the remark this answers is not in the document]
julianladisch: The `@Deprecated` annotation is better because SonarLint will report it and IDEs will show the name as deprecated, for example in strike-out fashion.

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -151,24 +151,73 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+  public String getWhereFull() {
+    String s = getWhere();
+    if (s.isEmpty()) {
+      return "";
+    }
+    return "WHERE " + s;
+  }
+
+  /**
+   * @return sort by query excluding SORT BY prefix or empty string if no sorting
```
julianladisch: Should be "SQL sort expression excluding ORDER BY ..."

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -151,24 +151,73 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+    }
+    try {
+      return field.toSql(query).getOrderBy();
+    } catch (QueryValidationException e) {
+      throw new CQLQueryValidationException(e);
+    }
+  }
+
+  /**
+   * @return sort by query including SORT BY prefix or empty string if no sorting
```
julianladisch: Should be "SQL sort expression including ORDER BY ..."

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1754,7 +1754,7 @@ public void executeParamConnectionException(TestContext context) throws Exceptio
 
   @Test
   public void executeTrans(TestContext context) {
-    Async async1 = context.async();
+    Async async = context.async();
```
julianladisch: You cannot reuse an Async. After async.complete() has been invoked once all other async.awaitSuccess will not wait but succeed immediately.
Please revert.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2199,8 +2197,7 @@ public void streamGetDistinctOnWithFacets(TestContext context) throws IOExceptio
       add(new FacetField("jsonb->>'edition'"));
     }};
 
-    postgresClient = createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
-
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
```
julianladisch: @After tearDown will close postgresClient. Why do you want to keep it open?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2224,51 +2221,206 @@ public void getDistinctOn(TestContext context) throws IOException {
...
 
-    String whereClause =  "WHERE jsonb->>'order_format' = 'Other'";
-    Async async2 = context.async();
+    String whereClause = "WHERE jsonb->>'order_format' = 'Other'";
     postgresClient.get(MOCK_POLINES_TABLE, Object.class, "*", whereClause, false, false,
       false, null, distinctOn, handler -> {
         ResultInfo resultInfo = handler.result().getResultInfo();
         context.assertEquals(1, resultInfo.getTotalRecords());
-        async2.complete();
+        async.complete();
```
julianladisch: You cannot reuse an Async. The second awaitSuccess() will not wait until the second assertEquals has been executed.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2224,51 +2221,206 @@ public void getDistinctOn(TestContext context) throws IOException {
...
     //with facets and where clause RMB-355
-    Async async2 = context.async();
     postgresClient.get(MOCK_POLINES_TABLE, Object.class, "*", whereClause, true, false,
       false, facets, distinctOn, handler -> {
         ResultInfo resultInfo = handler.result().getResultInfo();
         context.assertEquals(1, resultInfo.getTotalRecords());
         List<Facet> retFacets = resultInfo.getFacets();
         context.assertEquals(1, retFacets.size());
-        async2.complete();
+        async.complete();
```
julianladisch: You cannot reuse an Async.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2224,51 +2221,206 @@ public void getDistinctOn(TestContext context) throws IOException {
...
+
+    CQL2PgJSON cql2pgJson = new CQL2PgJSON(MOCK_POLINES_TABLE + ".jsonb");
+
+    CQLWrapper cqlWrapper = new CQLWrapper(cql2pgJson, "cql.allRecords=1");
+    postgresClient.get(MOCK_POLINES_TABLE, Object.class, "*",
+      cqlWrapper, false, true, false, null, null/*facets*/, handler -> {
+        context.assertTrue(handler.succeeded());
+        ResultInfo resultInfo = handler.result().getResultInfo();
+        context.assertEquals(6, resultInfo.getTotalRecords());
+        context.asyncAssertSuccess();
```
julianladisch: This handler will never be executed, the unit test will succeed without running the asserts.

See https://vertx.io/docs/vertx-unit/java/#_asynchronous_assertions how to do it:

    postgresClient.get(MOCK_POLINES_TABLE, Object.class, "*",
      cqlWrapper, false, true, false, null, null/*facets*/, context.asyncAssertSuccess(result -> {
        ResultInfo resultInfo = handler.result().getResultInfo();
        context.assertEquals(6, resultInfo.getTotalRecords());
      }});

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2224,51 +2221,206 @@ public void getDistinctOn(TestContext context) throws IOException {
...
+    async.awaitSuccess();
+
+    String distinctOn = "jsonb->>'order_format'";
+    cqlWrapper = new CQLWrapper(cql2pgJson, "cql.allRecords=1");
+    postgresClient.get(MOCK_POLINES_TABLE, Object.class, "jsonb",
+      cqlWrapper, true, true, false, distinctOn, facets, handler -> {
+        context.assertTrue(handler.succeeded());
+        ResultInfo resultInfo = handler.result().getResultInfo();
+        context.assertEquals(4, resultInfo.getTotalRecords());
+        async.complete();
```
julianladisch: You cannot reuse a succeeded Async.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2224,51 +2221,206 @@ public void getDistinctOn(TestContext context) throws IOException {
...
+    String distinctOn = "jsonb->>'order_format'";
+    cqlWrapper = new CQLWrapper(cql2pgJson, "cql.allRecords=1");
+    postgresClient.get(MOCK_POLINES_TABLE, Object.class, "*",
+      cqlWrapper, true, true, false, distinctOn, facets, handler -> {
+        context.assertTrue(handler.succeeded());
+        ResultInfo resultInfo = handler.result().getResultInfo();
+        context.assertEquals(4, resultInfo.getTotalRecords());
+        List<Facet> retFacets = resultInfo.getFacets();
+        context.assertEquals(0, retFacets.size());
+        async.complete();
```
julianladisch: You cannot reuse a succeded Async.

julianladisch: There is no need for Async, you can run these three get requests in parallel and make use of context.asyncAssertSuccess.

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -123,14 +132,16 @@ public CQLWrapper addWrapper(CQLWrapper wrapper, String operator){
...
     }
-    try {
-      sb.append(field.cql2pgJson(query));
-    } catch (QueryValidationException e) {
-      throw new CQLQueryValidationException(e);
+    if (field != null && query != null) {
+      try {
+        sb.append(field.toSql(query).getWhere());
+      } catch (QueryValidationException e) {
+        throw new CQLQueryValidationException(e);
```
julianladisch: Can you add a test for this catch clause to CQLWrapperTest?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1489,6 +1492,35 @@ public TotaledResults(ResultSet set, Integer total) {
     }
   }
 
+  /* low-level SQL getter, based on CQLWrapper.. Does not handle facets yet */
+  /* this function should eventually replace doGetObsolete */
```
julianladisch: rename doGetObsolete -> doGet

adamdickmeiss:
Thx for the feedback

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criterion.java
```diff
@@ -231,7 +243,12 @@ private static Criterion processQueryIntern(Iterator<JsonNode> iter, Criterion c
     return cc;
   }
 
-  public String selects2String() {
+  /**
+   * @return SELECT clause - without SELECT - possibly empty string
+   * @deprecated not used by RMB itself anymore and should not be publicly used
+   */
+  @Deprecated
+  public String selects2String1() {
```
julianladisch: Either keep the name and make it deprecated, or completely remove it.
I suggest to remove it because this is part of a breaking change and not used: https://github.com/search?q=org%3Afolio-org+selects2String&type=Code

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criterion.java
```diff
@@ -252,7 +269,12 @@ public String selects2String() {
     return sb.toString();
   }
 
-  public String from2String() {
+  /**
+   * @return as clause for SELECT of empty string if none
+   * @deprecated not used by RMB itself anymore and should not be publicly used
+   */
+  @Deprecated
+  public String from2String1() {
```
julianladisch: Either keep the name (don't rename to from2Strin1) and make it deprecated, or completely remove it.
I suggest to remove it because this is part of a breaking change and not used: https://github.com/search?q=org%3Afolio-org+from2String&type=Code

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -76,9 +86,30 @@ public CQLWrapper setField(CQL2PgJSON field) {
...
+    if (whereClause != null) {
+      return whereClause;
+    }
+    if (criterion != null) {
+      return criterion.toString();
+    }
+    return query; // CQL query
+  }
+
+  public String getType() {
```
julianladisch: Can you make this method private, or add javadoc explaining the possible return values?

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -23,6 +26,13 @@ public CQLWrapper() {
     super();
   }
 
+  public CQLWrapper(Criterion criterion) {
+    super();
```
julianladisch: Can you remove this redundant `super()` in this an all other CQLWrapper constructors?

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -23,6 +26,13 @@ public CQLWrapper() {
     super();
   }
```
julianladisch: Can you remove this redundant constructor?

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -151,24 +151,73 @@ private void spaceAppend(StringBuilder sb, String text) {
...
+   * @return query including SQL clauses of WHERE, ORDER BY
+   */
+  public String getWithoutLimOff() {
+    StringBuilder sb = new StringBuilder(getWhereFull());
+    spaceAppend(sb, getOrderByFull());
+    return sb.toString();
+  }
+
+  /**
+   * @return full query including SQL clauses of WHERE, ORDER BY, OFFSET and LIMIT.
```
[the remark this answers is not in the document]
julianladisch: If this is an intended breaking change please mention it in doc/upgrading.md.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criterion.java
adamdickmeiss: Sure.  Removal is the way to go . And sorry about that rename that I missed.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criterion.java
adamdickmeiss: Also here.

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: Sure. I had it public for a moment when logging queries in RestVerticle.

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: Sure

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
adamdickmeiss: Sure

julianladisch: `super()` has been removed, but the constructor is still there.

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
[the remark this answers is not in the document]
adamdickmeiss: Right. I'll document it. Would be very surprised it this is used by a module, except for debugging/printing purposes.

PR #553, opened 2019-11-12
Use log4j's MDC rather than slf4j's MDC RMB-511

adamdickmeiss:
This is because we're using log4j as logging implementation for RMB
and we've seen calls to MDC fail in some cases when using the slf4j verion.


adamdickmeiss:
I'm sure we can remove the direct use of slf4j.. Problem is if 3rd party is logging straight to slf4j. Vert.x itself we know is using jul so that's fine.. But netty or other? I'll check it out.

PR #554, opened 2019-11-12
RMB-510: Get rid of hack that allows empty body for tenant init post.

adamdickmeiss:


README.md
```diff
@@ -1565,7 +1565,10 @@ The converter will generate LIKE queries for `==` queries. For example `?query=f
 
 ##### Posting information
 
-Posting a new tenant can optionally include a body. The body should contain a JSON conforming to the https://github.com/folio-org/raml/blob/master/schemas/moduleInfo.schema schema. The `module_to` entry is mandatory if a body is included in the request, indicating the version module for this tenant. The `module_from` entry is optional and indicates an upgrade for the tenant to a new module version.
+Posting a new tenant must include a body. The body should contain a JSON conforming to the https://github.com/folio-org/raml/blob/master/schemas/moduleInfo.schema schema. The `module_to` entry is mandatory if a body is included in the request, indicating the version module for this tenant. The `module_from` entry is optional and indicates an upgrade for the tenant to a new module version.
```
julianladisch: "if a body is included" should be removed.

doc/upgrading.md
```diff
@@ -13,6 +13,9 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
 
 ## Version 29
 
+* RMB-510 Tenant POST `/_/tenant` must include a body. Older versions of RMB
+  made a special case for zero length body in which case the JSON content
+  was not decoded.
```
julianladisch: The second sentence should be removed, the upgrading information is about *how* to upgrade a module. Information about *why* the upgrade is needed is off-topic in this upgrade recipe and can be looked up via the issue number; you may convert the ticket number into a clickable link `[RMB-510](https://issues.folio.org/browse/RMB-510)`, though.
Adding an example for a minimal JSON body may speed up the upgrading.

PR #557, opened 2019-11-13
Get rid of ParsedQuery class and RMB-506

adamdickmeiss:


adamdickmeiss:
No need to review, just yet. Needs more work!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1739,57 +1735,41 @@ QueryHelper buildQueryHelper(
...
     }
-    if (offsetClause != null) {
-      Matcher matcher = OFFSET_MATCH_PATTERN.matcher(offsetClause);
-      if (matcher.find()) {
-        queryHelper.offset = Integer.parseInt(matcher.group(1));
-      }
+    if (distinctOnClause.equals("*")) { // only estimate on non distinct cases
+      queryHelper.countQuery = SELECT + "count_estimate_smart('"
+        + org.apache.commons.lang.StringEscapeUtils.escapeSql(queryHelper.countQuery)
+        + "')";
```
julianladisch: These three lines have no test coverage. Can you write an integration test that runs this with same sample data in the database or can you at least write a unit test for this method?

adamdickmeiss: Yes.. My bad.. Of course these lines should be in use..

adamdickmeiss:
I need to document the exactCount propery in schema.json as well..

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1654,7 +1650,7 @@ public TotaledResults(ResultSet set, Integer total) {
 
     vertx.runOnContext(v1 -> {
       try {
-        QueryHelper queryHelper = buildSelectQueryHelper(transactionMode, table, fieldName, where, returnIdField, facets, distinctOn);
+        QueryHelper queryHelper = buildSelectQueryHelper(transactionMode, table, fieldName, where, returnIdField, distinctOn);
```
julianladisch: Please also remove the `facets` parameter of `doStreamGet`.

adamdickmeiss: Yes.. that's actually RMB-359 resurfacing!!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1621,10 +1617,12 @@ public TotaledResults(ResultSet set, Integer total) {
     boolean returnIdField, boolean setId, List<FacetField> facets, String distinctOn,
     Handler<T> streamHandler, Handler<AsyncResult<Void>> replyHandler
   ) {
+    // streamGet appears to offer facets, but it does not implement it
```
julianladisch: Can you move this comment into a javadoc for the facets parameter?
Can you reply with a failed future with an exception (for example UnsupportedOperationException) if facets is neither null nor empty?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientITBase.java
```diff
@@ -41,6 +42,24 @@ protected static void setUpClass(TestContext context) throws Exception {
...
         "CREATE SCHEMA " + schema + " AUTHORIZATION " + schema,
         "GRANT ALL PRIVILEGES ON SCHEMA " + schema + " TO " + schema);
+    loadFuncs(context, PostgresClient.getInstance(vertx));
+  }
+
+  private static void loadFuncs(TestContext context, PostgresClient postgresClient) {
+    Async async = context.async();
+    try {
+      String sql = IOUtils.toString(
+        PostgresClientITBase.class.getClassLoader().getResourceAsStream("templates/db_scripts/funcs.sql"), "UTF-8");
```
julianladisch: Can this copy & paste been avoided?

funcs.sql is a copy of general_functions.ftl any can become out of sync. I suggest to delete funcs.sql and to read general_functions.ftl instead. Use replaceAll to replace the variables.

adamdickmeiss: Yeah.. It's annoying.. Violation of DRY. I'll see if I can fix that.

PR #561, opened 2019-11-15
Rmb 487 support fieldarray notation

evaluk:


cql2pgjson/src/test/resources/templates/db_scripts/compoundIndex.json
```diff
@@ -23,6 +23,13 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
+          "fieldName": "ginfielddotstarplain",
+          "multiFieldNames": "field3.info[*],field3.info[*]",
```
julianladisch: Can you use an example with two different arrays?

cql2pgjson/src/test/resources/templates/db_scripts/compoundIndex.json
```diff
@@ -46,6 +53,13 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
+          "fieldName": "ftfielddotstarplain",
+          "multiFieldNames": "field3.info[*],field3.info[*]",
```
julianladisch: Can you use an example with two different arrays?

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -129,7 +134,12 @@ private static  boolean appendExpandedSimpleTerm(String[] rawExpandedTerm, Strin
   private static boolean appendExpandedArrayTerm(String[] rawExpandedTerm, StringBuilder expandedTerm, int currentTermIdx,
       int tokenIdx) {
     String arrowToken = "->";
-    expandedTerm.append(arrowToken).append("'").append(rawExpandedTerm[currentTermIdx].substring(0,tokenIdx)).append("',").append("'").append(rawExpandedTerm[currentTermIdx+1]).append("'");
+    if(currentTermIdx == rawExpandedTerm.length - 1) {
+      arrowToken = "->";
+      expandedTerm.append(arrowToken).append("'").append(rawExpandedTerm[currentTermIdx].substring(0,tokenIdx)).append("'");
+    } else {
+      expandedTerm.append(arrowToken).append("'").append(rawExpandedTerm[currentTermIdx].substring(0,tokenIdx)).append("',").append("'").append(rawExpandedTerm[currentTermIdx+1]).append("'");
+    }
```
julianladisch: appendExpandedTerm, formatExpandedTerm, appendExpandedSimpleTerm, appendExpandedArrayTerm are complicated code and difficult to understand. They reimplement the code in index2sqlText and index2sqlJson from Cql2PgJson. My suggestion is to wait until "RMB-518: Extract index2sqlText and index2sqlJson for reuse" #563 has been merged. That will make the implementation in Index.java very easy.

adamdickmeiss: Please do not make term manipulation in dbschema.. That belongs to cql2pgjson ... Is it difficult to make that switch?

adamdickmeiss: Most guidelines have a blank after a reserved word.. Eg "if" in this case.

evaluk: I have re worked the code to us the utility functions

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,53 +85,40 @@ public String getFinalSqlExpression(String tableLoc) {
         if(i != 0) {
           result .append(" , ");
         }
-        appendExpandedTerm(tableLoc, splitIndex[i], result);
+        result.append(formatExpandedTerm(tableLoc, splitIndex[i]));
```
julianladisch: Can you revert this back to `appendExpandedTerm(tableLoc, splitIndex[i], result);`?

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,53 +85,40 @@ public String getFinalSqlExpression(String tableLoc) {
...
 
   }
-  public static void appendExpandedTerm(String tableLoc, String  currentTerm, StringBuilder result) {
-    String [] rawExpandedTerm = currentTerm.split("\\.");
-    StringBuilder expandedTerm = formatExpandedTerm(tableLoc + ".jsonb",rawExpandedTerm);
-    result.append(expandedTerm);
-  }
 
-  public static StringBuilder formatExpandedTerm(String table, String[] rawExpandedTerm) {
+  public static StringBuilder formatExpandedTerm(String table, String term) {
```
julianladisch: Can you change this to `static void appendExpandedTerm(String table, String term, StringBuilder result)`?
Is there a need to have this method `public`?

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,53 +85,40 @@ public String getFinalSqlExpression(String tableLoc) {
...
-        wasArrayIndex = appendExpandedSimpleTerm(rawExpandedTerm, expandedTerm, j, idx);
+        result.append(table).append(".").append(Cql2PgUtil.cqlNameAsSqlText("jsonb",term));
       }
-    }
-    if(wasArrayIndex) {
-      result.append("concat_array_object_values(").append(expandedTerm).append(")");
-    } else {
-      result.append(expandedTerm);
-    }
     return result;
```
julianladisch: Can you make the three cases more visible? Example:

```
// term does not contain "[*]"
if (idx == -1) {
  result.append(...
  return result;
}

// term ends with "[*]"
if (idx == term.length() - 3) {
  result.append(...
  return result;
}

// term ends with "[*].field"
result.append(...
return result;
```

The code in method formatArrayWithTerm is easy enough to move it into the calling method.

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -83,53 +85,40 @@ public String getFinalSqlExpression(String tableLoc) {
...
-  }
-  private static boolean appendExpandedArrayTerm(String[] rawExpandedTerm, StringBuilder expandedTerm, int currentTermIdx,
-      int tokenIdx) {
-    String arrowToken = "->";
-    expandedTerm.append(arrowToken).append("'").append(rawExpandedTerm[currentTermIdx].substring(0,tokenIdx)).append("',").append("'").append(rawExpandedTerm[currentTermIdx+1]).append("'");
-    return true;
+  private static void formatArrayWithTerm(String term, StringBuilder expandedTerm, StringBuilder result,
+      int arrayTermPresent) {
+    result.append("concat_array_object_values(").append(expandedTerm);
+    String arrayTerm = term.substring( arrayTermPresent + "[*].".length(), term.length());
```
julianladisch: Can you directly append the substring without creating arrayTerm?
`result.append(term, arrayTermPresent + "[*].".length(), term.length());`
or
`result.append(term, idx + 4, term.length());`

PR #570, opened 2019-11-25
Fix TenantLoading on Windows RMB-524

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -199,23 +197,29 @@ private static void endWithXHeaders(HttpClientRequest req, Map<String, String> h
...
+    if (base == -1) {
+      f.handle(Future.failedFuture("No basename for " + path));
+      return null;
+    }
+    if (suf > base) {
+      id = path.substring(base, suf);
+    } else {
+      id = path.substring(base);
+    }
+    return id;
```
julianladisch: There is no need for `id` variable, you can directly return the value.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -242,24 +241,35 @@ private static String getId(LoadingEntry loadingEntry, URL url, String content,
...
+  static String getContent(URL url, Future<Void> f) {
     try {
       InputStream stream = url.openStream();
-      content = IOUtils.toString(stream, StandardCharsets.UTF_8);
+      String content = IOUtils.toString(stream, StandardCharsets.UTF_8);
       stream.close();
-      if (loadingEntry.contentFilter != null) {
-        content = loadingEntry.contentFilter.apply(content);
-      }
+      return content;
```
julianladisch: `return IOUtils.toString(url, StandardCharsets.UTF_8);` (or at least use try-with-resources)

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
```diff
@@ -560,6 +564,33 @@ public void test404IdRaw(TestContext context) {
...
+    promise = Promise.promise();
+    context.assertEquals(null, TenantLoading.getIdBase("a", promise.future()));
+    context.assertTrue(promise.future().isComplete());
+    promise = Promise.promise();
+    context.assertEquals("/a", TenantLoading.getIdBase("/path/a", promise.future()));
+    context.assertFalse(promise.future().isComplete());
+    promise = Promise.promise();
+    context.assertEquals("/c", TenantLoading.getIdBase("/a.b/c", promise.future()));
+    context.assertFalse(promise.future().isComplete());
+  }
```
julianladisch: `testGetIdBase` contains unnecessary boilerplate code that makes it difficult to understand what is tested. How about something like this?

    private void assertGetIdBase(String path, String expectedBase) {
      Future<Void> future = Promise.<Void>promise().future();
      assertEquals(expectedBase, TenantLoading.getIdBase(path, future));
      assertFalse(future.isComplete());
    }

    private void assertGetIdBaseFail(String path) {
      Future<Void> future = Promise.<Void>promise().future();
      assertEquals(null, TenantLoading.getIdBase(path, future));
      assertTrue(future.failed());
      assertEquals("No basename for " + path, future.cause().getMessage());
    }

    @Test
    public void testGetIdBase() {
      assertGetIdBase("/path/a.json", "/a");
      assertGetIdBase("/path/a", "/a");
      assertGetIdBase("/a.b/c", "/c");
      assertGetIdBaseFail("a.json");
      assertGetIdBaseFail("a");
    }

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
```diff
@@ -560,6 +564,33 @@ public void test404IdRaw(TestContext context) {
...
+
+  @Test
+  public void testContent(TestContext context) throws MalformedURLException {
+    URL url = new URL("file:///does.not.exist");
+    Promise promise = Promise.promise();
+    context.assertEquals(null, TenantLoading.getContent(url, promise.future()));
```
julianladisch: Can you add `assertTrue(promise.future().failed())` ?

PR #571, opened 2019-11-25
Rmb 498 truncate b tree string for 2712 index row size

evaluk:


evaluk:
I need to correct the syntax unit tests in other test areas but I wanted to get opinions on the logic I used.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -939,4 +953,23 @@ private String queryBySql(boolean hasIndex, IndexTextAndJsonValues vals, CQLTerm
     return sql;
   }
 
+  private String appendFullTermIfTooLong(String comparator, String index, String termUnmod, String sql) {
+    if(termUnmod.length() >= 600) {
```
julianladisch: Can we be sure that for all unicode strings Java's length calculation and PostgreSQL's length calculation is always the same?
If not this if should be converted into SQL.

cql2pgjson/src/test/resources/jo-ka-lea.sql
```diff
@@ -12,5 +12,5 @@ INSERT INTO users (id,user_data,groupId) VALUES
     "name": "Ka Keller", "status": "Inactive", "email": "ka@example.com",
     "address": {"city": "Fred", "zip": 1900}, "lang": ["en", "dk", "fi"]}', '88888888-8888-8888-8888-888888888888'),
   ('33333333-3333-3333-3333-33333333333a','{"id":"33333333-3333-3333-3333-333333333333",
-    "name": "Lea Long", "status": "Active - Not Yet", "email": "lea@example.com",
+    "name": "Lea Long", "status": "Active - Not Yet", "email": "lea@example.com", "alternateEmail": "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff@example.com",
```
julianladisch: We need two records where the first 600 characters are the same but that differ afterwards.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -212,7 +212,9 @@ public void cql2pgJsonException(CQL2PgJSON cql2pgJson, String cql,
...
+    "alternateEmail==\"ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff@example.com\"                       # Lea Long",})
```
julianladisch: More tests are needed.
* a value where the first difference is after 600 characters.
* `<>`
* `<` `>` `<=` `>=`

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
evaluk: ok so i spent some time looking at this and I agree with the need to sync up the length measurements.  Because the string length of Java could not match what the database is using.  Couple questions I have about this.  

My first is what the return value would be for a single function to handle any query.  RETURNS SETOF XXXXX.  What should XXXXX be?  It would have to be parameterized to the table its called on?  I dont see a way to do that.  

Second I would assume I would have to send in the query and execute it the same way I am doing in Java.  This would essentially be replacing parts of the java we have with sql.

julianladisch: I don't see the need for a function. This is an example query. 'foo' will be replaced by the search string.
```
SELECT * FROM t WHERE 
    left(jsonb->>'title', 600) = left('foo', 600) 
    AND CASE WHEN length('foo') > 600
             THEN jsonb->>'title' = 'foo' 
             ELSE TRUE
        END;
```

julianladisch: If you want to create a function you need to create a separate function for each field where an b-tree index is created. Example:
```
CREATE OR REPLACE FUNCTION titlelike(instance, text) RETURNS boolean AS $$
  SELECT left($1.jsonb->>'title', 600) LIKE left($2, 600)
    AND CASE WHEN length($2) > 600
             THEN  $1.jsonb->>'title' LIKE $2
             ELSE TRUE END;
$$ LANGUAGE SQL;
```

evaluk: I believe I have resolved this.  Will work through failures now

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -578,6 +592,9 @@ public void like(String testcase) throws QueryValidationException {
     "*         sortBy name/sort.ascending/string     # Jo Jane; Ka Keller; Lea Long",
     "*         sortBy name/sort.descending           # Lea Long; Ka Keller; Jo Jane",
     "*         sortBy name/sort.descending/string    # Lea Long; Ka Keller; Jo Jane",
+    "*         sortBy alternateEmail                 # Jo Jane; Lea Long; Ka Keller",
+    "*         sortBy alternateEmail/sort.ascending  # Jo Jane; Lea Long; Ka Keller",
+    "*         sortBy alternateEmail/sort.descending # Ka Keller; Jo Jane; Lea Long",
```
julianladisch: The correct sort order for alternateEmail/sort.descending is `Ka Keller; Lea Long; Jo Jane`.
Note that sort.descending must be the reverse of sort.ascending because all values are distinct.
```
"fff" + 600 x "f" + "gfffffffffff@example.com"          # Jo Jane
"fff" + 600 x "f" + "hfffffffffff@example.com"          # Lea Long
"ffg" + 600 x "f" + "ffffffffffff@example.com"          # Ka Keller
```

Please fix the code: `ORDER BY left(term,600) DESC, term DESC`

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -11,7 +11,7 @@ public void multiFieldNamesGIN() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("fullname == \"John Smith\"").toString();
-    String expected = "WHERE lower(concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName')) LIKE lower('John Smith')";
+    String expected = "WHERE CASE WHEN length('John Smith') <= 600 THEN left(lower(concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName')),600) LIKE left(lower('John Smith'),600)  ELSE left(tablea->>'fullname',600)  LIKE left('John Smith',600)  AND tablea->>'fullname'  LIKE  'John Smith' END";
```
julianladisch: Currently compoundIndex.json doesn't have any b-tree indexes. The length case clause should only be used if there exist a b-tree index.

julianladisch: A test case for a b-tree index on a compound field should be added, though.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -649,6 +666,19 @@ public void numberInStringField(String testcase) throws CQL2PgJSONException {
...
+    "alternateEmail>=\"fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffhfffffffffff@example.com\" # Lea Long; Ka Keller",
```
julianladisch: Please fix the code to make these correct test cases succeed.

domain-models-runtime/src/main/resources/templates/db_scripts/indexes.ftl
```diff
@@ -3,7 +3,7 @@
     <#list table.index as indexes>
     <#if indexes.tOps.name() == "ADD">
     CREATE INDEX IF NOT EXISTS ${table.tableName}_${indexes.fieldName}_idx ON ${myuniversity}_${mymodule}.${table.tableName}
-    (${indexes.getFinalSqlExpression(table.tableName)})
+    (left(${indexes.getFinalSqlExpression(table.tableName)},600))
```
julianladisch: This is for the non-unique b-tree index only.

Please also add this for the unique b-tree.

evaluk: I do not understand what needs to be done here.  Can you explain what the difference is?

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
evaluk: see comment below, but I do not understand what needs to be done here to change this.

domain-models-runtime/src/main/resources/templates/db_scripts/indexes.ftl
julianladisch: left(..., 600) is missing in line 19.

evaluk: it appears that there is already a unit test checking for this index inside of SchemaMakerTest.ginIndex.  It failed the test after I modified it

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -883,19 +883,29 @@ private String queryByLike(String index, boolean hasIndex, IndexTextAndJsonValue
       sql = arrayNode(index, node, modifiers, relationModifiers, schemaIndex, vals, targetTable);
     } else {
       Index schemaIndex = null;
+      Index otherIndex = null;
       if (targetTable != null) {
         schemaIndex = DbSchemaUtils.getIndex(index, targetTable.getGinIndex());
+        otherIndex = DbSchemaUtils.getIndex(index, targetTable.getIndex());
```
julianladisch: Can you check for a unique index if it isn't a (non-unique b-tree) index?

domain-models-runtime/src/main/resources/templates/db_scripts/indexes.ftl
```diff
@@ -16,7 +16,7 @@
     <#list table.uniqueIndex as indexes>
     <#if indexes.tOps.name() == "ADD">
     CREATE UNIQUE INDEX IF NOT EXISTS ${table.tableName}_${indexes.fieldName}_idx_unique ON ${myuniversity}_${mymodule}.${table.tableName}
-    ( ${indexes.getFinalSqlExpression(table.tableName)} )
+    (left( ${indexes.getFinalSqlExpression(table.tableName)},600) )
```
julianladisch: Can you make lines 6 and 19 having the same whitespace?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -395,7 +395,7 @@ private SqlSelect toSql(CQLSortNode node) throws QueryValidationException {
       }
 
       // We assume that a CREATE INDEX for this has been installed.
-      order.append(wrapInLowerUnaccent(vals.getIndexText(), modifiers)).append(desc);
+      order.append(wrapForLength(wrapInLowerUnaccent(vals.getIndexText(), modifiers))).append(desc).append(", ").append(wrapInLowerUnaccent(vals.getIndexText(), modifiers)).append(desc);
```
julianladisch: Can you wrap this line?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -883,19 +883,29 @@ private String queryByLike(String index, boolean hasIndex, IndexTextAndJsonValue
...
+        indexMod = schemaIndex.getSqlExpression();
       } else {
-        sql = wrapInLowerUnaccent(indexText, schemaIndex);
+        indexMod = wrapInLowerUnaccent(indexText, schemaIndex);
+      }
+      if(otherIndex != null) {
+        indexMod = wrapForLength(indexMod);
+        sql = "CASE WHEN length(" + term + ") <= 600 THEN "
+         + indexMod +  likeOperator + wrapForLength(wrapInLowerUnaccent(term, schemaIndex));
+        sql = appendLengthCase(likeOperator,likeOperator, indexText, term, comparator.equals("<>"), sql);
```
julianladisch: The methods queryByLike and queryBySql contain almost the same code for the CASE...END clause. Can you extract the code into a method, for example
`private String caseLength(String operator, String index, String term)`
and let this method return the complete CASE WHEN ... THEN ... ELSE ... END clause?
This will also improve readability because the complete CASE ... END clause is at one place and is no longer split over several methods.

evaluk: I extracted these 2 functions but they are different enough that I cannot merge them yet.  I need to look at this closer.  I can do this if we want to wait on this PR.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -919,25 +925,66 @@ private String queryByLike(String index, DbIndex dbIndex, IndexTextAndJsonValues
...
+    String lengthCaseComparator = "";
+    if(comparator.equals("<") ){
+      lengthCaseComparator = "<=";
+    } else if(comparator.equals(">")) {
+      lengthCaseComparator = ">=";
+    } else {
+      lengthCaseComparator = comparator;
+    }
+    return lengthCaseComparator;
+  }
```
julianladisch: This is more easy to read and understand:

    private String lengthCaseComparator(String comparator) {
      switch (comparator) {
      case "<":
        return "<=";
      case ">":
        return ">=";
      default:
        return comparator;
      }
    }

"determine" in the method name is redundant.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -919,25 +925,66 @@ private String queryByLike(String index, DbIndex dbIndex, IndexTextAndJsonValues
...
+      String s = String.format("%s, CQL >>> SQL: %s >>> %s", indexMod, node.toCQL(), sql);
       logger.log(Level.WARNING, "Doing SQL query without index for {0}", s);
     }
 
-    logger.log(Level.FINE, "index {0} generated SQL {1}", new Object[] {index, sql});
+    logger.log(Level.FINE, "index {0} generated SQL {1}", new Object[] {indexMod, sql});
+    return sql;
+  }
+
+  private String createSQLLengthCase(String comparator, String index, String term, boolean not, Index schemaIndex) {
```
julianladisch: The `not` parameter is always false and therefore should be removed.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -919,25 +925,66 @@ private String queryByLike(String index, DbIndex dbIndex, IndexTextAndJsonValues
...
+    String sql;
+    sql = "CASE WHEN length(" + wrapInLowerUnaccent(term, schemaIndex) + ") <= 600 THEN " + wrapInLowerUnaccent(index, schemaIndex) + " " + comparator + " " + wrapInLowerUnaccent(term, schemaIndex);
+    String lengthCaseComparator = determineLengthCaseComparator(comparator);
+    sql = appendLengthCase(lengthCaseComparator,comparator, wrapInLowerUnaccent(index, schemaIndex), wrapInLowerUnaccent(term, schemaIndex), not, sql);
+    return sql;
```
julianladisch: This method is difficult to read and understand.
"CASE WHEN" and "THEN" is here, but "ELSE ... END" is hidden in the method appendLengthCase.
Can you move all the code from appendLengthCase into this method?
Then it is possible to directly return the result without a temporary sql variable:

    return "CASE WHEN length(" + wrapInLowerUnaccent(term, schemaIndex) + ...

(This is also faster because the compiler automatically uses a StringBuilder for string concatenation).

Can you limit the line length to 120 characters and use line breaks and indentation for readability? Example: https://github.com/folio-org/raml-module-builder/blob/v29.1.2/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L1021

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -919,25 +925,66 @@ private String queryByLike(String index, DbIndex dbIndex, IndexTextAndJsonValues
...
+  }
+
+  private String createLikeLengthCase(String comparator, String indexText, Index schemaIndex, String likeOperator, String term) {
+    String sql;
+    sql = "CASE WHEN length(" + wrapInLowerUnaccent(term, schemaIndex) + ") <= 600 THEN " + wrapForLength(indexText) +  likeOperator + wrapInLowerUnaccent(term, schemaIndex);
+    sql = appendLengthCase(likeOperator,likeOperator, indexText, wrapInLowerUnaccent(term, schemaIndex), comparator.equals("<>"), sql);
     return sql;
   }
```
julianladisch: Same comments, so createSQLLengthCase above.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -881,15 +882,20 @@ private String queryByLike(String index, DbIndex dbIndex, IndexTextAndJsonValues
       sql = arrayNode(index, node, modifiers, relationModifiers, schemaIndex, vals, targetTable);
     } else {
       String likeOperator = comparator.equals("<>") ? " NOT LIKE " : " LIKE ";
```
julianladisch: Either remove the heading and trainling whitespace here, or remove it below when adding lengthCaseComparator.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -6,12 +6,39 @@
...
   public void multiFieldNamesGIN() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("fullname == \"John Smith\"").toString();
-    String expected = "WHERE lower(concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName')) LIKE lower('John Smith')";
+    String expected = "WHERE CASE WHEN length(lower('John Smith')) <= 600 THEN left(lower(concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName')),600) "
+        + "LIKE lower('John Smith') ELSE left(lower(concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName')),600)  LIKE  left(lower('John Smith'),600) "
+        + "AND lower(concat_space_sql(tablea.jsonb->>'firstName' , tablea.jsonb->>'lastName'))  "
+        + "LIKE  lower('John Smith') END";
     assertEquals(expected, sql);
```
julianladisch: Can you wrap the lines at "THEN", "ELSE" and "AND" for readability, see multiFieldNamesUniqueIndex above?

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -29,7 +56,10 @@ public void SQLExpressionGIN() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tableb");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("address == \"Boston MA\"").toString();
-    String expected = "WHERE lower(concat_space_sql(jsonb->>'city', jsonb->>'state')) LIKE lower('Boston MA')";
+    String expected = "WHERE CASE WHEN length(lower('Boston MA')) <= 600 THEN left(lower(concat_space_sql(jsonb->>'city', jsonb->>'state')),600) "
+        + "LIKE lower('Boston MA') ELSE left(lower(concat_space_sql(jsonb->>'city', jsonb->>'state')),600)  "
+        + "LIKE  left(lower('Boston MA'),600) AND lower(concat_space_sql(jsonb->>'city', jsonb->>'state'))  "
+        + "LIKE  lower('Boston MA') END";
     assertEquals(expected, sql);
```
julianladisch: This unit test is about an sqlExpession. The CASE WHEN ELSE END with left(...) has already been tested in multiFieldNames* method above and should therefore not pollute this unit tests.
Reducing the test to the relevant part makes it more easy to understand and maintain. Example:
    
    String expected = "lower(concat_space_sql(jsonb->>'city', jsonb->>'state')) LIKE lower('Boston MA')";
    assertThat(sql, containsString(expected));

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -56,7 +86,9 @@ public void multiFieldnamesSpacesGIN() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablec");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("tablecginindex == \"Boston MA\"").toString();
-    String expected = "WHERE lower(concat_space_sql(tablec.jsonb->>'firstName' , tablec.jsonb->>'lastName')) LIKE lower('Boston MA')";
+    String expected = "WHERE CASE WHEN length(lower('Boston MA')) <= 600 THEN left(lower(concat_space_sql(tablec.jsonb->>'firstName' , tablec.jsonb->>'lastName')),600) "
+        + "LIKE lower('Boston MA') ELSE left(lower(concat_space_sql(tablec.jsonb->>'firstName' , tablec.jsonb->>'lastName')),600)  "
+        + "LIKE  left(lower('Boston MA'),600) AND lower(concat_space_sql(tablec.jsonb->>'firstName' , tablec.jsonb->>'lastName'))  LIKE  lower('Boston MA') END";
```
julianladisch: See above.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -101,7 +133,10 @@ public void multiFieldnamesGINStar() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("ginfieldstar == \"Boston MA\"").toString();
-    String expected = "WHERE lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field1','city') , concat_array_object_values(tablea.jsonb->'field2','state'))) LIKE lower('Boston MA')";
+    String expected = "WHERE CASE WHEN length(lower('Boston MA')) <= 600 THEN left(lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field1','city') , concat_array_object_values(tablea.jsonb->'field2','state'))),600) "
+        + "LIKE lower('Boston MA') ELSE left(lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field1','city') , concat_array_object_values(tablea.jsonb->'field2','state'))),600)  "
+        + "LIKE  left(lower('Boston MA'),600) AND lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field1','city') , concat_array_object_values(tablea.jsonb->'field2','state')))  "
+        + "LIKE  lower('Boston MA') END";
```
julianladisch: See above.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -110,7 +145,10 @@ public void multiFieldnamesGINDotStar() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("ginfielddotstar == \"Boston MA\"").toString();
-    String expected = "WHERE lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field3'->'info','city') , concat_array_object_values(tablea.jsonb->'field3'->'info','state'))) LIKE lower('Boston MA')";
+    String expected = "WHERE CASE WHEN length(lower('Boston MA')) <= 600 THEN left(lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field3'->'info','city') , concat_array_object_values(tablea.jsonb->'field3'->'info','state'))),600) "
+        + "LIKE lower('Boston MA') ELSE left(lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field3'->'info','city') , concat_array_object_values(tablea.jsonb->'field3'->'info','state'))),600)  "
+        + "LIKE  left(lower('Boston MA'),600) AND lower(concat_space_sql(concat_array_object_values(tablea.jsonb->'field3'->'info','city') , concat_array_object_values(tablea.jsonb->'field3'->'info','state')))  "
+        + "LIKE  lower('Boston MA') END";
```
julianladisch: See above.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -119,15 +157,21 @@ public void multiFieldnamesGINDotStarPlain() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tablea");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("ginfielddotstarplain == \"Boston MA\"").toString();
-    String expected = "WHERE lower(concat_space_sql(concat_array_object(tablea.jsonb->'field3'->'info') , concat_array_object(tablea.jsonb->'field3'->'data'))) LIKE lower('Boston MA')";
+    String expected = "WHERE CASE WHEN length(lower('Boston MA')) <= 600 THEN left(lower(concat_space_sql(concat_array_object(tablea.jsonb->'field3'->'info') , concat_array_object(tablea.jsonb->'field3'->'data'))),600) "
+        + "LIKE lower('Boston MA') ELSE left(lower(concat_space_sql(concat_array_object(tablea.jsonb->'field3'->'info') , concat_array_object(tablea.jsonb->'field3'->'data'))),600)  "
+        + "LIKE  left(lower('Boston MA'),600) AND lower(concat_space_sql(concat_array_object(tablea.jsonb->'field3'->'info') , concat_array_object(tablea.jsonb->'field3'->'data')))  "
+        + "LIKE  lower('Boston MA') END";
```
julianladisch: See above.

cql2pgjson/src/test/java/org/folio/cql2pgjson/CompoundIndexTest.java
```diff
@@ -119,15 +157,21 @@ public void multiFieldnamesGINDotStarPlain() throws Exception {
...
   @Test
   public void multiFieldnamesMultipartGIN() throws Exception {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON("tabled");
     cql2pgJson.setDbSchemaPath("templates/db_scripts/compoundIndex.json");
     String sql = cql2pgJson.toSql("tabledginindex == \"Boston MA\"").toString();
-    String expected = "WHERE lower(concat_space_sql(tabled.jsonb->'proxy'->'personal'->>'city' , tabled.jsonb->'proxy'->'personal'->>'state')) LIKE lower('Boston MA')";
+    String expected = "WHERE CASE WHEN length(lower('Boston MA')) <= 600 THEN left(lower(concat_space_sql(tabled.jsonb->'proxy'->'personal'->>'city' , tabled.jsonb->'proxy'->'personal'->>'state')),600) "
+        + "LIKE lower('Boston MA') ELSE left(lower(concat_space_sql(tabled.jsonb->'proxy'->'personal'->>'city' , tabled.jsonb->'proxy'->'personal'->>'state')),600)  "
+        + "LIKE  left(lower('Boston MA'),600) AND lower(concat_space_sql(tabled.jsonb->'proxy'->'personal'->>'city' , tabled.jsonb->'proxy'->'personal'->>'state'))  "
+        + "LIKE  lower('Boston MA') END";
```
julianladisch: See above.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
evaluk: there is one case where the operator is determined by the comparator that is sent in.  On line 963 it checks comparator.equals("<>")

PR #572, opened 2019-11-26
TenantLoading in sequence, accept 422 RMB-522 RMB-523

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -202,121 +202,144 @@ private static void endWithXHeaders(HttpClientRequest req, Map<String, String> h
...
+            return null;
+          }
+          break;
+        case RAW_PUT:
+        case RAW_POST:
+          break;
+      }
+      return id;
+    } catch (Exception ex) {
+      f.handle(Future.failedFuture("exception: " + ex.getLocalizedMessage()));
```
julianladisch: This swallows the stacktrace and does not add information to the message. My suggestion:
```suggestion
      f.handle(Future.failedFuture(ex));
```
Note: ex.getLocalizedMessage() uses the locale of the JVM, this is not the locale of the client; therefore ex.getMessage() should be used instead.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -202,121 +202,144 @@ private static void endWithXHeaders(HttpClientRequest req, Map<String, String> h
...
+      return null;
+    }
+  }
+
+  private static void handleException(Throwable ex, HttpMethod method, String uri, Future<Void> f) {
+    String diag = method.name() + " " + uri + ": " + ex.getMessage();
+    log.error(diag);
```
julianladisch: The stacktrace should be logged as well:
```suggestion
    log.error(diag, ex);
```

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
```diff
@@ -268,15 +268,36 @@ public void testPutFailure(TestContext context) {
 
     TenantLoading tl = new TenantLoading();
     tl.addJsonIdContent("loadRef", "tenant-load-ref", "data", "data");
-    putStatus = 422;
+    putStatus = 423;
```
julianladisch: This number looks like a typo. Can you use 500 instead?

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
```diff
@@ -268,15 +268,36 @@ public void testPutFailure(TestContext context) {
...
+    TenantAttributes tenantAttributes = new TenantAttributes()
+      .withModuleTo("mod-1.0.0")
+      .withParameters(parameters);
+    Map<String, String> headers = new HashMap<String, String>();
+    headers.put("X-Okapi-Url-to", "http://localhost:" + Integer.toString(port));
+
+    TenantLoading tl = new TenantLoading();
+    tl.addJsonIdContent("loadRef", "tenant-load-ref", "data", "data");
+    putStatus = 423;
+    tl.withAcceptStatus(423);
```
julianladisch: This number looks like a typo. Can you use 500 instead?

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -202,121 +202,144 @@ private static void endWithXHeaders(HttpClientRequest req, Map<String, String> h
...
-      }
-      log.warn(method1.name() + " " + putUri.toString() + ": " + ex.getMessage());
-    });
-    endWithXHeaders(reqPut, headers, content);
+        resPut.exceptionHandler(ex -> handleException(ex, method1, putUri.toString(), f));
+      });
+      reqPut.exceptionHandler(ex -> handleException(ex, method1, putUri.toString(), f));
+      endWithXHeaders(reqPut, headers, content);
+    } catch (Exception ex) {
+      log.error(ex.getLocalizedMessage(), ex);
```
julianladisch:
```suggestion
      log.error(ex.getMessage(), ex);
```

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -202,121 +202,144 @@ private static void endWithXHeaders(HttpClientRequest req, Map<String, String> h
...
+      f.handle(Future.failedFuture("exception: " + ex.getLocalizedMessage()));
```
julianladisch:
```suggestion
      f.handle(Future.failedFuture(ex));
```

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
```diff
@@ -268,15 +268,36 @@ public void testPutFailure(TestContext context) {
...
-  public void testOkStatusCode(TestContext context) {
+  public void testokWithAcceptStatus(TestContext context) {
+    Async async = context.async();
+    List<Parameter> parameters = new LinkedList<>();
+    parameters.add(new Parameter().withKey("loadRef").withValue("true"));
+    TenantAttributes tenantAttributes = new TenantAttributes()
+      .withModuleTo("mod-1.0.0")
+      .withParameters(parameters);
+    Map<String, String> headers = new HashMap<String, String>();
+    headers.put("X-Okapi-Url-to", "http://localhost:" + Integer.toString(port));
```
julianladisch:
```suggestion
    headers.put("X-Okapi-Url-to", "http://localhost:" + port);
```

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
```diff
@@ -268,15 +268,36 @@ public void testPutFailure(TestContext context) {
...
+
+    TenantLoading tl = new TenantLoading();
+    tl.addJsonIdContent("loadRef", "tenant-load-ref", "data", "data");
+    putStatus = 423;
+    tl.withAcceptStatus(423);
+    tl.perform(tenantAttributes, headers, vertx, res -> {
+      context.assertTrue(res.succeeded());
+      async.complete();
+    });
```
julianladisch: `Async async` is not needed:
tl.perform(tenantAttributes, headers, vertx, context.asyncAssertSuccess())

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantLoadingTest.java
adamdickmeiss: I believe this is resolved now.

julianladisch: The unneccessary call to Integer.toString is still there, but it is a minor issue.

PR #573, opened 2019-11-26
RMB-471: Fix ddlgen/Index.java StringBuilder code smells

julianladisch:
* Remove unneccessary use of temporary StringBuilder
* Replace duplicate String constant by final variable
* Convert to guard clauses to avoid nesting/indentation
* Use line breaks for .append clauses for readability
* Reword comments
* Code formatting to comply with coding standards

### DIFF
--- a/dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
+++ b/dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
@@ -24,24 +24,31 @@ public class Index extends TableIndexes {
   public boolean isCaseSensitive() {
     return caseSensitive;
   }
+
   public void setCaseSensitive(boolean caseSensitive) {
     this.caseSensitive = caseSensitive;
   }
+
   public String getWhereClause() {
     return whereClause;
   }
+
   public void setWhereClause(String whereClause) {
     this.whereClause = whereClause;
   }
+
   public boolean isStringType() {
     return stringType;
   }
+
   public void setStringType(boolean stringType) {
     this.stringType = stringType;
   }
+
   public boolean isRemoveAccents() {
     return removeAccents;
   }
+
   public void setRemoveAccents(boolean removeAccents) {
     this.removeAccents = removeAccents;
   }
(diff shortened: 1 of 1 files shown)


dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -76,41 +77,44 @@ public void setSqlExpression(String sqlExpression) {
...
-      for (int i = 0;i < splitIndex.length;i++) {
-        if(i != 0) {
-          result .append(" , ");
-        }
-        appendExpandedTerm(tableLoc, splitIndex[i],result);
+    String [] splitIndex = this.getMultiFieldNames().split(" *, *");
+
+    StringBuilder result = new StringBuilder("concat_space_sql(");
+    for (int i = 0; i < splitIndex.length; i++) {
+      if(i != 0) {
```
adamdickmeiss: I do like a space between if and the (.. Just me being pedantic

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -76,41 +77,44 @@ public void setSqlExpression(String sqlExpression) {
...
+      result.append(table).append(".").append(Cql2PgUtil.cqlNameAsSqlText(JSONB, term));
       return;
     }
-    //case where the [*] is found
-    if(idx == term.length() - arrayToken.length()) {
-      expandedTerm.append(table).append(".").append(Cql2PgUtil.cqlNameAsSqlJson("jsonb",term.substring(0,idx)));
-      result.append("concat_array_object(").append(expandedTerm).append(")");
+    //case where [*] is found at the end
+    if(idx == term.length() - ARRAY_TOKEN.length()) {
+      result.append("concat_array_object(")
```
adamdickmeiss: And here as well

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
julianladisch: Fixed all code formatting issues.

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
julianladisch: Fixed all code formatting issues.

PR #576, opened 2019-11-28
Allow moduleVersionFrom with version only RMB-528 RMB-404

adamdickmeiss:
Use Okapi's SemVer and ModuleId for dealing with module IDs and
semantic versions.


adamdickmeiss:
I'll see what I can do on OKAPI-781.

PR #578, opened 2019-12-04
Fix Criterion Get does not select id RMB-532

adamdickmeiss:


adamdickmeiss:
Yes.. There should be some test for this.. But frankly, don't have time today. I can assure this fixes unit tests for mod-configuration .. which has zero changes in code from version 28 with this in.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2108,7 +2108,7 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
     List<FacetField> facets, Handler<AsyncResult<Results<T>>> replyHandler) {
 
     get(conn, table, clazz, DEFAULT_JSONB_FIELD_NAME, filter, returnCount,
-      false, setId, facets, replyHandler);
+      true, setId, facets, replyHandler);
```
julianladisch: Can you explain why this is required? returnIdField=true is deprecated since RMB 25.0.0 (RMB-346) because the id field is always copied into jsonb->'id'. The client should use jsonb->'id' and not that id internal database column.

adamdickmeiss: Remember that this is Criterion, not CQL .. I've updated the issue https://issues.folio.org/browse/RMB-532
with more description

adamdickmeiss: I should of course make a test case.. But this is not done for the 29.1.1 release where the single line change was made.

adamdickmeiss: Regressions like this illustrates that it is problematic with multiple "query" languages.. Really we should have CQL only IMHO

julianladisch: I've fixed the real cause: setId=true.
And reverted this hack.

PR #580, opened 2019-12-10
RMB-533: Check 5 index types

julianladisch:


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -871,21 +868,20 @@ String queryByFt(String indexText, String term, String comparator, Index schemaI
...
     String comparator, CqlModifiers modifiers, Table targetTable) throws QueryValidationException {
 
     String indexText = vals.getIndexText();
     String sql = null;
+    Index schemaIndex;
 
     List<Modifier> relationModifiers = modifiers.getRelationModifiers();
     if (!relationModifiers.isEmpty()) {
-      final Index schemaIndex = DbSchemaUtils.getIndex(index, this.dbTable.getGinIndex());
+       schemaIndex = dbIndex.getGinIndex();  // FIXME
```
adamdickmeiss: Shouldn't the schemaIndex below which consider all types be moved before the relationModifiers check?

adamdickmeiss: I mean so that assignment is just done once.

julianladisch: Yes, the FIXME indicates that this pull request doesn't fix a similar problem for the arrayNode method.
It only fixes it for simple fields (non-array).

You explicitly requested that this bug fix should be as small as possible because we want to backport it and want to reduce any side effects of the backport.

Do you want a fix for arrayNode as well?

adamdickmeiss: Wasn't aware of problems with arrayNode as long as it got an Index schemaIndex.
Most certainly, I can see that of the existing tests passes if the assignment of Index is done once.

adamdickmeiss: Looking at arrayNode, I don't see why it has anything to do with the problem. It uses FT for the relation modifiers, because ONLY = is allowed anyway and makes a recursion wihtout modifiers, so ends up in queryByLike recursively (or queryByFT if =).

julianladisch: I'm not convinced that the requested change improves anything, but it doesn't seem to do any harm so I've put that it.

PR #583, opened 2019-12-13
RMB-536: Missing f_unaccent for compound fields full text index

evaluk:
difference in generation and query resolved by pushing wrapping into the index function.

### DIFF
--- a/dbschema/src/main/java/org/folio/cql2pgjson/util/Cql2PgUtil.java
+++ b/dbschema/src/main/java/org/folio/cql2pgjson/util/Cql2PgUtil.java
@@ -250,6 +250,7 @@ public final class Cql2PgUtil {
     }
     result.append('\'');
   }
+
   public static String wrapInLowerUnaccent(String term, boolean lower, boolean unaccent) {
     if (lower) {
       if (unaccent) {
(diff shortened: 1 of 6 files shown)


dbschema/src/main/java/org/folio/cql2pgjson/util/Cql2PgUtil.java
```diff
@@ -250,4 +250,21 @@ public static void appendQuoted(CharSequence s, int start, int end, StringBuilde
...
+    } else {
+      if (unaccent) {
+        return "f_unaccent(" + term + ")";
+      } else {
+        return term;
+      }
+    }
+  }
+
+
```
julianladisch: Can you remove at least one empty line?

dbschema/src/main/java/org/folio/cql2pgjson/util/Cql2PgUtil.java
```diff
@@ -250,4 +250,21 @@ public static void appendQuoted(CharSequence s, int start, int end, StringBuilde
     }
     result.append('\'');
   }
+  public static String wrapInLowerUnaccent(String term, boolean lower, boolean unaccent) {
```
julianladisch: Can you insert an empty line between methods?

PR #588, opened 2019-12-18
RMB-462: Drop old index for fullTextIndex dictionary change (29.1 backport)

julianladisch:
Merge pull request #582 from folio-org/RMB-462-fullTextIndex-dictionary


julianladisch:
Not needed in b29.1

PR #602, opened 2020-01-09
RMB-550: Drop all indexes on module upgrade

julianladisch:


julianladisch:
https://github.com/folio-org/raml-module-builder#the-post-tenant-api
"tOps": "DELETE" indicates that the index should be deleted. Module developers need to take care that they change "tOps": "ADD" to "tOps": "DELETE" and do not completely remove the index entry from schema.json. (RMB-552 will catch that.)

hjiebsco:
> https://github.com/folio-org/raml-module-builder#the-post-tenant-api
> "tOps": "DELETE" indicates that the index should be deleted. Module developers need to take care that they change "tOps": "ADD" to "tOps": "DELETE" and do not completely remove the index entry from schema.json. (RMB-552 will catch that.)

That's right. Forgot that.

PR #608, opened 2020-01-23
Rmb 500 stream get with total count

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
...
+    if (!transaction) {
+      connection.close();
+    }
+  }
+
   @SuppressWarnings({"unchecked", "squid:S00107"})
-  private <T> void doStreamGet(SQLConnection connection, boolean transactionMode, String table, Class<T> clazz,
-      String fieldName, CQLWrapper wrapper, boolean returnIdField, String distinctOn, List<FacetField> facets,
-      Handler<T> streamHandler, Handler<AsyncResult<Void>> replyHandler) {
+  private <T> void doStreamGet(SQLConnection connection, boolean transactionMode,
```
julianladisch: Can you avoid using transactionMode? We have closeAndHandleResult. The caller of a method should be in charge for both opening and closing the connection. Closing the connection in the method is error-prone because there are several places and one can easily miss one.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
    * @param replyHandler
    */
   @SuppressWarnings({"unchecked", "squid:S00107"})
```
julianladisch: Can you add a comment what "squid:S00107" means?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
...
+
+  public <T> void streamGet(String table, T entity, String fieldName,
+    CQLWrapper filter, boolean returnIdField, String distinctOn,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
+
+    streamGet(table, entity, fieldName, filter, returnIdField, distinctOn,
+      Collections.emptyList(), replyHandler);
+  }
+
+  @SuppressWarnings({"unchecked", "squid:S00107"})
```
julianladisch: Can you add javadoc?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
    * @param replyHandler
    */
   @SuppressWarnings({"unchecked", "squid:S00107"})
-  public <T> void streamGet(String table, T entity, String fieldName, CQLWrapper filter, boolean returnIdField,
-      String distinctOn, Handler<T> streamHandler, Handler<AsyncResult<Void>> replyHandler) {
+  public <T> void streamGet(String table, T entity, String fieldName,
```
julianladisch: Can you add javadoc whether this is legacy?
Is this method deprecated? If yes add deprecation javadoc and annotation.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
...
+    CQLWrapper filter, boolean returnIdField, String distinctOn,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
+
+    streamGet(table, entity, fieldName, filter, returnIdField, distinctOn,
+      Collections.emptyList(), replyHandler);
+  }
+
+  @SuppressWarnings({"unchecked", "squid:S00107"})
+  public <T> void streamGet(String table, T entity, String fieldName,
```
julianladisch: <T> entity is not used. Can you replace it by Class<T> clazz?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
...
+  }
+
   @SuppressWarnings({"unchecked", "squid:S00107"})
-  private <T> void doStreamGet(SQLConnection connection, boolean transactionMode, String table, Class<T> clazz,
-      String fieldName, CQLWrapper wrapper, boolean returnIdField, String distinctOn, List<FacetField> facets,
-      Handler<T> streamHandler, Handler<AsyncResult<Void>> replyHandler) {
+  private <T> void doStreamGet(SQLConnection connection, boolean transactionMode,
+    String table, Class<T> clazz, String fieldName, CQLWrapper wrapper,
+    boolean returnIdField, String distinctOn, List<FacetField> facets,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
```
julianladisch: Can you use PostgresClientStreamResult<T>?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1609,79 +1609,113 @@ public TotaledResults(ResultSet set, Integer total) {
...
-              replyHandler.handle(Future.succeededFuture());
-            });
+          Map<String, Method> externalColumnSetters = getExternalColumnSetters(resultsHelper.columnNames,
+            resultsHelper.clazz, isAuditFlavored);
 
-          } else {
-            if (!transactionMode) {
-              connection.close();
+          sqlRowStream.resultSetClosedHandler(v -> sqlRowStream.moreResults()).handler(r -> {
+            JsonObject row = convertRowStreamArrayToObject(sqlRowStream, r);
```
julianladisch: Can you move this into the try clause below?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClientStreamResult.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public PostgresClientStreamResult(ResultInfo resultInfo) {
+    this.resultInfo = resultInfo;
+  }
+
+  ResultInfo resultInto() {
+    return this.resultInfo;
+  }
+
+  public void handler(Handler<T> streamHandler) {
```
julianladisch: Can you change `void` to `PostgresClientStreamResult` and return `this` for chaining calls (fluent API)? (all 3 setters)
Can you add javadoc? (all 3 setters)

adamdickmeiss:
Thx for your feedback which, as far as I can tell, all makes sense.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+   * @param clazz
+   * @param fieldName
+   * @param filter
+   * @param returnIdField
+   * @param distinctOn may be null
+   * @param replyHandler AsyncResult; on success with result {@link org.folio.rest.persist.PostgresClientStreamResult}
+   */
+  public <T> void streamGet(String table, Class<T> clazz, String fieldName,
+    CQLWrapper filter, boolean returnIdField, String distinctOn,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
```
julianladisch: Can you replace `<PostgresClientStreamResult>` by `<PostgresClientStreamResult<T>>`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+   * @param filter
+   * @param returnIdField must be true if facets are in passed
+   * @param distinctOn may be null
+   * @param facets may not be null (Collections.emptyList() for no facets)
+   * @param replyHandler AsyncResult; on success with result {@link org.folio.rest.persist.PostgresClientStreamResult}
+   */
+  @SuppressWarnings({"unchecked", "squid:S00107"})    // Method has >7 parameters
+  public <T> void streamGet(String table, Class<T> clazz, String fieldName,
+    CQLWrapper filter, boolean returnIdField, String distinctOn,
+    List<FacetField> facets, Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
```
julianladisch: Can you replace `<PostgresClientStreamResult>` by `<PostgresClientStreamResult<T>>`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+   * @param returnIdField
+   * @param distinctOn
+   * @param facets
+   * @param replyHandler
+   */
+  @SuppressWarnings({"unchecked", "squid:S00107"})    // Method has >7 parameters
+  private <T> void doStreamGet(AsyncResult<SQLConnection> connResult,
+    String table, Class<T> clazz, String fieldName, CQLWrapper wrapper,
+    boolean returnIdField, String distinctOn, List<FacetField> facets,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
```
julianladisch: Can you replace `<PostgresClientStreamResult>` by `<PostgresClientStreamResult<T>>`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+   * @param returnIdField
+   * @param distinctOn
+   * @param facets
+   * @param replyHandler
+   */
+  @SuppressWarnings({"unchecked", "squid:S00107"})    // Method has >7 parameters
+  private <T> void doStreamGetCount(SQLConnection connection,
+    String table, Class<T> clazz, String fieldName, CQLWrapper wrapper,
+    boolean returnIdField, String distinctOn, List<FacetField> facets,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
```
julianladisch: Can you replace `<PostgresClientStreamResult>` by `<PostgresClientStreamResult<T>>`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+      });
+    } catch (Exception e) {
+      log.error(e.getMessage(), e);
+      replyHandler.handle(Future.failedFuture(e));
+    }
+  }
 
+  private <T> void doStreamGetQuery(SQLConnection connection, String selectQuery,
+    ResultInfo resultInfo, Class<T> clazz, List<FacetField> facets,
+    Handler<AsyncResult<PostgresClientStreamResult>> replyHandler) {
```
julianladisch: Can you replace `<PostgresClientStreamResult>` by `<PostgresClientStreamResult<T>>`?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
     createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
     postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", null, false, null,
-        streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
-          context.assertEquals(6, objectCount.get());
+      streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        context.assertEquals(6, objectCount.get());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertSuccess` instead? Example:

    @Test
    public void streamGetLegacy(TestContext context) throws IOException {
      AtomicInteger objectCount = new AtomicInteger();

      final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";

      createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
      postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", null, false, null,
            streamHandler -> objectCount.incrementAndGet(), context.asyncAssertSuccess(asyncResult -> {
            context.assertEquals(6, objectCount.get());
          }));
    }

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
+    postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", wrapper, false, null,
+      streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        context.assertEquals(3, objectCount.get());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertSuccess` instead?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=");
+    Async async = context.async();
+    AtomicInteger objectCount = new AtomicInteger();
+    postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", wrapper, false, null,
+      streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
+        context.assertTrue(asyncResult.failed());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertFailure` instead?
Can you either test the value of `objectCount` or remove that variable?

    @Test
    public void streamGetLegacySyntaxError(TestContext context) throws IOException, FieldException {
      final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";
      createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
      CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=");
      AtomicInteger objectCount = new AtomicInteger();
      postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", wrapper, false, null,
        streamHandler -> objectCount.incrementAndGet(), context.asyncAssertFailure(asyncResult -> {
          context.assertEquals(0, objectCount.get());
        }));
    }

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
+    Async async = context.async();
+    AtomicInteger objectCount = new AtomicInteger();
+    postgresClient.streamGet("noSuchTable", new Object(), "jsonb", wrapper, false, null,
+      streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
+        context.assertTrue(asyncResult.failed());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertFailure` instead?
Can you either test the value of `objectCount` or remove that variable?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
+    Async async = context.async();
+    postgresClient.streamGet("noSuchTable", Object.class, "jsonb", wrapper, false, null,
+      asyncResult -> {
+        context.assertTrue(asyncResult.failed());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertFailure()` instead?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, false, null,
+      asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        PostgresClientStreamResult sr = asyncResult.result();
+        context.assertEquals(3, sr.resultInto().getTotalRecords());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertFailure` instead?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, false, null,
+      asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        PostgresClientStreamResult sr = asyncResult.result();
+        context.assertEquals(3, sr.resultInto().getTotalRecords());
+        sr.handler(streamHandler -> objectCount.incrementAndGet());
+        sr.endHandler(x -> {
```
julianladisch: Can you add

            sr.exceptionHandler(exception -> context.fail(exception));

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
     Async async = context.async();
 
     final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";
 
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=Millenium edition");
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, false, null,
+      asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        PostgresClientStreamResult sr = asyncResult.result();
```
julianladisch: Can you replace these three lines by

    context.asyncAssertSuccess(sr -> {

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    facets.add(new FacetField("jsonb->>'edition'"));
+    facets.add(new FacetField("jsonb->>'title'"));
     createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
     CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
-    postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", wrapper, false, null,
-        streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, true, null,
+      facets, asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        PostgresClientStreamResult sr = asyncResult.result();
```
julianladisch: Can you replace these three lines by

    context.asyncAssertSuccess(sr -> {

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
     final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";
+    List<FacetField> facets = new ArrayList<FacetField>();
+    facets.add(new FacetField("jsonb->>'edition'"));
+    facets.add(new FacetField("jsonb->>'title'"));
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=Millenium edition");
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, true, null,
+      facets, asyncResult -> {
+        context.assertTrue(asyncResult.succeeded());
+        PostgresClientStreamResult sr = asyncResult.result();
```
julianladisch: Can you replace these three lines by

    context.asyncAssertSuccess(sr -> {

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
     createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
-    postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", null, false, "jsonb->>'edition'",
-        streamHandler -> objectCount.incrementAndGet(), asyncResult -> {
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=First edition");
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, true, null,
+      badFacets, asyncResult -> {
+        context.assertTrue(asyncResult.failed());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and use `context.asyncAssertFailure()` instead?
Can you either remove `objectCount` or check the value?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,51 +2222,447 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=");
+    Async async = context.async();
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, false, null,
+      asyncResult -> {
+        context.assertTrue(asyncResult.failed());
+        async.complete();
+      });
+    async.awaitSuccess();
```
julianladisch: Can you remove `Async async` and replace it by `context.asyncAssertFailure()´?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
-      }
-    });
+    Class<T> clazz = (Class<T>) entity.getClass();
+    streamGet(table, clazz, fieldName, filter, returnIdField, distinctOn,
+      res -> {
+        if (res.failed()) {
+          replyHandler.handle(Future.failedFuture(res.cause()));
+          return;
+        }
+        PostgresClientStreamResult streamResult = res.result();
```
julianladisch: Can you change `PostgresClientStreamResult` to `PostgresClientStreamResult<T>`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+    ResultInfo resultInfo, Class<T> clazz, List<FacetField> facets,
+    Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
+
+    connection.queryStream(selectQuery, stream -> {
+      if (stream.failed()) {
+        log.error(stream.cause().getMessage(), stream.cause());
+        replyHandler.handle(Future.failedFuture(stream.cause()));
+        return;
+      }
+      PostgresClientStreamResult streamResult = new PostgresClientStreamResult(resultInfo);
```
julianladisch: Can you use `PostgresClientStreamResult<T> streamResult = new PostgresClientStreamResult<>(resultInfo);`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+
+      SQLRowStream sqlRowStream = stream.result();
+      Promise promise = Promise.promise();
```
julianladisch: Can you use `Promise<PostgresClientStreamResult<T>> promise = Promise.promise();`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+      ResultsHelper<T> resultsHelper = new ResultsHelper<>(sqlRowStream, clazz);
+      boolean isAuditFlavored = isAuditFlavored(resultsHelper.clazz);
+      Map<String, Method> externalColumnSetters = getExternalColumnSetters(resultsHelper.columnNames,
+        resultsHelper.clazz, isAuditFlavored);
+      // fetch rows and produce ResultInfo when facets are all read
+      sqlRowStream.resultSetClosedHandler(v -> sqlRowStream.moreResults()).handler(r -> {
+        JsonObject row = convertRowStreamArrayToObject(sqlRowStream, r);
+        try {
+          Object objRow = null;
```
julianladisch: Can you use `T objRow = null;`?

adamdickmeiss: And add cast later? Isn't deserializeRow wrong..?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+          // deserializeRow can not determine if count or user object
+          // in case where user T=Object
+          // skip the initial count result when facets are in use
+          if (resultsHelper.offset == 0 && !facets.isEmpty()) {
+            resultsHelper.facet = true;
           } else {
-            if (!transactionMode) {
-              connection.close();
+            objRow = deserializeRow(resultsHelper, externalColumnSetters, isAuditFlavored, row);
```
julianladisch: Can you use `objRow = (T) deserializeRow(` to clarify what type we process?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: I'll add the cast.. deserializeRow should be changed soon.. It's a mess. IMHO

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Yep

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,207 @@ public TotaledResults(ResultSet set, Integer total) {
...
+
+      SQLRowStream sqlRowStream = stream.result();
+      Promise<PostgresClientStreamResult<T>> promise = Promise.promise();
+      ResultsHelper<T> resultsHelper = new ResultsHelper<>(sqlRowStream, clazz);
+      boolean isAuditFlavored = isAuditFlavored(resultsHelper.clazz);
+      Map<String, Method> externalColumnSetters = getExternalColumnSetters(resultsHelper.columnNames,
+        resultsHelper.clazz, isAuditFlavored);
+      // fetch rows and produce ResultInfo when facets are all read
+      sqlRowStream.resultSetClosedHandler(v -> sqlRowStream.moreResults()).handler(r -> {
+        JsonObject row = convertRowStreamArrayToObject(sqlRowStream, r);
```
julianladisch: Can you move this line (`JsonObject row = convertRowStreamArrayToObject(...`= into the `try` clause below so that any exception that `convertRowStreamArrayToObject` throws will get caught and handled?

adamdickmeiss: Sure

adamdickmeiss:
Other question.. Should we guard against many exceptions thrown.. I was thinking that perhaps only the first exception should be fired.. And rfest ignored.. There could be many identical exceptions if it happens inside a user handler.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,54 +2222,377 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+        sr.endHandler(x -> {
+          events.append("[endHandler]");
+          async.complete();
+        });
+        sr.exceptionHandler(x -> {
+          events.append("[exception]");
+        });
+      }));
+    async.awaitSuccess();
+    context.assertEquals("[handler][exception][handler][exception][handler][exception][endHandler]", events.toString());
```
julianladisch: This looks strange to me.
After the exception handler is called I do not expect any other handlers being called, neither the item handler nor the end handler.
If you really want that this may happen it should be clearly documented so that the using modules can take care of it (for example by setting a flag that ignores item handler and end handler invocations).

If the exception handler is the last handler that gets invoked the end handler and the exception handler can be merged into a `Handler<AsyncResult<Void>>`.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClientStreamResult.java
```diff
@@ -0,0 +1,110 @@
...
+    return this.resultInfo;
+  }
+
+  /**
+   * Sets stream handler for each returned item of type T
+   *
+   * @param streamHandler
+   * @return this instance (fluent)
+   */
+  public PostgresClientStreamResult<T> handler(Handler<T> streamHandler) {
```
julianladisch: Can you either
* rename `handler` to a more descriptive name, for example `itemHandler`, or
* implement some interface where that name is needed (https://github.com/eclipse-vertx/vertx-sql-client/blob/3.8.5/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowStreamImpl.java#L59 implements `io.vertx.core.streams.ReadStream<T>`)

adamdickmeiss: It was modelled after ReadStream<T>, so I'll PostgresClientStreamResult will implement that.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: Agreed. I'll change that, then.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,210 @@ public TotaledResults(ResultSet set, Integer total) {
...
+   * @param returnIdField
+   * @param distinctOn may be null
+   * @param streamHandler called for each record
+   * @param replyHandler called when query is complete
+   * @deprecated This function is deprecated because it is impossible to avoid it has
+   * no callback for the case where the HTTP headers are generated.
+   * Furthermore, the hit count is not passed.
+   * Thus, the only option, is to use the replyHandler to produce this.
+   * Use streamGet with {@link PostgresClientStreamResult} instead.
+   * {@link #streamGet(java.lang.String, java.lang.Object, java.lang.String, org.folio.rest.persist.cql.CQLWrapper, boolean, java.lang.String, io.vertx.core.Handler, io.vertx.core.Handler)}
```
julianladisch: Can you wrap this long line?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,210 @@ public TotaledResults(ResultSet set, Integer total) {
...
+    Class<T> clazz = (Class<T>) entity.getClass();
+    streamGet(table, clazz, fieldName, filter, returnIdField, distinctOn,
+      res -> {
+        if (res.failed()) {
+          replyHandler.handle(Future.failedFuture(res.cause()));
+          return;
+        }
+        PostgresClientStreamResult<T> streamResult = res.result();
+        streamResult.handler(streamHandler);
+        streamResult.endHandler(x -> replyHandler.handle(Future.succeededFuture()));
```
julianladisch: Can you add

            streamResult.exceptionHandler(e -> replyHandler.handle(Future.failedFuture(e)));

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,210 @@ public TotaledResults(ResultSet set, Integer total) {
...
    * @param table
    * @param entity
+   * @param fieldName usually "jsonb"
+   * @param filter usually CQL query
+   * @param returnIdField
+   * @param distinctOn may be null
+   * @param streamHandler called for each record
+   * @param replyHandler called when query is complete
+   * @deprecated This function is deprecated because it is impossible to avoid it has
+   * no callback for the case where the HTTP headers are generated.
```
julianladisch: Can you reword this sentence?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,210 @@ public TotaledResults(ResultSet set, Integer total) {
...
+            replyHandler.handle(promise.future());
+          }
+          sqlRowStream.close();
+          log.error(e.getMessage(), e);
+          streamResult.fireExceptionHandler(e);
         }
-        log.error(e.getMessage(), e);
-        replyHandler.handle(Future.failedFuture(e));
-      }
+      }).endHandler(v2 -> {
```
julianladisch: Can you add .exceptionHandler(e -> ... ?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClientStreamResult.java
```diff
@@ -0,0 +1,120 @@
...
+   */
+  @Override
+  public PostgresClientStreamResult<T> endHandler(Handler<Void> endHandler) {
+    this.endHandler = endHandler;
+    return this;
+  }
+
+  /**
+   * Sets handler to be called if an exception occurs Both exceptions in user
+   * handlers and PostgresClient are caught.
```
julianladisch: Can you add that after the exception has been called neither the item handler nor the end handler will be called?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,52 +2222,435 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+  }
+
+  @Test
+  public void streamGetLegacySyntaxError(TestContext context) throws IOException, FieldException {
+    final String tableDefiniton = "id UUID PRIMARY KEY , jsonb JSONB NOT NULL, distinct_test_field TEXT";
+    createTableWithPoLines(context, MOCK_POLINES_TABLE, tableDefiniton);
+    CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON("jsonb"), "edition=");
+    AtomicInteger objectCount = new AtomicInteger();
+    postgresClient.streamGet(MOCK_POLINES_TABLE, new Object(), "jsonb", wrapper,
+      false, null, streamHandler -> objectCount.incrementAndGet(), context.asyncAssertFailure());
```
julianladisch: Can you replace `objectCount.incrementAndGet()` by `context.fail()` and remove `objectCount`?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,52 +2222,435 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+          context.assertEquals(3, objectCount.get());
+          async.complete();
+        });
+      }));
+    async.await();
+  }
+
+  @Test
+  public void streamGetUnsupported(TestContext context) throws IOException, FieldException {
     AtomicInteger objectCount = new AtomicInteger();
```
julianladisch: Can you remove this unused variable?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,52 +2222,435 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+    postgresClient.streamGet(MOCK_POLINES_TABLE, Object.class, "jsonb", wrapper, false, null,
+      context.asyncAssertSuccess(sr -> {
+        String message;
+        message = null;
+        try {
+          sr.pause();
+        } catch (Exception ex) {
+          message = ex.getMessage();
+        }
+        context.assertEquals("Not supported yet: pause", message);
```
julianladisch: The try-catch-idiom is more easy to read and understand: https://github.com/junit-team/junit4/wiki/Exception-testing#trycatch-idiom

        try {
          sr.pause();
          context.fail();
        } catch (Exception ex) {
          assertThat(ex.getMessage(), is("Not supported yet: pause"));
        }

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2219,52 +2222,435 @@ public void selectDistinctOn(TestContext context) throws IOException {
...
+        message = null;
+        try {
+          sr.fetch(0);
+        } catch (Exception ex) {
+          message = ex.getMessage();
+        }
+        context.assertEquals("Not supported yet: fetch", message);
+        async.complete();
+      }));
+    async.await();
```
julianladisch: Can you remove this async?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Good catch!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,217 @@ public TotaledResults(ResultSet set, Integer total) {
...
    * @param entity
+   * @param fieldName usually "jsonb"
+   * @param filter usually CQL query
+   * @param returnIdField
+   * @param distinctOn may be null
+   * @param streamHandler called for each record
+   * @param replyHandler called when query is complete
+   * @deprecated This function is deprecated because either you'll have to
+   * buffer whole HTTP buffer in memory to produce HTTP status; or you'll have to
+   * return a fake error. Furthermore, this API does provide totalCount
```
julianladisch:
```suggestion
   * return a fake error. Furthermore, this API does not provide totalCount
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,217 @@ public TotaledResults(ResultSet set, Integer total) {
...
+   * @param distinctOn
+   * @param facets
+   * @param replyHandler
+   */
+  @SuppressWarnings({"unchecked", "squid:S00107"})    // Method has >7 parameters
+  private <T> void doStreamGet(AsyncResult<SQLConnection> connResult,
+    String table, Class<T> clazz, String fieldName, CQLWrapper wrapper,
+    boolean returnIdField, String distinctOn, List<FacetField> facets,
+    Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
+    if (connResult.failed()) {
```
julianladisch: Can you insert a blank line to separate the argument list from the code?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1596,91 +1595,217 @@ public TotaledResults(ResultSet set, Integer total) {
...
+      log.error(connResult.cause().getMessage(), connResult.cause());
+      replyHandler.handle(Future.failedFuture(connResult.cause()));
+      return;
```
julianladisch: Can you write a unit test for this connResult.failed() case?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClientStreamResult.java
```diff
@@ -0,0 +1,121 @@
...
+  @Override
+  public PostgresClientStreamResult<T> endHandler(Handler<Void> endHandler) {
+    this.endHandler = endHandler;
+    return this;
+  }
+
+  /**
+   * Sets handler to be called if an exception occurs Both exceptions in user
+   * handlers and PostgresClient are caught. This handler will be called only once
+   * and neither the item handler, nor the endHandler will be called.
```
julianladisch:
```suggestion
   * and neither the item handler, nor the endHandler will be called afterwards.
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Good catch. I think, in general, it's a good idea - for multiline declarations like that to have a blank line.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: ha!

PR #614, opened 2020-02-14
Rmb-559 2 PgUtil.streamGet

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -458,6 +466,127 @@ public static void deleteById(String table, String id,
...
+      List<Diagnostic> diag = new ArrayList<>();
+      diag.add(new Diagnostic().withCode("500").withMessage(message));
+      result.resultInto().setDiagnostics(diag);
+      streamTrailer(response, result.resultInto());
+    });
+    result.endHandler(res -> streamTrailer(response, result.resultInto()));
+    result.handler(res -> {
+      if (cnt[0]++ > 0) {
+        response.write(String.format(",%n"));
+      }
```
julianladisch: The count number is not needed. We only need to know whether it is the first or any other record. Can you change to a `first` boolean variable?
The array is misleading because there is a single value only. Can you change this hack to an Atomic... variable?

    AtomicBoolean first = new AtomicBoolean(true);

...

      if (first.get()) {
        first.set(false);
      } else {
        response.write(String.format(",%n"));
      }

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -458,6 +466,127 @@ public static void deleteById(String table, String id,
...
+      try {
+        response.write(OBJECT_MAPPER.writeValueAsString(res));
+      } catch (JsonProcessingException ex) {
+        logger.error(ex.getCause(), ex);
```
julianladisch:
```suggestion
        logger.error(ex.getMessage(), ex);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -458,6 +466,127 @@ public static void deleteById(String table, String id,
...
+    String element, RoutingContext routingContext, Map<String, String> okapiHeaders,
+    Context vertxContext) {
+
+    HttpServerResponse response = routingContext.response();
+    try {
+      List<FacetField> facetList = FacetManager.convertFacetStrings2FacetFields(facets, JSON_COLUMN);
+      CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON(table + "." + JSON_COLUMN), cql, limit, offset);
+      streamGet(table, clazz, wrapper, facetList, element, routingContext, okapiHeaders, vertxContext);
+    } catch (Exception e) {
+      logger.info("e={}", e.getCause(), e);
```
julianladisch:
```suggestion
      logger.error(e.getMessage(), e);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -458,6 +466,127 @@ public static void deleteById(String table, String id,
...
+   * "required": [
+   *   "instances",
+   *   "totalRecords"
+   * ]
+   *</pre>
+   * @param <T> Class for each item returned
+   * @param table SQL table
+   * @param clazz The item class
+   * @param cql CQL query
+   * @param offset offset >= 0; < 0 to ignore
```
julianladisch:
```suggestion
   * @param offset offset >= 0; < 0 for no offset
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -458,6 +466,127 @@ public static void deleteById(String table, String id,
...
+   * @param limit  limit >= 0 ; <0 to ignore
```
julianladisch:
```suggestion
   * @param limit  limit >= 0 ; < 0 for no limit
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: OK with that, I assume you also considered the shoter first.getAndSet ..
I also see that if there's an exception a comma will be omitted.. which generates invalid JSON

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: Yep

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: Yep

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: sure

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -170,7 +155,87 @@ private void postBook(TestContext context, String parameterString, int expectedS
...
+    // need at least one record in result before we can trigger this error
+    buf = checkURLs(context, "http://localhost:" + port + "/rmbtests/test?query=badclass%3Dtrue", 200);
+    books = Json.decodeValue(buf, Books.class);
+    context.assertEquals(2, books.getTotalRecords());
+    context.assertEquals(1, books.getResultInfo().getDiagnostics().size());
+    context.assertTrue(books.getResultInfo().getDiagnostics().get(0).getMessage().contains("Cannot deserialize instance of"));
+    context.assertEquals(2, books.getResultInfo().getTotalRecords());
+    context.assertEquals(0, books.getBooks().size());
+
+    buf = checkURLs(context, "http://localhost:" + port + "/rmbtests/test?query=slim%3Dtrue", 200);
```
julianladisch: Can you add a comment what feature this SlimBook test checks?

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -170,7 +155,87 @@ private void postBook(TestContext context, String parameterString, int expectedS
...
+    JsonObject jo = new JsonObject(buf);
+    log.info("jo={}", jo.encodePrettily());
```
julianladisch: Can you remove this logging or change it to debug level?

adamdickmeiss: sure

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
adamdickmeiss: I'll check that we can map to  SlimBook.class.

julianladisch: What makes SlimBook different from Books that it may cause a test failure that the Books tests above don't catch?
Why is this SlimBook test needed?

adamdickmeiss:
Thx for the rename.

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
adamdickmeiss: Because it's a test that we can get a response with only a subset of the Books.

adamdickmeiss: It means we can also , easily, implement MODINVSTOR-439

julianladisch: Thanks, now I understand.
Can you extend the comment accordingly?

    // see that we handle a subset of the Book properties: id and status
    // use case: replace "SELECT jsonb FROM ..." by "SELECT jsonb_build_object('id', id, 'status', jsonb->'status') FROM ..."

adamdickmeiss:
Thx

PR #618, opened 2020-03-03
RMB-575: add "sqlExpressionQuery" parameter for schema.json

julianladisch:
* Add sqlExpressionQuery parameter for schema.json
* Add tsquery_and, tsquery_or and tsquery_phrase to general_functions.ftl
* Use "WITH SCHEMA public" for PostgreSQL "CREATE EXTENSION"
* Upgrade junit version
* Upgrade hamcrest version


domain-models-runtime/src/main/resources/templates/db_scripts/general_functions.ftl
```diff
@@ -67,18 +67,20 @@ $$ LANGUAGE sql IMMUTABLE PARALLEL SAFE STRICT;
...
+-- normalize_digits(' 01 2- 3 -- 45* -a 7 -8 9') = '012345* a 7 -8 9'
 -- normalize_digits('978 92 8011 565 9(Vol. 1011-1021)') = '9789280115659 (Vol. 1011-1021)'
 CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.normalize_digits(text) RETURNS text AS $$
-  SELECT    translate((regexp_match($1, '^([0-9 \t-]*)(.*)'))[1], E' \t-', '')
-         || CASE WHEN (regexp_match($1, '^([0-9 \t-]*)(.*)'))[1] = '' THEN ''
-                 WHEN (regexp_match($1, '^([0-9 \t-]*)(.*)'))[2] = '' THEN ''
+  SELECT    translate((regexp_match($1, '^([0-9 \t-]*(?:\*[ \t]*)?)(.*)'))[1], E' \t-', '')
+         || CASE WHEN (regexp_match($1, '^([0-9 \t-]*(?:\*[ \t]*)?)(.*)'))[1] = '' THEN ''
+                 WHEN (regexp_match($1, '^([0-9 \t-]*(?:\*[ \t]*)?)(.*)'))[2] = '' THEN ''
                  ELSE ' '
```
adamdickmeiss: You are redefining a function. If not replaced properly, the old definition sticks. Therefore, rename it (and the old definition will stay for older systems (eg 29.3.0 RMB), but will never be used).

julianladisch: Can you explain in which cases "CREATE OR REPLACE FUNCTION" may not replace the old definition?

adamdickmeiss: I can't. Would error out if parameters or other would change. Will indexes based on updating the function just be updated? Or is that a mood question because AFAIK we are always regenerating indices anyway?

julianladisch: The change allows a star at the end of the barcode. This star is punctuation that the full text index removes and ignores. Therefore no index regenerating is needed for this change.
This star affects the query string only where it signals right truncation.

adamdickmeiss: ok

PR #622, opened 2020-03-13
RMB-552: Smart index recreation on module upgrade

julianladisch:


skoczko:
Yay -- @julianladisch does it complete RMB-552?

PR #628, opened 2020-04-01
RMB-594: Sorted order of GenerateRunner class generation

julianladisch:


julianladisch:
Failing a clash is out of scope for this issue and is handled in https://issues.folio.org/browse/RMB-596 "GenerateRunner should fail on duplicate class name".
I want deterministic builds even if clashes get detected.
The fix for this issue is trivial (two lines of code to add sorting) and is independent of clash detection. There is no need to block this sorting issue until clash detection and reporting is completed.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -166,6 +166,7 @@ public void generate(String inputDirectory) throws Exception {
     int numMatches = 0;
 
     File []ramls = new File(inputDirectory).listFiles( (dir, name) -> name.endsWith(".raml") );
+    Arrays.sort(ramls);
```
adamdickmeiss, line 169: Is sorting of raml files sufficient? I mean if I declared two public functions in two files a.java and b.java with this it will magically work if a sorts before b. It might be determistic, but in no way logical. Eg if I move function a.java to q.java it would work again...

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -223,6 +224,7 @@ public static void createLookupList(File directory, String name, List<String> su
       paths = pathStream.map(basePath::relativize)
           .filter(path -> pathMatchers.stream()
                             .anyMatch(pathMatcher -> pathMatcher.matches(path)))
+          .sorted()
```
adamdickmeiss, line 227: WHy does sorting the paths work? Is that necessary? I don't understand why sorting paths would have any effect on those types.

julianladisch: The created file currently has random order and with this fix it will always have exactly the same order each time. We want deterministic behaviour.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/GenerateRunner.java
julianladisch: Yes, this does not solve the cause of the problem.
However, it either always fails the build or it always succeeds.
If it fails it will also fail on the developer local machines and enables them to debug and fix it.

PR #637, opened 2020-04-24
RMB-591 Return actual counts for the last page results

MikhailFokanov:
If it is the last page and result count is less than limit, we know for sure the accurate result count.


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1879,6 +1874,10 @@ QueryHelper buildQueryHelper(
     if (offset != -1) {
       queryHelper.offset = offset;
     }
+    int limit = wrapper.getLimit().get();
+    if (limit != -1) {
```
julianladisch, line 1878: -1 means no limit, therefore limit should be set to MAX_INT.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -43,37 +63,11 @@
...
-import io.vertx.ext.asyncsql.AsyncSQLClient;
-import io.vertx.ext.sql.ResultSet;
-import io.vertx.ext.sql.SQLConnection;
-import io.vertx.ext.sql.SQLRowStream;
-import io.vertx.ext.sql.UpdateResult;
-import java.util.Optional;
-import java.util.UUID;
-
-import org.folio.rest.tools.utils.NetworkUtils;
-
```
julianladisch, line 76: Please disable automatic import line reordering in your IDE.
If you want to reorder the import lines create a separate pull request to keep this pull request focused on the count issue.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1928,7 +1927,13 @@ QueryHelper buildQueryHelper(
           replyHandler.handle(Future.failedFuture(query.cause()));
           return;
         }
-        replyHandler.handle(Future.succeededFuture(resultSetMapper.apply(new TotaledResults(query.result(), total))));
+        ResultSet result = query.result();
+        Integer actualTotal = total;
+        if (queryHelper.limit > 0 && result.getNumRows() < queryHelper.limit) {
+          actualTotal = queryHelper.offset +result.getNumRows();
+        }
```
julianladisch, line 1934: This algorithm is a copy of the algorithm in PgUtil.
Please extract it into a method for reuse and write unit tests for this extracted method to show that it is correct.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1097,26 +1097,51 @@ public static void setOptimizedSqlSize(int size) {
...
-    }
+    totalRecords = getTotalRecords(resultSize, totalRecords, offset, limit);
 
     return collection(collectionClazz, recordList, totalRecords);
   }
 
+  /**
+   * Function to correct estimated result count:
+   * If the resultsCount is equal to 0, the result should be not more than offset
+   * If the resultsCount is equal to limit, the result should be not less than offset
```
julianladisch, line 1122:
```suggestion
   * If the resultsCount is equal to limit, the result should be not less than offset + limit
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1097,26 +1097,51 @@ public static void setOptimizedSqlSize(int size) {
...
+      return estimateCount;
+    }
+    if (resultsCount == 0) {
+      estimateCount = Math.min(offset, estimateCount);
+    } else if (resultsCount == limit) {
+      estimateCount = Math.max(offset + limit, estimateCount);
+    } else {
+      estimateCount = offset + resultsCount;
+    }
+    return estimateCount;
```
julianladisch, line 1142: There is no need to change `estimateCount`, you can directly return the result, for example

    return Math.min(offset, estimateCount);

This also avoids parameter assignment: https://checkstyle.org/config_coding.html#ParameterAssignment

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -147,10 +155,11 @@ private static String randomUuid() {
...
 
   /**
-   * Return a handler that, when invoked, asserts that the AsyncResult succeeded, the Response has httpStatus,
-   * and the toString() of the entity of Response contains the snippet.
+   * Return a handler that, when invoked, asserts that the AsyncResult succeeded, the Response has
+   * httpStatus, and the toString() of the entity of Response contains the snippet.
    */
-  private Handler<AsyncResult<Response>> asyncAssertSuccess(TestContext testContext, int httpStatus, String snippet) {
+  private Handler<AsyncResult<Response>> asyncAssertSuccess(TestContext testContext, int httpStatus,
+    String snippet) {
```
julianladisch, line 162: After a line break within a method's parameter four additional spaces should be added to the indentation.
Using two is misleading because it looks as if it were starting the actual code block.
Can you adjust your IDE's settings and fix all those wrapped lines?

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1022,438 +1074,562 @@ private void setUpUserDBForTest(TestContext testContext, PostgresClient pg) {
...
     }
+
     public static Response respond201WithApplicationJson(Object entity, HeadersFor201 headers) {
       return ResponseImpl.respond201WithApplicationJson(entity, headers);
     }
+
     public static Response respond204() {
       return ResponseImpl.respond204();
     }
+
```
julianladisch, line 1632: 860 lines were changed in this file.
Most of them are not related to the counts issue.
Please create a separate pull request for the whitespace and line wrapping changes.
Without the whitespace and line wrapping changes this pull request can more easily been code reviewed and been backported.

julianladisch: There are still 515 lines with whitespace/line wrapping changes unrelated to this counts issue in PgUtilIT.

julianladisch: Still there.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1006,6 +1008,23 @@ public void optimizedSqlCanSetSize() {
...
+    assertEquals(PgUtil.getTotalRecords(0, 20, 10, 10),  (Integer)10);
+
+    assertEquals(PgUtil.getTotalRecords(10, 30, 10, 10),  (Integer)30);
+
+    assertEquals(PgUtil.getTotalRecords(10, 20, 20, 10),  (Integer)30);
+
+    assertEquals(PgUtil.getTotalRecords(5, 20, 20, 10), (Integer) 25);
+
+    assertEquals(PgUtil.getTotalRecords(5, 20, 20, 10), (Integer) 25);
```
julianladisch, line 1025: Either use

    assertEquals(expected, actual);

or

    assertThat(actual, is(expected));

Note the different order of expected and actual. Otherwise we get misleading error messages when an assertion fails.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2447,6 +2453,11 @@ public ResultsHelper(SQLRowStream sqlRowStream, Class<T> clazz) {
     return results;
   }
 
+  private <T> int getResultListRowCounts(List<T> list) {
+    return (int) list.stream().filter(e -> !(e instanceof Facet)
+      && !((e instanceof Map) && ((Map) e).containsKey("count"))).count();
+  }
+
```
julianladisch, line 2460:
```suggestion
  /**
   * @return number of list entries excluding the Facet count and total count entries
   */
  @SuppressWarnings("rawtypes")
  private <T> int getResultListRowCounts(List<T> list) {
    return (int) list.stream()
        .filter(e -> !(e instanceof Facet) &&
                     !((e instanceof Map) &&
                         ((Map) e).size() == 1) && ((Map) e).containsKey(COUNT_FIELD))
        .count();
  }

```
* Add javadoc
* Suppress rawtypes warning
* Replace String constant
* Line wrapping that supports understanding
* Ensure map.size() == 1 because there may be data records that contain a count element.
* Can you add `"count":9` to each mockdata/poLines.json line to catch this?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -45,6 +45,10 @@
   /** empty = no environment variables */
   private JsonObject empty = new JsonObject();
 
+  private static int DEFAULT_OFFSET = 0;
```
julianladisch, line 48:
```suggestion
  private static final int DEFAULT_OFFSET = 0;
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -45,6 +45,10 @@
...
+
+  private static int DEFAULT_LIMIT = 10;
```
julianladisch, line 50:
```suggestion
  private static final int DEFAULT_LIMIT = 10;
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1006,6 +1008,23 @@ public void optimizedSqlCanSetSize() {
     assertThat(PgUtil.getOptimizedSqlSize(), is(oldSize));
   }
 
+  @Test
+  public void getTotalRecordsTest() {
+    assertNull(PgUtil.getTotalRecords(10, null, 0, 0));
+
+    assertEquals(PgUtil.getTotalRecords(10, 20, 0, 0),  (Integer)20);
+
+    assertEquals(PgUtil.getTotalRecords(0, 20, 10, 10),  (Integer)10);
```
julianladisch, line 1017: limit and offset should be different to be sure that the result the method takes the offset and not the limit.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
julianladisch, line 1025: That last two test cases are the same.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -85,11 +85,11 @@
     System.setProperty(LoggerFactory.LOGGER_DELEGATE_FACTORY_CLASS_NAME, "io.vertx.core.logging.Log4j2LogDelegateFactory");
   }
 
-  @BeforeClass
+  /*@BeforeClass
   public static void doesNotCompleteOnWindows() {
     final String os = System.getProperty("os.name").toLowerCase();
     org.junit.Assume.assumeFalse(os.contains("win")); // RMB-261
-  }
+  }*/
```
julianladisch, line 92: If it runs on Windows please remove this code (it is still in git history).

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
julianladisch: Thanks, fixed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -2,45 +2,48 @@
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Context;
 import io.vertx.core.Future;
```
julianladisch, line 5:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
MikhailFokanov: I managed to tun it with postgres-conf, but it still causes timeout with embedded postgres. So I uncommented the code.

julianladisch: Thanks.

julianladisch:
I merge this to master to get this tested with mod-inventory-storage on perf as planned.
The code change not to use count_estimate_default was approved and committed as pr #634.
If we agree that we should go back to always use count_estimate_default we need a new pr.

PR #639, opened 2020-04-28
Bump maven-checkstyle-plugin version to 3.1.1

julianladisch:


craigmcnally:
What is the reason for the change?  Just to stay up-to-date?  Is there new functionality we're looking to use?

craigmcnally:
I'm guessing there are breaking changes given the major version bump 2.X.Y -> 3.X.Y...  I can't find any release notes, but did see this (which doesn't appear to affect us):

> Please note that the following parameters have been completely removed from the plugin configuration:
> 
> sourceDirectory: use sourceDirectories instead;
> testSourceDirectory: use testSourceDirectories instead.

julianladisch:
Yes, this pull request is to stay up to date and have all bugs fixed that were released: https://checkstyle.sourceforge.io/releasenotes.html
Checkstyle is used in IDEs and at build time only and is not delivered with RMB therefore it cannot affect any RMB based modules.

PR #640, opened 2020-04-28
Update to vertx-pg-client RMB-246

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/PgExceptionFacade.java
```diff
@@ -18,13 +18,15 @@
...
       fields = Collections.emptyMap();
       return;
     }
 
-    ErrorMessage errorMessage = ((GenericDatabaseException) throwable).getErrorMessage();
-    fields = errorMessage.getFields();
+    fields = new HashMap<>();
+    fields.put('M', ((PgException) throwable).getMessage());
+    fields.put('D', ((PgException) throwable).getDetail());
+    fields.put('C', ((PgException) throwable).getCode());
```
julianladisch, line 29: getTable() and getIndex() can no longer return anything and should be removed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgExceptionFacade.java
```diff
@@ -18,13 +18,15 @@
    * @param throwable a GenericDatabaseException; any other Throwable is handled gracefully
    */
   public PgExceptionFacade(Throwable throwable) {
-    if (!(throwable instanceof GenericDatabaseException)) {
+    if (!(throwable instanceof PgException)) {
       fields = Collections.emptyMap();
       return;
     }
```
julianladisch, line 24: Avoid duplicate code:

    fields = PgExceptionUtil.getBadRequestFields(throwable);
    if (fields == null) {
      fields = Collections.emptyMap();
    }

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1097,20 +1098,20 @@ public static void setOptimizedSqlSize(int size) {
...
-    List<JsonObject> jsonList = resultSet.getRows();
-    List<T> recordList = new ArrayList<>(jsonList.size());
     int totalRecords = 0;
-    for (JsonObject object : jsonList) {
-      String jsonb = object.getString(JSON_COLUMN);
+    List<T> recordList = new ArrayList<>(resultSet.size());
+    RowIterator<Row> iterator = resultSet.iterator();
+    while (iterator.hasNext()) {
+      Row row = iterator.next();
+      String jsonb = row.getValue(JSON_COLUMN).toString();
```
julianladisch, line 1109: getValue returns a JsonObject: https://github.com/eclipse-vertx/vertx-sql-client/tree/master/vertx-pg-client#supported-data-types
We don't want vertx-pg-client to create a JsonObject.
We should request a String: `SELECT jsonb::text`

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClientStreamResult.java
```diff
@@ -99,9 +100,10 @@ void fireEndHandler() {
    * @param cause
    */
   void fireExceptionHandler(Throwable cause) {
-    if (exceptionHandler != null) {
+    if (!failed && exceptionHandler != null) {
       exceptionHandler.handle(cause);
     }
+    failed = true;
```
julianladisch, line 106: `failed = true` should be set before calling exceptionHandler just in case that exceptionHandler tries to call fireHandler or again fireExceptionHandler.

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -0,0 +1,16 @@
+package org.folio.rest.persist;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
```
julianladisch, line 4:
```suggestion
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1746,43 +1730,62 @@ public TotaledResults(ResultSet set, Integer total) {
...
 
-<T> void doStreamRowResults(SQLRowStream sqlRowStream, Class<T> clazz,
-    List<FacetField> facets, ResultInfo resultInfo,
-    PostgresClientStreamResult<T> streamResult,
-    Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
+  private List<String> populateColumnNames(Row row)
+  {
+    int i = 0;
+    List<String> columnNames = new ArrayList<>();
+    while (row.getColumnName(i) != null) {
```
julianladisch, line 1755: This should be a for loop.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1746,43 +1730,62 @@ public TotaledResults(ResultSet set, Integer total) {
...
-      doStreamRowResults(stream.result(), clazz, facets, resultInfo, streamResult, replyHandler);
+      doStreamRowResults(stream, clazz, facets, tx, streamResult, replyHandler);
     });
   }
 
-<T> void doStreamRowResults(SQLRowStream sqlRowStream, Class<T> clazz,
-    List<FacetField> facets, ResultInfo resultInfo,
-    PostgresClientStreamResult<T> streamResult,
-    Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
+  private List<String> populateColumnNames(Row row)
```
julianladisch, line 1751: The method should be static.
Can you rename the method to `columnNames` or `getColumnNames`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgExceptionFacade.java
adamdickmeiss: Oh yes.. Saw that the "hard way" in mod-inventory-storage, which in one place checked for 't'

domain-models-runtime/src/main/java/org/folio/rest/persist/PgExceptionFacade.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClientStreamResult.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: I can't that to work.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/OutStream.java
```diff
@@ -13,6 +17,17 @@
...
   private Object data;
 
+  public OutStream() {
+  }
+
+  public OutStream(RowSet<Row> result) {
+    RowIterator<Row> it = result.iterator();
+    StringBuilder s = new StringBuilder();
+    while (it.hasNext()) {
+      s.append(it.next().toString());
```
julianladisch, line 27: This does not export the column names and therefore fails the new tests in AdminAPI.
Expected valid JSON: `[{"relname":"users","buffered":"2320 kB","buffers_percent":"0.4","percent_of_relation":"100.3"}, {"relname":"pg_depend_reference_index","buffered":"600 kB","buffers_percent":"0.1","percent_of_relation":"31.4"}]`
Actual invalid JSON: `[users, 2320 kB, 0.4, 100.3][pg_depend_reference_index, 600 kB, 0.1, 31.4]`

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -130,6 +126,35 @@ public void configFileTenant() throws Exception {
...
 
+  @Test
+  public void testPgConnectOptionsEmpty() {
+    JsonObject conf = new JsonObject();
+    PgConnectOptions options = PostgresClient.createPgConnectOptions(conf);
+    Assert.assertEquals("localhost", options.getHost());
+    Assert.assertEquals(5432, options.getPort());
+    Assert.assertEquals("user", options.getUser());
+    Assert.assertEquals("pass", options.getPassword());
+    Assert.assertEquals("db", options.getDatabase());
```
julianladisch, line 137: The same assert method should be used throughout a file.
Can you replace `Assert.assertEquals(expected, actual)` by `assertThat(actual, is(expected))` in testPgConnectOptionsEmpty and testPgConnectOptionsFull?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -130,6 +126,35 @@ public void configFileTenant() throws Exception {
...
+
+    PgConnectOptions options = PostgresClient.createPgConnectOptions(conf);
+    Assert.assertEquals("myhost", options.getHost());
+    Assert.assertEquals(5433, options.getPort());
+    Assert.assertEquals("myuser", options.getUser());
+    Assert.assertEquals("mypassword", options.getPassword());
+    Assert.assertEquals("mydatabase", options.getDatabase());
+  }
+
+  /* DISABLED
```
julianladisch, line 157: Use `@Ignore("some reason")` https://github.com/junit-team/junit4/wiki/Ignoring-tests
to ignore a test and to give an explanation when this test can be re-enabled.
This also allows IDEs to find references/usages.
JUnit will report the number of skipped tests.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
adamdickmeiss: Sure

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
adamdickmeiss: OK (or I'll just fix it).

domain-models-runtime/src/test/java/org/folio/rest/impl/AdminAPIIT.java
```diff
@@ -13,6 +13,7 @@
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
+import io.vertx.core.json.pointer.JsonPointerIterator;
```
julianladisch, line 16:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1003,9 +960,70 @@ public void saveBatchXTrans(TestContext context) {
...
+
+    postgresClient.rollbackTx(trans, context.asyncAssertFailure(res ->
+        context.assertEquals("failure", res.getMessage())));
+  }
+
+
+  @Test
+  public void endTxNullConnection(TestContext context) {
+    postgresClient = createFoo(context);
+    Promise<SQLConnection> promise = Promise.promise();
```
julianladisch, line 989:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1003,9 +960,70 @@ public void saveBatchXTrans(TestContext context) {
...
+      SQLConnection conn = new SQLConnection(null, trans1.tx);
+      trans2.complete(conn);
+      postgresClient.endTx(trans2.future(), context.asyncAssertSuccess());
+    }));
+  }
+
+  @Test
+  public void endTxNullTransaction(TestContext context) {
+    postgresClient = createFoo(context);
+    Promise<SQLConnection> promise = Promise.promise();
```
julianladisch, line 1002:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1003,9 +960,70 @@ public void saveBatchXTrans(TestContext context) {
...
+      SQLConnection conn = new SQLConnection(trans1.conn, null);
+      trans2.complete(conn);
+      postgresClient.endTx(trans2.future(), context.asyncAssertFailure());
+    }));
+  }
+
+  @Test
+  public void rollbackTxNullTransaction(TestContext context) {
+    postgresClient = createFoo(context);
+    Promise<SQLConnection> promise = Promise.promise();
```
julianladisch, line 1015:
```suggestion
```

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/OutStream.java
adamdickmeiss: Yes. Saw that and should be fixed by now

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1857,13 +2240,12 @@ public void executeTrans(TestContext context) {
 
   @Test
   public void executeTransSyntaxError(TestContext context) {
-    Async async = context.async();
     postgresClient = postgresClient();
     postgresClient.startTx(trans -> {
```
julianladisch, line 2244: Only if ALL async callbacks have a context.asyncAssertXXX we don't need a context.async().

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
adamdickmeiss: It does not even compile, uses classes from old driver, eg PostgresSQLConnectionIimpl .

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2084,7 +2474,7 @@ public void selectStreamTrans(TestContext context) {
       postgresClient.selectStream(trans, "SELECT i FROM numbers WHERE i IN (21, 23, 25) ORDER BY i",
           context.asyncAssertSuccess(select -> {
             intsAsString(select, context.asyncAssertSuccess(string -> {
-              postgresClient.endTx(trans, context.asyncAssertSuccess());
+              // postgresClient.endTx(trans, context.asyncAssertSuccess());
```
julianladisch, line 2471: Please uncomment.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2484,6 +2897,7 @@ public void streamGetConnectionFailed(TestContext context) throws IOException, F
         x -> context.assertEquals("connection error", x.getMessage())));
   }
 
+  /* DISABLED
```
julianladisch, line 2894: Please uncomment.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: ? don't see it in that function, but a lot of functions could be change to use Async less often.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: yep

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -353,6 +388,7 @@ public void testProcessQueryException() {
     );
   }
 
+  /* DISABLED
```
julianladisch, line 391: Can you uncomment this test?

doc/upgrading.md
```diff
@@ -13,6 +14,30 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+    `iterator`.
+  * `PostgresClient.getClient()` is no longer public. If you need a
+    connection, use PostgresClient.startTx(). If you don't have
+    transactions, you don't need access to SQLConnection.
+  * Exceptions thrown by new client is `io.vertx.pgclient.PgException`.
+    Was com.github.jasync.sql.db.postgresql.exceptions.GenericDatabaseException
+    before. The getMessage() only contains message! Not details, code.
+  * `PgExceptionFacade.getTable` removed.
+  * `PgExceptionFacade.getIndex` removed.
+  * `PostgresClient.mutate` removed (deprecated since Oct 2018).
```
julianladisch, line 39: There are several PostgresClient.get methods that have been deprecated since 6 months. Should we remove them as well?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
adamdickmeiss: Yes. Will do. Can not just be commented out, because uses Impl classes from old driver, so I'l have to find new ones or implement interfaces etc.

doc/upgrading.md
adamdickmeiss: I think it depends on how much they are used.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -461,7 +492,10 @@ private void init() throws Exception {
       startEmbeddedPostgres();
     }
 
-    client = io.vertx.ext.asyncsql.PostgreSQLClient.createNonShared(vertx, postgreSQLClientConfig);
+    PgConnectOptions connectOptions = createPgConnectOptions(postgreSQLClientConfig);
+
+    PoolOptions poolOptions = new PoolOptions().setMaxSize(5);
```
wwelling, line 497: Do we want to hard code the max size for the connection pool?

adamdickmeiss: We will probably have an env setting for this.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -580,79 +594,73 @@ public static String pojo2json(Object entity) throws Exception {
...
+        done.handle(Future.failedFuture(e.getCause()));
       }
     });
   }
 
+  static void finalizeTx(AsyncResult<Void> txResult, SqlConnection conn, Handler<AsyncResult<Void>> done ) {
+    if (conn != null) {
+      conn.close();
+    }
+    if (txResult.failed() && !"Transaction already completed".equals(txResult.cause().getMessage())) {
```
wwelling, line 620: It would be nice not to have to compare to the transaction exception message. But I do not see a way to get the integer status.

adamdickmeiss: Yeah.. Not pretty.. Reverse mapping those message are really problematic.

adamdickmeiss:
The connections that is created from PgPool are not released. Since we have a PgPool per tenant, this does not scale.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
julianladisch: I went through all usages of deprecated methods reported by sonar and could not find any deprecated PostgresClient.get usage:
https://sonarcloud.io/organizations/folio-org/rules?open=java%3AS1874&q=depreca#coding-rules-detail-quality-profiles

doc/upgrading.md
julianladisch: I went through all usages of deprecated methods reported by sonar and could not find any deprecated PostgresClient.get usage:
https://sonarcloud.io/organizations/folio-org/rules?open=java%3AS1874&q=depreca#coding-rules-detail-quality-profiles

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
julianladisch: Split off into a separate issue: https://issues.folio.org/browse/RMB-614

doc/upgrading.md
```diff
@@ -13,6 +14,31 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+    old SQL client. `io.vertx.ext.sql.SQLConnection` ->
+    `org.folio.rest.persist.SQLConnection`.
+  * All functions that previusly returned `UpdateResult` now returns
+    `RowSet<Row>`. From that result the number of rows affacted by
+    SQL was `getUpdated()` it is now `rowCount()`.
+  * All functions that previusly returned `ResultSet` now returns
+    `RowSet<Row>`. From that result, the number of rows affacted by
+    SQL is `rowCount()`. The size() method returns number of rows
+    returned. An iterator to go through rows is obtained by calling
+    `iterator`.
```
julianladisch, line 30: vertx-pg-client returns JSONB as a JsonObject, the old client returned it as a String.

adamdickmeiss: Added that, thx

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2801,31 +2802,92 @@ public void selectSingle(AsyncResult<SQLConnection> conn, String sql, Handler<As
...
+    for (int i = 0; i < sql.length(); i++) {
+      if (sql.charAt(i) =='?') {
+        preparedSql.append("$" + pos);
+        pos++;
+      } else {
+        preparedSql.append(sql.charAt(i));
+      }
+    }
+    return preparedSql.toString();
+  }
```
julianladisch, line 2823: Unmasking `??`, including the params.isEmpty() case, is missing. Better:

```
  /**
   * Replace each "??" by "?" and each other "?" by "$1", "$2", ...
   *
   * The old vertx-mysql-postgresql-client-jasync driver uses double question mark
   * to mask any non-placeholder question mark:
   * <a href="https://github.com/jasync-sql/jasync-sql/blob/1.0.17/postgresql-async/src/main/java/com/github/jasync/sql/db/postgresql/PreparedStatementHolder.kt#L23">
   * jasync-sql PreparedStatementHolder</a>
   */
  private static String legacySql(String sql) {
    StringBuilder preparedSql = new StringBuilder();
    int pos = 1;
    boolean masked = false;
    for (int i = 0; i < sql.length(); i++) {
      if (sql.charAt(i) != '?' || masked) {
        preparedSql.append(sql.charAt(i));
        masked = false;
      } else if (i + 1 < sql.length() && sql.charAt(i + 1) == '?') {
        masked = true;
      } else {
        preparedSql.append('$').append(pos);
        pos++;
      }
    }
    return preparedSql.toString();
  }
```

julianladisch:
```
  @Test
  public void selectParamQuestionMarkJsonb(TestContext context) {
    postgresClient()
    // "jsonb ? field" checks whether field is a key in jsonb.
    // double the question mark to mask from parameter substitution
    .select("SELECT '{\"field\": \"value\"}'::jsonb ?? ?",
        new JsonArray().add("field"), context.asyncAssertSuccess(select -> {
          context.assertEquals(true, select.iterator().next().getBoolean(0));
        }));
  }
```
```
  @Test
  public void selectParamQuestionMarkRegexp(TestContext context) {
    postgresClient()
    // ? in the regexp makes the space optional.
    // double the question mark to mask from parameter substitution
    .select("SELECT 'abc 123' ~ '^[a-z]+ ??[0-9]+$' AND 'xyz' = ?",
        new JsonArray().add("xyz"), context.asyncAssertSuccess(select -> {
          context.assertEquals(true, select.iterator().next().getBoolean(0));
        }));
  }
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2801,31 +2802,92 @@ public void selectSingle(AsyncResult<SQLConnection> conn, String sql, Handler<As
...
+
+  private static Tuple legacyArguments(JsonArray params) {
+    Tuple tuple = Tuple.tuple();
+    for (int i = 0; i < params.size(); i++) {
+      tuple.addUUID(UUID.fromString(params.getString(i)));
```
julianladisch, line 2828: This assumes that all parameters are UUIDs. This is the case for the current unit tests, but we cannot assume that.
Example:
```
DELETE FROM tenant_raml_module_builder.foo WHERE jsonb->>'key' = ?
```

julianladisch: I don't see any possibility to support the legacy `?` syntax.
I suggest to make a breaking change and switch to `$1` syntax where people must pass in Tuple and where the Tuple elements must be a UUID, not a String, where needed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2801,31 +2802,92 @@ public void selectSingle(AsyncResult<SQLConnection> conn, String sql, Handler<As
    * @param replyHandler  The query result or the failure.
    */
   public void selectSingle(String sql, JsonArray params, Handler<AsyncResult<JsonArray>> replyHandler) {
```
julianladisch, line 2804: This is inconsistent.
`select(String sql, JsonArray params, Handler<AsyncResult<RowSet<Row>>> replyHandler)`
returns RowSet<Row> but selectSingle returns JsonArray, I expect Row.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -461,7 +507,19 @@ private void init() throws Exception {
...
+    log.debug("Setup timer = " + activityTimerId + " for tenant " + tenantId);
+    activityTimerId = vertx.setPeriodic(postgreSQLClientConfig.getInteger(CONNECTION_RELEASE_DELAY), res -> {
+      log.debug("expire timer for tenant " + tenantId);
+      if (!pgPoolActivity && client != null) {
+        log.debug("Expire pool for tenant " + tenantId);
+        client.close();
+        client = null;
+      }
+      pgPoolActivity = false;
+    });
```
julianladisch, line 522: The expire timer should be split into a separate pull request so that we can discuss it separately.

julianladisch: A module can create any number of PostgresClients, if a PostgresClient is no longer used does the timer stop and allow the garbage collection to free the PostgresClient?
What happens if a bulk insert or a transaction takes three minutes? Does the expire timer cancel it?

adamdickmeiss: I don't think it should be a separate PR. We can't have a situation where connections are never released.

For same tenant only one PostgresClient is returned (it's a static constructor).

Connections that are returned by pool are not released. Only the ones that are unused.· So w could release every second if we wanted to.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Yep. Can't assume it. It happened to work fine for mod-inventory-storage and unit tests. I'll change it. This will make it more work to upgrade, obviously.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: OK. But we'll just not do this and require people to use the new syntax, I suppose.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: ok. Makes sense. Will change it to return Row

adamdickmeiss: If the API returns not even a single row.. What should we do?
1. throw exception
2. success with fake a Row with no columns (that's what it does at the moment).
3. success with null Row

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -461,7 +507,19 @@ private void init() throws Exception {
...
     }
 
-    client = io.vertx.ext.asyncsql.PostgreSQLClient.createNonShared(vertx, postgreSQLClientConfig);
+    client = null;
+    startPool();
+
+    log.debug("Setup timer = " + activityTimerId + " for tenant " + tenantId);
+    activityTimerId = vertx.setPeriodic(postgreSQLClientConfig.getInteger(CONNECTION_RELEASE_DELAY), res -> {
+      log.debug("expire timer for tenant " + tenantId);
+      if (!pgPoolActivity && client != null) {
```
wwelling, line 516: With `startPool` being called at an arbitrary time by `getConnection` , is it possible that the client might be closed immediately after the `client != null` check of `startPool`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
wwelling: It could return an empty Optional and have the consumer handle the not found exception.

julianladisch: The javadoc of the old selectSingle says:

    Run a select query and return the first record, or null if there is no result.

I think we can stay with the old solution. A null check is more natural than a column count check.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Yes.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Yes. Thx

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: I'm going to remove all this. Connections that are handed out by the pool are also closed, upon close of the pool.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -444,6 +451,45 @@ public static void closeAllClients() {
     CompositeFuture.join(list);
   }
 
+  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig) {
+    PgConnectOptions pgConnectOptions = new PgConnectOptions();
+    String host = sqlConfig.getString(HOST);
+    if (host != null) {
```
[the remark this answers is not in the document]
adamdickmeiss: I think != null is very convenient.. When we say that empty strings are "null", then what is an empty string.. Stripping blanks.. UNICODE blanks? I like that if a environment variable or JSON property is set in configuration, then that value is used. Whatever it is.

doc/upgrading.md
```diff
@@ -13,6 +14,34 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+    `iterator`. JSONB is returned as JsonObject, the old client returned
+    it as a String.
+  * `PostgresClient.selectSingle` returns Row rather than `JsonArray`.
+  * SQL parameters changed from `JsonArray` to `Tuple`.
+  * `PostgresClient.getClient()` is no longer public. If you need a
+    connection, use PostgresClient.startTx(). For modules that wish to use
+    vertx-pg-client directly, PostgresClient.getConnection is offered -
+    it returns SqlConnection from the pool that is managed by PostgresClient.
+  * Exceptions thrown by new client is `io.vertx.pgclient.PgException`.
+    Was com.github.jasync.sql.db.postgresql.exceptions.GenericDatabaseException
```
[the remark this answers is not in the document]
adamdickmeiss: Sure. Will fix that.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: ok. Since it does not work as I thought it does, I'll take all this out. This means we probably can't use vertx-pg-client as it stands.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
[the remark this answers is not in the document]
wwelling: It would be environment variable or JSON configuration error of course. I suppose it makes no difference of setting the host to `""` or not setting it at all. Would still get a connection error either way. Might not be worth validating and throwing exception without communication to database.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2903,50 +2904,27 @@ public void selectStream(AsyncResult<SQLConnection> conn, String sql, JsonArray
...
-   * Execute an INSERT, UPDATE or DELETE statement.
-   * @param sql - the sql to run
-   * @param replyHandler - the result handler with UpdateResult
+   * Get vertx-pg-client connection
+   * @param replyHandler
    */
-  public void execute(String sql, Handler<AsyncResult<UpdateResult>> replyHandler)  {
-    long s = System.nanoTime();
-    client.getConnection(res -> {
+  public void getConnection(Handler<AsyncResult<SqlConnection>> replyHandler) {
```
julianladisch, line 2917: Can getConnection return a PgConnection (not a SqlConnection)? https://github.com/eclipse-vertx/vertx-sql-client/blob/3.9.0/vertx-pg-client/src/main/java/io/vertx/pgclient/PgConnection.java#L34

adamdickmeiss: Done!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -989,79 +996,53 @@ public void saveBatch(String table, JsonArray entities, Handler<AsyncResult<Resu
...
           .append(schemaName).append(DOT).append(table)
-          .append(" (id, jsonb) VALUES ");
-      for (int i = 0; i < idEntityPairs.size(); i += 2) {
-        if (i > 0) {
-          sql.append(',');
-        }
-        sql.append("(?,?)");
-      }
+          .append(" (id, jsonb) VALUES ($1, $2)");
       sql.append(RETURNING_ID);
```
julianladisch, line 1029:
```suggestion
          .append(" (id, jsonb) VALUES ($1, $2)")
          .append(RETURNING_ID);
```

adamdickmeiss: Using a plain String. That should be fine for a one time assignment, IIRC.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1023,16 +1074,16 @@ public void saveBatchNullConnection(TestContext context) {
 
   @Test
   public void saveBatchNullList(TestContext context) {
-    createFoo(context).saveBatch(FOO, (List<Object>)null, context.asyncAssertSuccess(save -> {
-      context.assertEquals(0, save.getNumRows());
+    createFoo(context).saveBatch(BAR, (List<Object>)null, context.asyncAssertSuccess(save -> {
+      context.assertNull(save);
```
julianladisch, line 1078: This is an unneccessary breaking change that is inconsistent with vertx-pg-client that returns an empty RowSet when "UPDATE ... WHERE ..." does not update any record.
Can you return an empty RowSet?

```
  private RowSet<Row> emptyRowSet(String ... columnNames) {
    return new RowSet<Row>() {
      List<String> columnsNames = Arrays.asList(columnNames);

      @Override
      public int rowCount() {
        return 0;
      }

      @Override
      public List<String> columnsNames() {
        return columnsNames;
      }

      @Override
      public int size() {
        return 0;
      }

      @Override
      public <V> V property(PropertyKind<V> propertyKind) {
        return null;
      }

      @Override
      public RowSet<Row> value() {
        return this;
      }

      @Override
      public RowIterator<Row> iterator() {
        return new RowIterator<Row>() {
          @Override
          public boolean hasNext() {
            return false;
          }

          @Override
          public Row next() {
            return null;
          }
        };
      }

      @Override
      public RowSet<Row> next() {
        return null;
      }
    };
  }
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1023,16 +1074,16 @@ public void saveBatchNullConnection(TestContext context) {
...
     }));
   }
 
   @Test
   public void saveBatchEmptyList(TestContext context) {
     List<Object> list = Collections.emptyList();
     createFoo(context).saveBatch(FOO, list, context.asyncAssertSuccess(save -> {
-      context.assertEquals(0, save.getNumRows());
+      context.assertNull(save);
```
julianladisch, line 1086: This is an unneccessary breaking change that is inconsistent with vertx-pg-client that returns an empty RowSet when "UPDATE ... WHERE ..." does not update any record.
Can you return an empty RowSet?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2582,10 +2584,11 @@ public ResultsHelper(SQLRowStream sqlRowStream, Class<T> clazz) {
    * @param columnNames
    * @param clazz
    * @param isAuditFlavored
+   * @param externalColumnSetters
    * @return
```
julianladisch, line 2588:
```suggestion
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2582,10 +2584,11 @@ public ResultsHelper(SQLRowStream sqlRowStream, Class<T> clazz) {
...
    */
-  <T> Map<String, Method> getExternalColumnSetters(List<String> columnNames, Class<T> clazz, boolean isAuditFlavored) {
-    Map<String, Method> externalColumnSetters = new HashMap<>();
+  <T> void getExternalColumnSetters(List<String> columnNames, Class<T> clazz, boolean isAuditFlavored,
```
julianladisch, line 2590: Method name getExternalColumnSetters is misleading because it doesn't return anything.
Can you replace it by a better method name, for example collectExternalColumnSetters or setupExternalColumnSetters?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2611,20 +2613,17 @@ public ResultsHelper(SQLRowStream sqlRowStream, Class<T> clazz) {
    * @param o
    * @param row
    */
-  void populateExternalColumns(Map<String, Method> externalColumnSetters, Object o, JsonObject row) {
+  void populateExternalColumns(Map<String, Method> externalColumnSetters, Object o, Row row)
+      throws InvocationTargetException, IllegalAccessException {
     for (Map.Entry<String, Method> entry : externalColumnSetters.entrySet()) {
       String columnName = entry.getKey();
       Method method = entry.getValue();
```
julianladisch, line 2620:
```suggestion
    externalColumnSetters.forEach((columnName, method) -> {
```
and properly indent the following lines.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2611,20 +2613,17 @@ public ResultsHelper(SQLRowStream sqlRowStream, Class<T> clazz) {
...
-      try {
-        Object value = row.getValue(columnName);
-        if (value instanceof JsonArray) {
-           method.invoke(o, Json.decodeValue(((JsonArray) value).encode(), method.getParameterTypes()[0]));
-        } else {
-          method.invoke(o, value);
+        String[] stringArray = row.getStringArray(columnName);
+        if (stringArray != null) {
+          method.invoke(o, Arrays.asList(stringArray));
+          continue;
```
julianladisch, line 2624: Using if-else is more appropriate and intuitive because the if block and the else block do the same thing, they differ only in the 2nd argument.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2886,67 +2867,65 @@ public void selectStream(String sql, JsonArray params, Handler<AsyncResult<SQLRo
...
-      if (res.failed()) {
-        replyHandler.handle(Future.failedFuture(res.cause()));
+  public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
+    getClient().getConnection(x -> {
+      if (x.failed()) {
+        replyHandler.handle(Future.failedFuture(x.cause()));
         return;
       }
-      replyHandler.handle(Future.succeededFuture(res.result().toString()));
+     replyHandler.handle(Future.succeededFuture((PgConnection) x.result()));
```
julianladisch, line 2918: It is not very likely but an RMB based module may update or patch vertx-pg-client and then it may return a connection where the casting fails. For better error reporting I suggest:
```suggestion
     try {
       replyHandler.handle(Future.succeededFuture((PgConnection) x.result()));
     } catch (Exception e) {
       replyHandler.handle(Future.failedFuture(e));
     }
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -3158,4 +3799,72 @@ public static String getMockData(String path) throws IOException {
...
+    RowSet rowSet = new FakeRowSet(1).withColumns(columns).withRows(rows);
+
+    Promise<RowSet<Row>> promise = Promise.promise();
+    promise.complete(rowSet);
+    PostgresClient.selectReturn(promise.future(), context.asyncAssertSuccess(res ->
+        context.assertEquals("value", res.getString(0))));
+  }
+
+  @Test
+  public void getTotalRecordsTest() {
```
julianladisch, line 3854: This should go into PostgresClientTest because it is a static method that does not require a database connection.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1076,15 +1137,15 @@ public void saveBatchJsonFail(TestContext context) {
 
   @Test
   public void saveBatchJsonNullArray(TestContext context) {
-    createFoo(context).saveBatch(BAR, (JsonArray)null, context.asyncAssertSuccess(save -> {
-      context.assertEquals(0, save.getNumRows());
+    createFoo(context).saveBatch(FOO, (JsonArray)null, context.asyncAssertSuccess(save -> {
+      context.assertNull(save);
```
julianladisch, line 1141: This is an unneccessary breaking change that is inconsistent with vertx-pg-client that returns an empty RowSet when "UPDATE ... WHERE ..." does not update any record.
Can you return an empty RowSet?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1076,15 +1137,15 @@ public void saveBatchJsonFail(TestContext context) {
...
     }));
   }
 
   @Test
   public void saveBatchJsonEmptyArray(TestContext context) {
     createFoo(context).saveBatch(FOO, new JsonArray(), context.asyncAssertSuccess(save -> {
-      context.assertEquals(0, save.getNumRows());
+      context.assertNull(save);
```
julianladisch, line 1148: This is an unneccessary breaking change that is inconsistent with vertx-pg-client that returns an empty RowSet when "UPDATE ... WHERE ..." does not update any record.
Can you return an empty RowSet?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1252,16 +1313,17 @@ public void saveBatchEmpty(TestContext context) {
     List<Object> list = Collections.emptyList();
     createFoo(context).saveBatch(FOO, list, res -> {
       assertSuccess(context, res);
-      context.assertEquals(0, res.result().getRows().size());
-      context.assertEquals("id", res.result().getColumnNames().get(0));
+      context.assertEquals(null, res.result());
```
julianladisch, line 1316: This is an unneccessary breaking change that is inconsistent with vertx-pg-client that returns an empty RowSet when "UPDATE ... WHERE ..." does not update any record.
Can you return an empty RowSet?

julianladisch: Where the first column name is `id`?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1751,15 +2227,16 @@ public void endTx(AsyncResult<SQLConnection> conn, Handler<AsyncResult<Void>> do
...
     Async async = context.async();
     JsonArray ids = new JsonArray().add(randomUuid()).add(randomUuid());
     insertXAndSingleQuotePojo(context, ids)
     .execute("DELETE FROM tenant_raml_module_builder.foo WHERE id='" + ids.getString(1) + "'", res -> {
       assertSuccess(context, res);
-      context.assertEquals(1, res.result().getUpdated());
+      context.assertEquals(1, res.result().rowCount());
       async.complete();
     });
+    async.await(1000);
```
julianladisch, line 2239:
```suggestion
```
This is not needed and can be too short when Jenkins has high load. We have a default at the beginning of PostgresClientIT that is sufficient: `Timeout rule = Timeout.seconds(15);`

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
julianladisch: This unit test will call postgresClient.startTx and then successfully complete. It does not wait until startTx has returned some result by calling the callback.
If you want to remove Async you need to add context.asyncAssertXXX to _every_ callback handler:
* postgresClient.startTx
* postgresClient.execute

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: Yes sir!

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1944,22 +2422,22 @@ public void executeTransParamNullConnection(TestContext context) throws Exceptio
...
+      context.assertEquals(1, result.get(0).rowCount());
+      context.assertEquals(1, result.get(1).rowCount());
       async.complete();
     });
   }
 
   /** @return List containg one empty JsonArray() */
-  private List<JsonArray> list1JsonArray() {
-    return Collections.singletonList(new JsonArray());
+  private List<Tuple> list1JsonArray() {
```
julianladisch, line 2439: Update javadoc and method name:
```
  /** @return List containg one empty Tuple */
  private List<Tuple> list1Tuple() {
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: I have a FakeRowSet in test helpers.. which is this + some extra. That of course, have to be moved.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: Yes

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Yes

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Sure

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Oh yes. This was cut'n paste from old code.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Makes sence

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: Yes. Plus fix the spelling

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: ok

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: ok

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: May throw exception, so I'm skipping that for now

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: Yes

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Fine.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Yes.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
julianladisch: Yes, good idea.

domain-models-runtime/src/main/java/org/folio/rest/persist/helpers/LocalRowSet.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public <V> V property(PropertyKind<V> propertyKind) {
+    return null;
+  }
+
+  @Override
+  public RowSet<Row> value() {
+    return null;
```
julianladisch, line 58:
```suggestion
    return this;
```
https://github.com/eclipse-vertx/vertx-sql-client/blob/3.9/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/RowSetImpl.java#L61

domain-models-runtime/src/main/java/org/folio/rest/persist/helpers/LocalRowSet.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public RowSet<Row> next() {
+    return null;
+  }
+
+  class FakeRowIterator implements RowIterator<Row> {
```
julianladisch, line 66: FakeRowIterator should be renamed to LocalRowIterator.

doc/upgrading.md
```diff
@@ -13,6 +14,37 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+* [RMB-246](https://issues.folio.org/browse/RMB-246) Switch to
+    [vertx-pg-client](https://vertx.io/docs/vertx-pg-client/java/).
+  * Class `SQLConnection` is now provided by RMB. The same class name
+    was used for the SQL client. `io.vertx.ext.sql.SQLConnection` ->
+    `org.folio.rest.persist.SQLConnection`.
+  * All functions that previusly returned `UpdateResult` now returns
+    `RowSet<Row>`. From that result the number of rows affacted by
+    SQL was `getUpdated()` it is now `rowCount()`.
+  * All functions that previusly returned `ResultSet` now returns
+    `RowSet<Row>`. From that result, the number of rows affacted by
```
julianladisch, line 28:
```suggestion
    `RowSet<Row>`. From that result, the number of rows affected by
```

doc/upgrading.md
julianladisch: Therefore I suggest to remove them.

doc/upgrading.md
```diff
@@ -13,6 +14,37 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+    SQL is `rowCount()`. The size() method returns number of rows
+    returned. An iterator to go through rows is obtained by calling
+    `iterator`. JSONB is returned as JsonObject, the old client returned
+    it as String.
```
julianladisch, line 32:
```suggestion
    `iterator`.
  * JSONB is returned as JsonObject, the old client returned it as String.
  * `id` and other UUID columns are returned as UUID and must be sent as UUID, old client used String.
```

doc/upgrading.md
adamdickmeiss: Yep

doc/upgrading.md
adamdickmeiss: Yep

PR #643, opened 2020-04-29
RMB-483: Envs.setEnv for Postgres testcontainer injection

julianladisch:
Make Envs.setEnv public. Then tests can start a Postgres testcontainer
and inject the connection settings into Envs.
PostgresClient uses Envs to read environment settings for Postgres.

Also add a shortcut method
`setEnv(String host, int port, String username, String password, String database)`
that takes the five parameters from a PostgreSQLContainer.


craigmcnally:
The testcontainer stuff is pretty cool - Out of curiosity, which module/codebase is planning on using this?  RMB?

julianladisch:
Okapi, RMB and most RMB based modules use embedded postgres for testing.
For the reasons explained in https://issues.folio.org/browse/RMB-483 and https://issues.folio.org/browse/FOLIO-1845 we want to switch to postgres testcontainers and will remove embedded postgres from RMB at some time (to be decided).
All RMB based modules are free to use whatever tool they want for running tests against postgres, but they probably will also switch to testcontainers.

PR #645, opened 2020-04-30
RMB-605: f_unaccent removes combining characters (diacritics, umlauts)

julianladisch:


domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerIT.java
```diff
@@ -241,6 +241,11 @@ private void assertSchemaSelectSingle(TestContext context, String sql, String ex
     assertSelectSingle(context, sql, expected);
   }
 
+  @Test
+  public void f_unaccent_combining_character(TestContext context) {
+    assertSchemaSelectSingle(context, "SELECT f_unaccent(E'a\\u0308 and a\\u0308')", "a and a");
```
craigmcnally, line 246: More than just `\\u0308` could be tested here, but given that they're all part of the same regex expression it's probably not a huge deal.

domain-models-runtime/src/main/resources/templates/db_scripts/general_functions.ftl
```diff
@@ -61,12 +61,45 @@ END;
 $$ LANGUAGE plpgsql STABLE STRICT;
 
 
--- function used to convert accented strings into unaccented string
-CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.f_unaccent(text)
-  RETURNS text AS
-$$
-  SELECT public.unaccent('public.unaccent', $1)  -- schema-qualify function and dictionary
-$$  LANGUAGE sql IMMUTABLE PARALLEL SAFE STRICT;
+-- Convert accented string into unaccented string.
```
craigmcnally, line 64: These comments are helpful - thank you for including them

PR #657, opened 2020-05-14
FOLIO-2563: Cancel long running queries

MikhailFokanov:


domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -554,11 +554,19 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
     String element, RoutingContext routingContext, Map<String, String> okapiHeaders,
     Context vertxContext) {
 
+   streamGet(table,clazz,cql,offset,limit,facets,element,routingContext,okapiHeaders,vertxContext,0);
+  }
+
+   public static <T> void streamGet(String table, Class<T> clazz,
+    String cql, int offset, int limit, List<String> facets,
+    String element, RoutingContext routingContext, Map<String, String> okapiHeaders,
+    Context vertxContext, int limitRequestExecutionTime) {
```
julianladisch, line 563: The variable names `limitRequestExecutionTime`, `executionTimeLimit`, and `EXECUTION_TIME_LIMIT` can be replaced by the much shorter `timeout` and `TIMEOUT` without any information loss. Dropping name variants improves consistency.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -554,11 +554,19 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
     String element, RoutingContext routingContext, Map<String, String> okapiHeaders,
     Context vertxContext) {
 
+   streamGet(table,clazz,cql,offset,limit,facets,element,routingContext,okapiHeaders,vertxContext,0);
+  }
+
+   public static <T> void streamGet(String table, Class<T> clazz,
```
julianladisch, line 560: Can you add javadoc for all parameters?
Can you add javadoc what unit (minutes, seconds, ...) the timeout variable has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -554,11 +554,19 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+    String cql, int offset, int limit, List<String> facets,
+    String element, RoutingContext routingContext, Map<String, String> okapiHeaders,
+    Context vertxContext, int limitRequestExecutionTime) {
+
     HttpServerResponse response = routingContext.response();
     try {
       List<FacetField> facetList = FacetManager.convertFacetStrings2FacetFields(facets, JSON_COLUMN);
       CQLWrapper wrapper = new CQLWrapper(new CQL2PgJSON(table + "." + JSON_COLUMN), cql, limit, offset);
-      streamGet(table, clazz, wrapper, facetList, element, routingContext, okapiHeaders, vertxContext);
+      streamGet(table, clazz, wrapper, facetList, element, routingContext, okapiHeaders, vertxContext,limitRequestExecutionTime);
```
julianladisch, line 569:
```suggestion
      streamGet(table, clazz, wrapper, facetList, element, routingContext, okapiHeaders, vertxContext, limitRequestExecutionTime);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -583,7 +591,7 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
   public static <T> void streamGet(String table, Class<T> clazz,
     CQLWrapper filter, List<FacetField> facetList, String element,
     RoutingContext routingContext, Map<String, String> okapiHeaders,
-    Context vertxContext) {
+    Context vertxContext, int limitRequestExecutionTime) {
```
julianladisch, line 594: Can you add javadoc what unit (hours, minutes, ...) the timeout parameter has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
julianladisch: Can you add
```
@SuppressWarnings({"squid:S107"})     // Method has >7 parameters
```
to suppress the warning that sonar raises? https://sonarcloud.io/project/issues?id=org.folio%3Araml-module-builder&pullRequest=657&resolved=false&types=CODE_SMELL

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1033,7 +1041,7 @@ public static void setOptimizedSqlSize(int size) {
    * @param asyncResultHandler
    */
   public static <T, C> void getWithOptimizedSql(String table, Class<T> clazz, Class<C> collectionClazz,
-      String sortField, String cql, int offset, int limit,
+      String sortField, String cql, int offset, int limit, int limitRequestExecutionTime,
```
julianladisch, line 1044: Can you add javadoc what unit (hours, minutes, ...) the timeout parameter has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -104,6 +105,8 @@
 
   static final String COUNT_FIELD = "count";
 
+  private static final long CHECK_FOR_QUERY_TIMEOUT_INTERVAL = 1000;
```
julianladisch, line 108: Can you add javadoc what unit (hours, minutes, ...) the interval has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -497,6 +502,27 @@ private void init() throws Exception {
     }
 
     client = createPgPool(vertx, postgreSQLClientConfig);
+
+    activeConnections = new CopyOnWriteArraySet<>();
+
+    vertx.setPeriodic(CHECK_FOR_QUERY_TIMEOUT_INTERVAL, timerId -> {
+      for (SQLConnection conn : activeConnections) {
```
julianladisch, line 509: Can you move the for loop into a separate method and unit test it?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1711,10 +1738,19 @@ public TotaledResults(RowSet<Row> set, Integer total) {
     Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
 
     streamGet(table, clazz, fieldName, filter, returnIdField, distinctOn,
-      null, replyHandler);
+      null, replyHandler, 0);
   }
 
-  /**
+  public <T> void streamGet(String table, Class<T> clazz, String fieldName,
```
julianladisch, line 1744: Can you add javadoc for the parameters?
Can you add javadoc what unit (hours, minutes, ...) the timeout has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1733,7 +1769,31 @@ public TotaledResults(RowSet<Row> set, Integer total) {
...
+   * @param <T>
+   * @param table
+   * @param clazz
+   * @param fieldName
+   * @param filter
+   * @param returnIdField must be true if facets are in passed
+   * @param distinctOn may be null
+   * @param facets for no facets: null or Collections.emptyList()
+   * @param replyHandler AsyncResult; on success with result {@link org.folio.rest.persist.PostgresClientStreamResult}
+   * @param executionTimeLimit
```
julianladisch, line 1786: Can you explain what unit (hours, minutes, ...) it has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1733,7 +1769,31 @@ public TotaledResults(RowSet<Row> set, Integer total) {
...
+   */
+  @SuppressWarnings({"squid:S00107"})    // Method has >7 parameters
+  public <T> void streamGet(String table, Class<T> clazz, String fieldName,
+    CQLWrapper filter, boolean returnIdField, String distinctOn,
+    List<FacetField> facets, Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler,
+      int executionTimeLimit) {
+
+    getSQLConnection(conn ->
+        streamGet(conn, table, clazz, fieldName, filter, returnIdField,
+            distinctOn, facets, replyHandler),executionTimeLimit);
```
julianladisch, line 1796:
```suggestion
            distinctOn, facets, replyHandler), executionTimeLimit);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2735,6 +2795,20 @@ public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler) {
...
   }
 
+    /**
+   * Run a select query.
+   *
+   * <p>To update see {@link #execute(String, Handler)}.
+   *
+   * @param sql - the sql query to run
+   * @param replyHandler  the query result or the failure
+   */
```
julianladisch, line 2805: Can you add javadoc what unit (hours, minutes, ...) the timeout parameter has?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3000,13 +3074,22 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
     });
   }
 
-  void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
+   void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
+    getSQLConnection(handler, 0);
+   }
```
julianladisch, line 3079:
```suggestion
  void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
    getSQLConnection(handler, 0);
  }
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3000,13 +3074,22 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
...
+
+  void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler, int executionTimeLimit) {
```
julianladisch, line 3081: Can you add javadoc what unit (hours, minutes, ...) the timeout parameter has, and what 0 means?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3000,13 +3074,22 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
...
         handler.handle(Future.failedFuture(res.cause()));
         return;
       }
-      handler.handle(Future.succeededFuture(new SQLConnection(res.result(), null)));
+      SQLConnection sqlConnection = new SQLConnection(res.result(), null, executionTimeLimit);
+      if (executionTimeLimit > 0) {
+         activeConnections.add(sqlConnection);
+      }
+      handler.handle(Future.succeededFuture(sqlConnection));
+      activeConnections.remove(sqlConnection);
```
julianladisch, line 3092: Can you write a unit test that shows that the remove is executed only after the query has completed?

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -6,9 +6,20 @@
 public class SQLConnection {
 
   final PgConnection conn;
+  final int executionTimeLimit;
+  final long acquiringTime;
```
julianladisch, line 10: You only need `final long timeoutTime`. Simply check whether `timeoutTime <= System.currentTimeMillis()`.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -554,11 +554,19 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
     String element, RoutingContext routingContext, Map<String, String> okapiHeaders,
     Context vertxContext) {
 
+   streamGet(table,clazz,cql,offset,limit,facets,element,routingContext,okapiHeaders,vertxContext,0);
```
julianladisch, line 557:
```suggestion
   streamGet(table, clazz, cql, offset, limit, facets, element, routingContext,
     okapiHeaders, vertxContext, 0);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: The parameter ordering makes it difficult to read when calling this method.
Can you put the handler at the end?
`getSQLConnection(int timeout, Handler<AsyncResult<SQLConnection>> handler)`

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -6,9 +6,20 @@
...
   final Transaction tx;
+
   public SQLConnection(PgConnection conn, Transaction tx) {
+   this(conn,tx,0);
```
julianladisch, line 14:
```suggestion
   this(conn, tx, 0);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -6,9 +6,20 @@
...
+  }
+
+
+  public SQLConnection(PgConnection conn, Transaction tx, int executionTimeLimit) {
     this.conn = conn;
     this.tx = tx;
+    this.executionTimeLimit = executionTimeLimit;
+    acquiringTime = System.currentTimeMillis();
+
```
julianladisch, line 23:
```suggestion
```

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -6,9 +6,20 @@
...
   final PgConnection conn;
+  final int executionTimeLimit;
+  final long acquiringTime;
   final Transaction tx;
+
   public SQLConnection(PgConnection conn, Transaction tx) {
+   this(conn,tx,0);
+  }
+
+
```
julianladisch, line 17:
```suggestion
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -497,6 +502,27 @@ private void init() throws Exception {
     }
 
     client = createPgPool(vertx, postgreSQLClientConfig);
+
+    activeConnections = new CopyOnWriteArraySet<>();
```
julianladisch, line 506: This overwrites static `activeConnections` when a second PostgresClient is created.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -497,6 +502,27 @@ private void init() throws Exception {
...
+    vertx.setPeriodic(CHECK_FOR_QUERY_TIMEOUT_INTERVAL, timerId -> {
+      for (SQLConnection conn : activeConnections) {
+        long elapsedTime = System.currentTimeMillis() - conn.acquiringTime;
+        if (elapsedTime > conn.executionTimeLimit) {
+          conn.conn.cancelRequest(ar -> {
+            if (ar.succeeded()) {
+              log.warn(
+                  String.format("Cancelling request due to timeout after : %d ms", elapsedTime));
+            } else {
+              log.warn("Failed to send cancelling request");
```
julianladisch, line 517:
```suggestion
              log.warn("Failed to send cancelling request", ar.cause());
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -497,6 +502,27 @@ private void init() throws Exception {
...
 
     client = createPgPool(vertx, postgreSQLClientConfig);
+
+    activeConnections = new CopyOnWriteArraySet<>();
+
+    vertx.setPeriodic(CHECK_FOR_QUERY_TIMEOUT_INTERVAL, timerId -> {
+      for (SQLConnection conn : activeConnections) {
+        long elapsedTime = System.currentTimeMillis() - conn.acquiringTime;
+        if (elapsedTime > conn.executionTimeLimit) {
+          conn.conn.cancelRequest(ar -> {
```
julianladisch, line 512: How do you ensure that the timeout of a first query doesn't cancel a second query running on the same connection?
```
SELECT 1, pg_sleep(2);  // runs 2 seconds, use timeout = 2500 ms
SELECT 2, pg_sleep(2);  // runs 2 seconds, use timeout = 2500 ms
```
Can you write a unit test for this?

MikhailFokanov: Should it be the case? If the connection is retrieved from pool, it means that request is started. The initial problem id that request should be limited with certain time. So the limitation of queries within the same connection in my opinion makes sense.

MikhailFokanov: Like this:  
  ```
@Test
  public void selectWithTimeout(TestContext context) {
      PostgresClient client = postgresClient();
      client.getSQLConnection(2500, conn -> {
        client.selectSingle(conn, "SELECT 1, pg_sleep(2);", context.asyncAssertSuccess());
        client.selectSingle(conn, "SELECT 1, pg_sleep(2);", context.asyncAssertFailure());
      });
  }
```

File: domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java [hunk 23]
```diff
@@ -497,6 +503,27 @@ private void init() throws Exception {
     }
 
     client = createPgPool(vertx, postgreSQLClientConfig);
+
+
+    vertx.setPeriodic(CHECK_FOR_QUERY_TIMEOUT_INTERVAL, PostgresClient::checkStaleConnections);
+
+  }
+
+  private static void checkStaleConnections(Long timerId) {
```
julianladisch, line 512:
```suggestion
  private static void checkStaleRequests(Long timerId) {
```
We want to cancel requests. Releasing connections is a different issue handled elsewhere (connection release delay).

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -497,6 +503,27 @@ private void init() throws Exception {
...
+    for (SQLConnection conn : activeConnections) {
+      if (conn.timeoutTime >= System.currentTimeMillis()) {
```
julianladisch, line 514: There is no need to call System.currentTimeMillis() in each for iteration, you may use

    long now = System.currentTimeMillis()

before the for loop.

julianladisch, line 514: conn.timeoutTime <= now

or

    now >= conn.timeoutTime

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -55,6 +55,8 @@
   @Rule
   public MockitoRule mockitoRule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
 
+  private int EXECUTION_TIME_LIMIT = 0;
```
julianladisch, line 58: In PostgresClientIT.java this is called `QUERY_TIMEOUT`.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2735,6 +2809,21 @@ public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler) {
...
+    /**
+   * Run a select query.
+   *
+   * <p>To update see {@link #execute(String, Handler)}.
+   *
+   * @param sql - the sql query to run
+   * @param replyHandler  the query result or the failure
+   */
+    public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler,
+        int limitRequestExecutionTime) {
```
julianladisch, line 2821:
```suggestion
        int queryTimeout) {
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2735,6 +2809,21 @@ public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler) {
...
+      getSQLConnection(limitRequestExecutionTime,
```
julianladisch, line 2822:
```suggestion
      getSQLConnection(queryTimeout,
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3001,12 +3090,21 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
   }
 
   void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
+    getSQLConnection(0, handler);
+  }
+
+  void getSQLConnection(int executionTimeLimit, Handler<AsyncResult<SQLConnection>> handler) {
```
julianladisch, line 3096: Can you also rename this executionTimeLimit to queryTimeout?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: Fixed

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: Fixed

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -643,6 +648,33 @@ public void updateSectionSingleQuote(TestContext context) {
     }));
   }
 
+  @Test
+  public void selectWithTimeout(TestContext context) {
+      PostgresClient client = postgresClient();
+      client.getSQLConnection(2500, conn -> {
+        client.selectSingle(conn, "SELECT 1, pg_sleep(2);", context.asyncAssertSuccess());
+        client.selectSingle(conn, "SELECT 1, pg_sleep(2);", context.asyncAssertFailure());
```
julianladisch, line 656: Only one query can run on conn at the same time, for parallel queries a second conn is needed.
Please check the SQLSTATE of the exception returned.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -497,6 +503,28 @@ private void init() throws Exception {
...
+            log.warn(
+                String.format("Cancelling request due to timeout after : %d ms",
+                    now - conn.timeoutTime));
+          } else {
+            log.warn("Failed to send cancelling request", ar.cause());
+          }
+        });
+        activeConnections.remove(conn);
+      }
+    }
```
julianladisch, line 527: The activeConnections CopyOnWriteArraySet is prone to race conditions.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2735,6 +2810,22 @@ public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler) {
...
   }
 
+    /**
+   * Run a select query.
+   *
+   * <p>To update see {@link #execute(String, Handler)}.
+   *
+   * @param sql - the sql query to run
+   * @param replyHandler the query result or the failure
+   * @param queryTimeout query timeout in milliseconds
```
julianladisch, line 2820:
```suggestion
   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1731,7 +1782,31 @@ public TotaledResults(RowSet<Row> set, Integer total) {
...
+   * Stream GET with CQLWrapper and facets {@link org.folio.rest.persist.PostgresClientStreamResult}
+   * @param <T>
+   * @param table
+   * @param clazz
+   * @param fieldName
+   * @param filter
+   * @param returnIdField must be true if facets are in passed
+   * @param distinctOn may be null
+   * @param facets for no facets: null or Collections.emptyList()
+   * @param queryTimeout query timeout in milliseconds
```
julianladisch, line 1800:
```suggestion
   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1711,10 +1740,32 @@ public TotaledResults(RowSet<Row> set, Integer total) {
...
+   * Stream GET with CQLWrapper, no facets {@link org.folio.rest.persist.PostgresClientStreamResult}
+   * @param <T>
+   * @param table
+   * @param clazz
+   * @param fieldName
+   * @param filter
+   * @param returnIdField
+   * @param distinctOn may be null
+   * @param replyHandler AsyncResult; on success with result {@link org.folio.rest.persist.PostgresClientStreamResult}
+   * @param queryTimeout query timeout in milliseconds
```
julianladisch, line 1756:
```suggestion
   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1029,11 +1062,12 @@ public static void setOptimizedSqlSize(int size) {
    * @param cql
    * @param okapiHeaders
    * @param vertxContext
+   * @param queryTimeout query timeout in milliseconds
```
julianladisch, line 1065:
```suggestion
   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -575,20 +606,21 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
    * @param filter
    * @param facetList
    * @param element
-   * @param routingContext
+   * @param queryTimeout query timeout in milliseconds
```
julianladisch, line 609:
```suggestion
   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -554,11 +554,42 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+   * @param offset
+   * @param limit
+   * @param facets
+   * @param element
+   * @param queryTimeout query timeout in milliseconds
+   * @param okapiHeaders
+   * @param vertxContext
+   * @param routingContext
+   */
+  @SuppressWarnings({"squid:S107"})     // Method has >7 parameters
```
julianladisch, line 581:
```suggestion
   @SuppressWarnings({"squid:S107"})     // Method has >7 parameters
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -554,11 +554,42 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+   *
+   * @param <T>
+   * @param table
+   * @param clazz
+   * @param cql
+   * @param offset
+   * @param limit
+   * @param facets
+   * @param element
+   * @param queryTimeout query timeout in milliseconds
```
julianladisch, line 576:
```suggestion
   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Assume two requests that have reached their timeoutTime. Calling cancelRequest for the first is an async call. Before returning the cancelRequest result other processing may happen, for example an SQL result for the second request is returned and processed, and the connection is reused to start a third SQL query. Then checkStaleRequests still has timeout information about the second request and may cancel the request on that connection, but it may cancel the third request.

Why don't you use one timer per request?

    long timerId = vertx.setTimer(queryTimeout, id -> {
      conn.conn.cancelRequest(...);
    });
    if request completes
      vertx.cancelTimer(timerId);

This does not require any considerations about race conditions.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -643,6 +648,40 @@ public void updateSectionSingleQuote(TestContext context) {
     }));
   }
 
+    @Test
+  public void selectWithTimeoutSuccess(TestContext context) {
+      PostgresClient client = postgresClient();
+      client.getSQLConnection(2000, conn -> {
+        client.selectSingle(conn, "SELECT 1, pg_sleep(1);", context.asyncAssertSuccess());
+      });
```
julianladisch, line 656:
```suggestion
      client.getSQLConnection(2000, asyncAssertTx(context, conn -> {
        client.selectSingle(conn, "SELECT 1, pg_sleep(1);", context.asyncAssertSuccess());
      }));
```
Each async callback needs a context.asyncAssertSuccess or context.asyncAssertFail or asyncAssertTx. Otherwise the test will succeed without executing the SELECT.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -643,6 +648,40 @@ public void updateSectionSingleQuote(TestContext context) {
...
+  }
+
+  @Test
+  public void selectWithTimeoutFailure(TestContext context) {
+      PostgresClient client = postgresClient();
+      client.getSQLConnection(1000, conn -> {
+        client.selectSingle(conn, "SELECT 1, pg_sleep(2);", context.asyncAssertFailure());
+      });
```
julianladisch, line 664:
```suggestion
      client.getSQLConnection(1000, asyncAssertTx(context, conn -> {
        client.selectSingle(conn, "SELECT 1, pg_sleep(2);", context.asyncAssertFailure(e -> {
          String sqlState = new PgExceptionFacade(e).getSqlState();
          assertThat(PgExceptionUtil.getMessage(e), sqlState, is("57014"));  // query_canceled
        }));
      }));
```
Each async callback needs a context.asyncAssertSuccess or context.asyncAssertFail or asyncAssertTx. Otherwise the test will succeed without executing the SELECT.

Please check the sqlState returned.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -643,6 +648,40 @@ public void updateSectionSingleQuote(TestContext context) {
...
+  }
+
+  @Test
+  public void connectionIsRemoved(TestContext context) throws InterruptedException {
+      PostgresClient client = postgresClient();
+      Object mutex = new Object();
```
julianladisch, line 670: TestContext uses Async returned by context.async() and not synchronized(mutex): https://vertx.io/docs/vertx-unit/java/#_asynchronous_testing

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Thank you, the unit test selectWithTimeoutFailure runs the sleep.
But there is no cancel because
* `handler.handle(Future.succeededFuture(sqlConnection))` returns immediately after the query has stated,
* `activeConnections.remove(sqlConnection)` is executed immediately after the query has started.

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -6,9 +6,16 @@
 public class SQLConnection {
 
   final PgConnection conn;
+  final long timeoutTime;
```
julianladisch, line 9: The timeoutTime variable should be removed because it is never read.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2971,15 +3042,41 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
...
+        long timerId = vertx.setTimer(queryTimeout, id -> sqlConnection.conn.cancelRequest(ar -> {
+          if (ar.succeeded()) {
+            log.warn(
+                String.format("Cancelling request due to timeout after : %d ms",
+                    queryTimeout));
+          } else {
+            log.warn("Failed to send cancelling request", ar.cause());
+          }
+        }));
+        activeConnectionsTimers.put(sqlConnection, timerId);
```
julianladisch, line 3065: Can you check whether it is possible to save the timerId in sqlConnection so that we can remove activeConnectionsTimer?

MikhailFokanov: I don't think that we should include this completely unrelated to SQLConnection class in this  class.

julianladisch: We previously stored timeoutTime, why can't we store timeout timerId in SQLConnection?
We already pass the SQLConnection to cancelConnectionTimeoutTimer:
`private void cancelConnectionTimeoutTimer(SQLConnection sqlConnection)`

MikhailFokanov: Timeout time is physical (e.g. domain model) attribute, but timerId is synthetic id, which is generated by vert.x

MikhailFokanov: @julianladisch, is my explanation ok?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2971,15 +3042,41 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
...
+      handler.handle(Future.succeededFuture(sqlConnection));
     });
   }
 
+  private void cancelConnectionTimeoutTimer(SQLConnection sqlConnection) {
+    Long timerId = activeConnectionsTimers.remove(sqlConnection);
+    if (timerId != null) {
+      vertx.cancelTimer(timerId);
+    } else {
+      log.warn("Cannot cancel timer: " + timerId);
```
julianladisch, line 3076: There should be no warning if cancelling was disabled with queryTimeout==0.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: timerId will be more reasonable, ok I will add it.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -102,6 +103,9 @@
 
   static final String COUNT_FIELD = "count";
 
+  /** queries timeout checking interval in milliseconds */
+  private static final long CHECK_FOR_QUERY_TIMEOUT_INTERVAL = 1000;
```
julianladisch, line 107: Is this still in use? If not, please remove.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -495,7 +499,9 @@ private void init() throws Exception {
     }
 
     client = createPgPool(vertx, postgreSQLClientConfig);
-  }
+
+
+   }
```
julianladisch, line 504: Can you undo these whitespace changes?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -694,6 +700,7 @@ public void endTx(AsyncResult<SQLConnection> trans, Handler<AsyncResult<Void>> d
       if (sqlConnection.conn != null) {
         sqlConnection.conn.close();
       }
+      cancelConnectionTimeoutTimer(sqlConnection);
```
julianladisch, line 703: Can you make closeAndHandleResult and cancelConnectionTimeoutTimer static?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2705,6 +2758,22 @@ public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler) {
...
+   * Run a select query.
+   *
+   * <p>To update see {@link #execute(String, Handler)}.
+   *
+   * @param sql - the sql query to run
+   * @param replyHandler the query result or the failure
+   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
+   */
+    public void select(String sql, Handler<AsyncResult<RowSet<Row>>> replyHandler,
+        int queryTimeout) {
```
julianladisch, line 2771: Can you move the replyHandler parameter to the end? That will be more consistent with all other methods and also much more readable when called because the replyHandler may be a multi-line argument. It will also ensure a smooth upgrade path to the Vert.x 4 futurisation.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: cancelConnectionTimeoutTimer uses _vertx_ from the PostgresCient instance

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -103,11 +103,15 @@
...
 
+  private int QUERY_TIMEOUT = 0;
+
+/*
   @BeforeClass
   public static void doesNotCompleteOnWindows() {
     final String os = System.getProperty("os.name").toLowerCase();
     org.junit.Assume.assumeFalse(os.contains("win")); // RMB-261
   }
+*/
```
julianladisch, line 114: Can you either remove `/*` and `*/`? You may create a separate issue and PR to completely remove `doesNotCompleteOnWindows()`.

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -7,8 +7,18 @@
 
   final PgConnection conn;
   final Transaction tx;
+  private Long timeId;
```
julianladisch, line 10: Can you explain why this variable is not final? See https://github.com/folio-org/raml-module-builder/pull/657#discussion_r432584436 how to do this.

MikhailFokanov: I don't like idea to add additional redundant complexity of additional lambda-callbacks just for making this variable not final.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
MikhailFokanov: done

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
julianladisch: You don't need to introduce the lambda to change `Long timeId` to `final Long timerId`. Only some reordering of lines, as shown in https://github.com/folio-org/raml-module-builder/pull/657#discussion_r432584436 , is needed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Yes, ok, I havn't seen that.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -495,7 +495,7 @@ private void init() throws Exception {
     }
 
     client = createPgPool(vertx, postgreSQLClientConfig);
-  }
+   }
```
julianladisch, line 498: Can you indent using 2 spaces, not 3 spaces?

domain-models-runtime/src/main/java/org/folio/rest/persist/SQLConnection.java
```diff
@@ -7,8 +7,18 @@
...
+
   public SQLConnection(PgConnection conn, Transaction tx) {
     this.conn = conn;
     this.tx = tx;
   }
+
+  public void setTimeId(Long timeId) {
+    this.timeId = timeId;
```
julianladisch, line 18: Can you rename it to timerId?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1709,10 +1706,32 @@ public TotaledResults(RowSet<Row> set, Integer total) {
...
+   * @param returnIdField
+   * @param distinctOn may be null
+   * @param replyHandler AsyncResult; on success with result {@link org.folio.rest.persist.PostgresClientStreamResult}
+   * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
+   */
+   @SuppressWarnings({"squid:S00107"})    // Method has >7 parameters
+  public <T> void streamGet(String table, Class<T> clazz, String fieldName,
+      CQLWrapper filter, boolean returnIdField, String distinctOn,
+      Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler,
+      int queryTimeout) {
```
julianladisch, line 1728: Can you switch the replyHandler and queryTimeout parameters?
It's much easier to read the method invocation when the muli-line replyHandler is at the end.

MikhailFokanov: done

MikhailFokanov: up

PR #659, opened 2020-05-15
RMB-569: Upgrade using schema.json comparison

julianladisch:
* SchemaMaker: new previousSchema variable, can trigger deletion of tables and foreign keys
* foreign_keys.ftl: Only create foreign key trigger if an "ADD" foreign key entry exists
* foreign_keys.ftl: Don't drop auditing table when auditing table name is null
* SchemaMakerTest: Extract duplicate code into schema and schemaMaker methods.
* TenantAPI: Use schema.json comparison for upgrades.

### DIFF
--- a/dbschema/src/main/java/org/folio/rest/persist/ddlgen/ForeignKeys.java
+++ b/dbschema/src/main/java/org/folio/rest/persist/ddlgen/ForeignKeys.java
@@ -129,7 +129,8 @@ public class ForeignKeys extends Field {
     return "ForeignKeys [tableAlias=" + tableAlias + ", targetTable=" + targetTable
         + ", targetTableAlias=" + targetTableAlias
         + ", fieldName=" + fieldName
-        + (targetPath == null ? "" : "targetPath=" + targetPath)
+        + (targetPath == null ? "" : ", targetPath=" + targetPath)
+        + ", tOps=" + gettOps()
         + "]";
   }
 
--- a/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
@@ -5,8 +5,6 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import javax.ws.rs.core.Response;
 
 import org.apache.commons.io.IOUtils;
@@ -217,7 +215,6 @@ public class TenantAPI implements Tenant {
         promise.complete(schema);
       } catch (Exception e) {
         promise.fail(e);
-        return;
       }
     });
     return promise.future();
@@ -293,22 +290,8 @@ public class TenantAPI implements Tenant {
    * @throws TemplateException when processing templates/db_scripts/schema.json fails
    */
   public Future<String> sqlFile(Context context, String tenantId, TenantAttributes tenantAttributes) {
-    AtomicBoolean tenantExists = new AtomicBoolean();
-
     return tenantExists(context, tenantId)
-    .compose(tenantExistsResult -> {
-      tenantExists.set(tenantExistsResult);
-      return previousSchema(context, tenantId, tenantExists.get());
-    }).compose(previousSchema -> {
-      try {
-        String sqlFile = sqlFile(tenantId, tenantExists.get(), tenantAttributes, previousSchema);
-        return Future.succeededFuture(sqlFile);
-      } catch (IOException e) {
-        throw new UncheckedIOException(e);
-      } catch (TemplateException e) {  // checked exception from main.tpl parsing
-        throw new IllegalArgumentException(e);
-      }
-    });
+    .compose(tenantExists -> sqlFile(context, tenantId, tenantAttributes, tenantExists));
   }
 
   @Validate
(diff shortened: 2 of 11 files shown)


dbschema/src/main/java/org/folio/rest/persist/ddlgen/ForeignKeys.java
```diff
@@ -129,7 +129,8 @@ public String toString() {
     return "ForeignKeys [tableAlias=" + tableAlias + ", targetTable=" + targetTable
         + ", targetTableAlias=" + targetTableAlias
         + ", fieldName=" + fieldName
-        + (targetPath == null ? "" : "targetPath=" + targetPath)
+        + (targetPath == null ? "" : ", targetPath=" + targetPath)
+        + ", tOps=" + gettOps()
```
adamdickmeiss, line 133: What the previous code at error ? Missing ","?

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -226,40 +264,75 @@ public String sqlFile(String tenantId, boolean tenantExists, TenantAttributes te
...
+  public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers,
       Handler<AsyncResult<Response>> handlers, Context context)  {
 
     String tenantId = TenantTool.tenantId(headers);
     log.info("sending... postTenant for " + tenantId);
-    if (entity != null) {
-      log.debug("upgrade from " + entity.getModuleFrom() + " to " + entity.getModuleTo());
+    if (tenantAttributes != null) {
+      log.debug("upgrade from " + tenantAttributes.getModuleFrom() + " to " + tenantAttributes.getModuleTo());
     }
```
adamdickmeiss, line 307: This logs what Okapi passes to it, right? Do we even use moduleFrom or moduleTo for anything mechanical anymore? If so, perhaps we could log what RMB sees as from-to-version?

dbschema/src/main/java/org/folio/rest/persist/ddlgen/ForeignKeys.java
julianladisch: toString() is used for debugging only. The comma doesn't have any influence on production code.

adamdickmeiss: ok

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
julianladisch: Only recently we started storing the correct version in the rmb_internal table: https://github.com/folio-org/raml-module-builder/commit/3dbf2ed9905bbca558b8b35e25b40115a65fea4a 
Therefore we must use the version number passed in from Okapi for Goldenrod.
SchemaMaker and schema.json use them.

adamdickmeiss: ok

craigmcnally: But this is a possible improvement for Q3?

julianladisch: Sure. This is why we store rmb version and schema.json in the db.

craigmcnally:
Part of me feels like this makes it easier to "accidentally" remove a table (and the data in it), but I can't really think of a realistic scenario where that might happen...  Maybe if a module is renaming a table?  in that case they'd need to make sure that migration scripts run before the generated scripts run.

domain-models-runtime/src/main/java/org/folio/rest/persist/ddlgen/SchemaMaker.java
```diff
@@ -122,6 +126,49 @@ private String handleDelete() throws IOException, TemplateException {
...
+      List<ForeignKeys> oldForeignKeys = oldTable.getForeignKeys();
+      if (oldForeignKeys == null || oldForeignKeys.isEmpty()) {
+        return;
+      }
+      List<ForeignKeys> newForeignKeys =
+          newTable.getForeignKeys() == null ? Collections.emptyList() : newTable.getForeignKeys();
+      List<ForeignKeys> allForeignKeys = new ArrayList<>(newForeignKeys);
+      oldForeignKeys.forEach(oldForeignKey -> {
+        if (newForeignKeys.stream().anyMatch(newForeignKey ->
+            oldForeignKey.getFieldName().equals(newForeignKey.getFieldName()))) {
```
grabsefx, line 159: @adamdickmeiss @julianladisch oldForeignKey.getFieldName() causes NPE in case of upgrading tables which contains cross-index foreign keys without field names like
`    {
      "tableName": "titles",
      "fromModuleVersion": "mod-orders-storage-11.0.0",
      "withMetadata": true,
      "foreignKeys": [
        {
          "fieldName": "poLineId",
          "targetTable":      "po_line",
          "targetTableAlias": "poLine",
          "tableAlias": "titles",
          "tOps": "ADD"
        },
        {
          "targetPath": ["poLineId", "purchaseOrderId"],
          "targetTable":      "purchase_order",
          "targetTableAlias": "purchaseOrder",
          "tableAlias": "titles",
          "tOps": "ADD"
        }
      ]
    },`

PR #661, opened 2020-05-15
Try using idleTimeout for PgConnectOptions RMB-620

adamdickmeiss:


domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -86,7 +87,6 @@ public void configDefault() throws Exception {
     assertThat(config.getString("host"), is("127.0.0.1"));
     assertThat(config.getInteger("port"), is(6000));
     assertThat(config.getString("username"), is("username"));
-    assertThat(config.getInteger("connectionReleaseDelay"), is(60000));
```
julianladisch, line 89: You probably want to check that it is null because the default is set in createPgConnectOptions

adamdickmeiss: I don't understand. It's always set to some non-null value in createPgConnectOptions .

PR #663, opened 2020-05-15
Update to Vert.x 3.9.0 RMB-619

adamdickmeiss:
Needs a little more work in "fake" connections.. so that execute fails with error (rather than query method returning null).

### DIFF
--- a/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
+++ b/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Collector;
 
 import io.vertx.core.AsyncResult;
@@ -36,9 +37,7 @@ import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowSet;
 import io.vertx.sqlclient.SqlResult;
 import io.vertx.sqlclient.Transaction;
-import io.vertx.sqlclient.Tuple;
 import io.vertx.sqlclient.impl.RowDesc;
-import org.drools.core.rule.QueryImpl;
 import org.folio.rest.persist.facets.FacetField;
 import org.folio.rest.persist.helpers.LocalRowSet;
 import org.folio.rest.tools.utils.NetworkUtils;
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -66,7 +66,7 @@
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-stack-depchain</artifactId>
-        <version>3.8.5</version>
+        <version>3.9.0</version>
```
julianladisch, line 69: Can you also update the vertx version in sample/pom.xml and sample2/pom.xml?
Can I remove the other RMB-619 branch mentioned on Slack and in the Development section on https://issues.folio.org/browse/RMB-619?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -321,73 +325,21 @@ public boolean isSSL() {
...
 
     @Override
-    public <R> PgConnection preparedBatch(String s, List<Tuple> list, Collector<Row, ?, R> collector, Handler<AsyncResult<SqlResult<R>>> handler) {
-      handler.handle(Future.failedFuture("not implemented"));
-      return this;
+    public void close() {
     }
   }
 
+  @Ignore
```
julianladisch, line 342: Can you explain the ignores?
For example `@Ignore("Needs more work to make the fake connection work again")`

adamdickmeiss: Back now (not ignored)

pom.xml
adamdickmeiss: Updated, sample and sample2.. Funny, mvn versions:set  does not set it (it was also wrong for the 29.5.0-SNAPSHOT, where sample, sample2 was pointing to an older version).

Perhaps it's time to retire sample and sample2?

I'm confused what branch you are talking about.

julianladisch: The pom.xml of sample and sample2 contains `<vertx-version>3.8.5</vertx-version>`, this should be updated.

julianladisch: sample can be removed.
sample2 servers as a complete integration test. If we can integrate this into domain-models-runtime we can remove sample2 as well.

julianladisch: I talk about this branch: https://github.com/folio-org/raml-module-builder/compare/RMB-619-vert.x-3.9.0

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
julianladisch: Very good!

pom.xml
adamdickmeiss: Done

adamdickmeiss: OK. It took a while to understand. Need a break. So we both did the same work?

PR #675, opened 2020-05-20
Fix stringType false seems to produce invalid SQL RMB-627

adamdickmeiss:


dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
```diff
@@ -133,6 +134,9 @@ public String getFinalTruncatedSqlExpression(String tableLoc) {
     if (this.getMultiFieldNames() != null) {
       return "left(" + getFinalSqlExpression(tableLoc) + ",600)";
     }
+    if (!this.isStringType()) {
+      return this.fieldPath;
+    }
```
julianladisch, line 139: If the javadoc says "stringType=false does not truncate" then it should go before the `if (this.getMultiFieldNames() != null)` that also truncates.

dbschema/src/test/java/org/folio/rest/persist/ddlgen/IndexTest.java
```diff
@@ -161,4 +161,24 @@ void sqlsetMultiFieldNamesWrap() {
...
     assertEquals("lower(concat_space_sql(test_table.jsonb->>'test1' , test_table.jsonb->'test2'->>'test3'))",idx.getFinalSqlExpression("test_table"));
   }
+
+  @Test
+  void indexStringTypeTrue() {
+    Index idx = new Index();
+    idx.setFieldName("testField");
+    idx.setStringType(true);
+    idx.setupIndex();
+    assertEquals("left(lower(f_unaccent(jsonb->>'testField')),600)",idx.getFinalTruncatedSqlExpression("test_table"));
```
julianladisch, line 171:
```suggestion
    assertEquals("left(lower(f_unaccent(jsonb->>'testField')),600)",
      idx.getFinalTruncatedSqlExpression("test_table"));
```

dbschema/src/test/java/org/folio/rest/persist/ddlgen/IndexTest.java
```diff
@@ -161,4 +161,24 @@ void sqlsetMultiFieldNamesWrap() {
...
+  }
+
+  @Test
+  void indexStringTypeFalse() {
+    Index idx = new Index();
+    idx.setFieldName("testField");
+    idx.setStringType(false);
+    idx.setupIndex();
+    assertEquals("(jsonb->'testField')",idx.getFinalTruncatedSqlExpression("test_table"));
```
julianladisch, line 180:
```suggestion
    assertEquals("(jsonb->'testField')", idx.getFinalTruncatedSqlExpression("test_table"));
```

dbschema/src/main/java/org/folio/rest/persist/ddlgen/Index.java
adamdickmeiss: Would it ever possible to combine the two?

dbschema/src/test/java/org/folio/rest/persist/ddlgen/IndexTest.java
adamdickmeiss: Reformatting all tests there (break or space after comma)

PR #678, opened 2020-05-25
Fix null pointer exception in deserializeRow RMB-629

adamdickmeiss:


domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2730,6 +2731,36 @@ public void streamGetWithFacetsAndFilter(TestContext context) throws IOException
...
+      context.assertEquals("First edition", resultInfo.getFacets().get(0).getFacetValues().get(0).getValue());
+      context.assertEquals(3, resultInfo.getFacets().get(0).getFacetValues().get(0).getCount());
+      context.assertEquals("title", resultInfo.getFacets().get(1).getType());
+      context.assertEquals(3, resultInfo.getFacets().get(1).getFacetValues().size());
+      context.assertEquals(1, resultInfo.getFacets().get(1).getFacetValues().get(0).getCount());
+      context.assertEquals(1, resultInfo.getFacets().get(1).getFacetValues().get(1).getCount());
+      context.assertEquals(1, resultInfo.getFacets().get(1).getFacetValues().get(2).getCount());
+      async.complete();
+    }));
+    async.awaitSuccess();
```
julianladisch, line 2761: Async is not needed because there is only one async callback and that callback has a context.asyncAssertSuccess() handler that automatically awaits success.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -3463,20 +3496,52 @@ public void getCQLWrapperWithFacets(TestContext context) throws IOException, Fie
...
+
+              context.assertEquals(1, retFacets.size());
+              context.assertEquals("edition", retFacets.get(0).getType());
+              context.assertEquals(1, retFacets.get(0).getFacetValues().get(0).getCount());
+              context.assertEquals("First edition", retFacets.get(0).getFacetValues().get(0).getValue().toString());
+              context.assertEquals(1, objs.size());
+              context.assertEquals("70fb4e66-cdf1-11e8-a8d5-f2801f1b9fd1", objs.get(0).getId());
+            } catch (Exception ex) {
+              context.fail(ex);
+            }
```
julianladisch, line 3539: The lines

            context.assertTrue(handler.succeeded());

            try {

            } catch (Exception ex) {
              context.fail(ex);
            }

can be removed when replacing `handler -> {` by `context.asyncAssertSuccess(result -> {`.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2545,7 +2545,7 @@ public ResultsHelper(Class<T> clazz) {
     return (int) list.stream()
         .filter(e -> !(e instanceof Facet) &&
                      !((e instanceof Map) &&
-                         ((Map) e).size() == 1 && ((Map) e).containsKey(COUNT_FIELD)))
+                      ((Map) e).size() == 1 && ((Map) e).containsKey(COUNT_FIELD)))
```
julianladisch, line 2548: This indentation change is misleading because the `!` in the previous line also includes this line.

domain-models-runtime/src/test/resources/mockdata/poLines.json
```diff
@@ -1,6 +1,6 @@
...
...chase","adjustment": "596336cf-eb47-4e09-9e28-e20aa464adb7","alerts": [],"cancellation_restriction": false,"cancellation_restriction_note": "","claims": ["57c5d666-cabe-11e8-a8d5-f2801f1b9fd1"],"collection": true,"cost": "57c3d792-cabe-11e8-a8d5-f2801f1b9fd1","description": "","details": "57c5d792-cabe-11e9-a8d5-f2801f1b9fd1","donor": "","eresource": "57c5d803-cabe-11e8-a8d5-f2801f1b9fd1","fund_distribution": ["5de2cad6-cde3-11e8-a8d5-f2801f1b9fd1"],"location": "57c5d9ea-cabe-11e8-a8d5-f2801f1b9f04","order_format": "Electronic Resource","owner": "Technical Services Department","payment_status": "Awaiting Payment","physical": "a0fb7332-cdf1-11e8-a8d5-f2801f1b9fd1","po_line_description": "","po_line_number": "80","po_line_workflow_status": "Closed","publication_date": "2019","publisher": "American Chemical Society","receipt_date": null,"reporting_codes": [],"requester": "","rush": false,"selector": "Buchanan","source": "5b4b6f41-c5c6-4280-858e-33fba452a334","tags": [],"title": "American Chemical Society Web Editions","vendor_detail": "a0fb8b56-cdf1-11e8-a8d5-f2801f1b9fd1", "metadata":{ "createdDate": "2018-07-18T00:00:00Z",  "createdByUserId": "28d102a2-d137-11e8-a8d5-f2801f1b9fd1"}}
```
julianladisch, line 6: Can you add a `count` field of type integer again (and add it to Poline.java) to make sure this is parsed as a field of a regular record and not as the count record?
You may remove other fields if you like.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: What my editor does. I'll reformat.

domain-models-runtime/src/test/resources/mockdata/poLines.json
adamdickmeiss: No. I'd rather not do that. Then my case is not illustrated.. count field will be seen as normal record *and* the getInteger(COUNT_FIELD) is not executed (because mapping works for the Poline Object).

julianladisch: Ok, can you add another unit test for this use case:
The RAML JSON file declares a record with "id" (uuid string) and "count" (integer) field.
This results in a POJO with id and count field.
Either RMB cannot handle this because of the name clash. Then it should be documented and RMB should reject such a RAML JSON entity with an error message.
Or RMB can properly handle this because RMB can distinguish a POJO count field from the query count field. Then we need a unit test for this case.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: I'm probably going to ignore the remove async stuff in the future. Bottomline is that async covers all cases, while the others can only be used in special cases.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
adamdickmeiss: OK

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Going away completely!

domain-models-runtime/src/test/resources/mockdata/poLines.json
adamdickmeiss: The bottomline is.. The count object can NOT be distinguished from another count object. The solution is obvious: change SQL to just not return that object at all. It's pointless, now that qe have a separate counting process anyway. Doing in next PR which simplifies a lot.

adamdickmeiss: Given a JSON object that can be deserilized to TWO pojos, we're in trouble. Let's not go there.

PR #687, opened 2020-06-02
RMB-636: Support PATCH requests

psmagin:


domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -251,12 +251,13 @@ public void start(Promise<Void> startPromise) throws Exception {
     final BodyHandler handler = BodyHandler.create();
 
     // IMPORTANT!!!
-    // the body of the request will be read into memory for ALL PUT requests
+    // the body of the request will be read into memory for ALL PUT, ALL PATCH requests
```
julianladisch, line 254: Can you update the documentation accordingly:
https://github.com/folio-org/raml-module-builder/blob/master/README.md

julianladisch, line 254: Okapi does not support PATCH yet: https://github.com/folio-org/okapi/blob/60b32aa929dc0d2726671e814ba95a0293b9bd13/okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java#L47
Please create a Jira for RMB and a Jira for Okapi.

julianladisch: Thanks!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -576,22 +576,6 @@ private void logPostgresConfig() {
...
-   *   Configuration
-   *   </a>.
-   * which is no longer in actual use.
-   *
-   * @return
-   */
-  public JsonObject getConnectionConfig(){
-    return postgreSQLClientConfig;
-  }
-
```
julianladisch, line 594: Can you merge latest master into your branch?

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
julianladisch: Thank you for creating https://issues.folio.org/browse/RMB-636 .
Do you want to create the Okapi Jira yourself, or do you want us to do it?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
psmagin: I am sorry, I have some problems with that. I created another PR https://github.com/folio-org/raml-module-builder/pull/690

psmagin: I will close this one.

psmagin:
See PR https://github.com/folio-org/raml-module-builder/pull/690

PR #690, opened 2020-06-03
RMB-636 - Support PATCH requests

psmagin:


psmagin:
I created Jira for okapi support https://issues.folio.org/browse/RMB-636. Should I make changes in code and create PR also?

julianladisch:
It will speed things up if you create the Okapi PR as well.

domain-models-runtime-it/src/main/java/org/folio/rest/impl/MyitemsApi.java
```diff
@@ -49,4 +49,11 @@ public void putMyitemsByMyitemId(String id, String lang, Myitem entity, Map<Stri
     PgUtil.put(TABLE, entity, id, okapiHeaders, vertxContext,
         PutMyitemsByMyitemIdResponse.class, asyncResultHandler);
   }
+
+  @Override
+  public void patchMyitemsByMyitemId(String id, Myitem entity, Map<String, String> okapiHeaders,
+                                     Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    PgUtil.put(TABLE, entity, id, okapiHeaders, vertxContext,
```
grabsefx, line 56: As far as I understand PgUtil.put method will completely overwrite an existing record. Probably it is needed to implement new PgUtil.patch(...) method. Please, correct me if I wrong.

julianladisch: This issue and PR is about RAML and RestVerticle only.
This is needed for https://issues.folio.org/browse/MODKBEKBJ-447 .
Adding PATCH support in PostgresClient or PgUtil is out of scope.

domain-models-runtime-it/src/main/java/org/folio/rest/impl/MyitemsApi.java
```diff
@@ -49,4 +49,11 @@ public void putMyitemsByMyitemId(String id, String lang, Myitem entity, Map<Stri
     PgUtil.put(TABLE, entity, id, okapiHeaders, vertxContext,
         PutMyitemsByMyitemIdResponse.class, asyncResultHandler);
   }
+
+  @Override
+  public void patchMyitemsByMyitemId(String id, Myitem entity, Map<String, String> okapiHeaders,
+                                     Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
```
julianladisch, line 55:
```suggestion
                                     Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
    // We test PATCH support of RAML and RestVerticle only.
    // PgUtil doesn't support patch yet, so we use put for the time being.
```

domain-models-runtime-it/src/main/java/org/folio/rest/impl/MyitemsApi.java
psmagin: The main goal of this PR is to have RMB ability to process PATCH requests. 

`MyItemApi` is using only for test purposes. 
I think that implementing PgUtil.patch(...) is out of scope. If you think that this method is needed than I can create a Jira ticket for it.

hjiebsco: It would be nice to add a comment in the code about using PgUtil.put to cover patch method for the time being.

domain-models-runtime-it/ramls/myitems.raml
```diff
@@ -44,6 +44,13 @@ resourceTypes:
       collection-item:
         exampleItem: !include examples/myitem.json
         schema: myitem
+    patch:
+      body:
+        application/json:
+          type: myitem
+      responses:
+        204:
+          description: Accepted
```
hjiebsco, line 53: I noticed there is no PUT or DELETE defined here. I am curious why adding PATCH would make a real difference?

psmagin: For PUT and DELETE respond common ResourceType from raml-util https://github.com/folio-org/raml-module-builder/blob/c3398f318548f22f460ee0840279fd07b5ce817a/domain-models-runtime-it/ramls/myitems.raml#L23

domain-models-runtime-it/src/main/java/org/folio/rest/impl/MyitemsApi.java
psmagin: @hjiebsco 
Comment was added

domain-models-runtime-it/ramls/myitems.raml
hjiebsco: I see. Should Patch be added to item-collection.raml then?

julianladisch: There are other tests for PUT and DELETE.
DELETE does not have a body and therefore doesn't need the body handler, PUT already has the body handler, and this PR adds PATCH for the body handler: https://github.com/folio-org/raml-module-builder/blob/960bc78bc74079ae2e1a339063ac6e06b4d06301/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java#L251-L265
If you comment out the line
`router.patch().handler(handler);`
the unit test MyitemsApiTest.patch will fail because the body is null:
```
java.lang.IllegalArgumentException: HV000116: The object to be validated must not be null.
        at org.hibernate.validator.internal.util.Contracts.assertNotNull(Contracts.java:41)
        at org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:190)
        at org.folio.rest.RestVerticle.isValidRequest(RestVerticle.java:1444)
```

hjiebsco: Thanks. So if I understand the PR correctly, it is about `router.patch().handler(handler);`. If a module wants to use PATCH, it has to define it just like `myitems.raml`. There is no plan to make PATCH into a reusable raml file like `item-collection.raml` yet.

PR #692, opened 2020-06-04
RMB-637: Run ANALYZE after index (re-)creation

julianladisch:
* Collect tables that need ANALYZE in rmb_internal_analyze
* At the end run ANALYZE for all of them

Prepending ${myuniversity}_${mymodule}. is no longer needed
because we set search_path.


julianladisch:
Sobha successfully tested it with a copy of the bugfest environment. Results are in the spreadsheet linked in https://issues.folio.org/browse/FOLIO-2625 .

PR #694, opened 2020-06-04
streamGet, getWithOptimizedSql wo queryTimeout RMB-639

adamdickmeiss:
We do need a builder pattern here.. Many arguments and many variants lead to a combinatorial explosion.


skoczko:
Am I correct this is re-introducing overloaded methods w/o the queryTimeout param? If so, I vote for this change so we remain backwards compatible

skoczko:
@MikhailFokanov Don't we also need an IT for the version with queryTimeout?

PR #698, opened 2020-06-08
RMB-643: Wrapped query parameter types

wwelling:
requires https://github.com/folio-org/raml/pull/125

### DIFF
--- a/domain-models-runtime-it/ramls/wrappedboolean.raml
+++ b/domain-models-runtime-it/ramls/wrappedboolean.raml
@@ -1,0 +1,25 @@
+#%RAML 1.0
+title: WrappedBoolean
+baseUri: http://github.com/org/folio/raml-module-builder/domain-models-runtime-it
+
+documentation:
+  - title: WrappedBoolean API
+    content: API used for Integration tests
+
+uses:
+  ramltojaxrs: raml-util/library/ramltojaxrs.raml
+
+/wrapped-boolean:
+  get:
+    queryParameters:
+      value:
+        type: boolean
+        (ramltojaxrs.types):
+          plugins:
+            - name: core.box
+        example: false
+        required: false
+    responses:
+      200:
+        body:
+          text/plain:
--- a/domain-models-runtime-it/src/test/java/org/folio/rest/WrappedBooleanApiTest.java
+++ b/domain-models-runtime-it/src/test/java/org/folio/rest/WrappedBooleanApiTest.java
@@ -1,0 +1,39 @@
+package org.folio.rest;
+
+import static io.restassured.RestAssured.*;
+import static org.hamcrest.Matchers.*;
+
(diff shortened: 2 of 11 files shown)


domain-models-runtime-it/src/test/java/org/folio/rest/WrappedBooleanApiTest.java
```diff
@@ -0,0 +1,39 @@
+package org.folio.rest;
+
+import static io.restassured.RestAssured.*;
+import static org.hamcrest.Matchers.*;
+
+import org.junit.jupiter.api.Test;
+
+public class WrappedBooleanApiTest extends ApiTestBase {
```
julianladisch, line 8: Can you remove `public`? Sonar complains about it: https://sonarcloud.io/project/issues?id=org.folio%3Araml-module-builder&pullRequest=698&resolved=false&types=CODE_SMELL

domain-models-runtime-it/ramls/wrappedboolean.raml
```diff
@@ -0,0 +1,25 @@
+#%RAML 1.0
```
julianladisch, line 1: Can you add a short boxing documentation to README.md with links to library/ramltojaxrs.raml and ramls/wrappedboolean.raml?

wwelling: Do you have a suggestion where at in the README would be appropriate?

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -1339,7 +1339,7 @@ else if(bodyContent.length() > 0) {
                 if (defaultVal != null) {
                   paramArray[order] = Integer.valueOf((String) defaultVal);
                 } else {
-                  paramArray[order] = 0;
+                  paramArray[order] = valueType.equals("int") ? 0 : null;
```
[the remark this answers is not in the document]
wwelling: I was thinking parameter type support could be more customizable. Something like parameter processors. Have default parameter processors in RMB and support custom parameter processors from modules with support for custom types.

This would make parameter support much more extensible and could help reduce number of arguments in endpoint methods.

I would much rather do that than just extract less than ideal code into a method to just be able to unit test it.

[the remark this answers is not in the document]
wwelling: Added integer query parameter to books API which covers the case for integer primitive query parameter. Added an example wrapped integer API and tests which covers the wrapped integer cases. This should cover the lines well enough. RestVerticle is in need of unit testing and possibly refactoring, but I feel that is out of scope of this issue.

wwelling:
@julianladisch are you going to make me write unit tests for query parameters for lines 1319-1388 for your approval?

PR #702, opened 2020-06-10
Rmb 644 generic sort implementation

adamdickmeiss:
Purpose of this is to make streamGet and get to use optimized sorting in general. First step is to make an extension to CQL2PgJSON .. then make CQLWrapper and PostgresClient use new facilities. In particular, CQL2PgJSON is now responsible for generating the complete SQL for query... means that it is aware of LIMIT and OFFSET .. However, for now, result counting is left out - and so this is a separate SQL query.. That could be changed later.


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -285,20 +292,39 @@ public String cql2pgJson(String cql) throws QueryValidationException {
    * @throws QueryValidationException
    */
   public SqlSelect toSql(String cql) throws QueryValidationException {
+    return toSql(cql, 0, 0);
```
julianladisch, line 295: limit=0 will return no record, do you want limit=-1 for all records?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -285,20 +292,39 @@ public String cql2pgJson(String cql) throws QueryValidationException {
...
+  }
+
+  public SqlSelect toSql(String cql, int offset, int limit) throws QueryValidationException {
```
julianladisch, line 298: Can you add javadoc (params, exception) for this public method?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -404,7 +430,74 @@ private SqlSelect toSql(CQLSortNode node) throws QueryValidationException {
...
+    } else {
+      StringBuilder sql = new StringBuilder(SELECT_STR + column + FROM_STR + table + WHERE_STR + where
+          + ORDER_BY + order.toString());
+      if (offset > 0) {
+        sql.append(OFFSET_STR + offset);
+      }
+      if (limit >= 0) {
+        sql.append(LIMIT_STR + limit);
+      }
+      fullQuery = sql.toString();
```
julianladisch, line 498:
```suggestion
      StringBuilder sql = new StringBuilder()
          .append(SELECT_STR).append(column)
          .append(FROM_STR).append(table)
          .append(WHERE_STR).append(where)
          .append(ORDER_BY).append(order.toString());
      if (offset > 0) {
        sql.append(OFFSET_STR).append(offset);
      }
      if (limit >= 0) {
        sql.append(LIMIT_STR).append(limit);
      }
      fullQuery = sql.toString();
```
StringBuilder has memory and performance advantages, but only if you replace `+` String concatenation by ` append()`.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -404,7 +430,74 @@ private SqlSelect toSql(CQLSortNode node) throws QueryValidationException {
       order.append(wrapForLength(wrapInLowerUnaccent(vals.getIndexText(), modifiers))).append(desc).append(", ")
       .append(wrapInLowerUnaccent(vals.getIndexText(), modifiers)).append(desc);
     }
-    return new SqlSelect(where, order.toString());
+    if (column == null) {
+      return new SqlSelect(where, order.toString(), null);
+    }
+    String fullQuery = null;
+    if (node.getSortIndexes().size() == 1 && limit > 0 && offset >= 0) {
```
julianladisch, line 437: Consider to invert the if expression and have the short code block of the current else clause at the beginning:
```
if (node.getSortIndexes().size() != 1 || limit <= 0 || offset < 0) {
  StringBuilder sql = ...
  return new SqlSelect(where, order.toString(), sql.toString());
}
ModifierSet modifierSet = ...
```
This will also reduce the indentation of the optimized path.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -285,20 +292,39 @@ public String cql2pgJson(String cql) throws QueryValidationException {
...
-      return toSql((CQLSortNode) node);
+      return toSqlSort((CQLSortNode) node, offset, limit, table, column);
+    }
+    String where = pg(node);
+    StringBuilder sql = new StringBuilder(SELECT_STR + column + FROM_STR + table + WHERE_STR + where);
+    if (offset > 0) {
+      sql.append(OFFSET_STR + offset);
+    }
+    if (limit >= 0) {
+      sql.append(LIMIT_STR + limit);
```
julianladisch, line 325: StringBuilder has memory and performance advantages, but only if you replace `+` String concatenation by append().
```suggestion
    StringBuilder sql = new StringBuilder().append(SELECT_STR).append(column)
                                           .append(FROM_STR).append(table)
                                           .append(WHERE_STR).append(where);
    if (offset > 0) {
      sql.append(OFFSET_STR).append(offset);
    }
    if (limit >= 0) {
      sql.append(LIMIT_STR).append(limit);
```

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -367,7 +393,8 @@ private static String wrapQueryExpression(String term, Index index) {
   }
 
   @SuppressWarnings("squid:S135")  // suppress "reduce to one continue in for loop"
-  private SqlSelect toSql(CQLSortNode node) throws QueryValidationException {
+  private SqlSelect toSqlSort(CQLSortNode node, int offset, int limit, String table, String column)
```
julianladisch, line 396: Can you remove `private` and write some unit tests to cover the border cases of limit and offset?

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
julianladisch: This is almost the same code as in toSql(CQLNode). Consider to extract it into a method
```
static String sqlBuilder(CharSequence column, CharSequence table, CharSequence where, CharSequence order, int offset, int limit)
```
or into an SqlSelect constructor
```
public SqlSelect(CharSequence column, CharSequence table, CharSequence where, CharSequence order, int offset, int limit)
```

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -129,8 +128,13 @@ public void select(CQL2PgJSON aCql2pgJson, String sqlFile, String cql, String ex
       assertEquals("CQL: " + cql + ", SQL: " + sql, expectedNames, actualNames);
     } catch (QueryValidationException | SQLException e) {
       logger.fine("select: " + e.getClass().getSimpleName()
-        + " for query " + cql + " : " + e.getMessage());
+          + " for query " + cql + " : " + e.getMessage());
       if (! expectedNames.startsWith("!")) {
+        int pos = e.getMessage().indexOf("Position:");
```
julianladisch, line 133:
```suggestion
        int pos = e.getMessage().indexOf("Position: ");
```
substring(pos + 10) skips the character after the colon. This character must be a space to avoid matching "Position:20".

cql2pgjson/src/test/java/org/folio/cql2pgjson/CQL2PgJSONTest.java
```diff
@@ -129,8 +128,13 @@ public void select(CQL2PgJSON aCql2pgJson, String sqlFile, String cql, String ex
...
+        if (pos != -1) {
+          int errorPos = Integer.parseInt(e.getMessage().substring(pos + 10)) - 1;
+          sql = sql.substring(0, errorPos) + "****^****" + sql.substring(errorPos);
```
julianladisch, line 136: These two lines should be wrapped into
```
try {
  ....
} catch (Exception e) {
  // ignore
}
```
because parseInt may fail if there is more text after the position number.

hjiebsco (commented):
FYI the work has been pushed back to mid-term backlog at core team grooming.

adamdickmeiss:
Closing as this will probably never be done. I'll keep the branch and comments if this should ever be revisited.

PR #704, opened 2020-06-12
RMB-555: Upgrade creates duplicate foreign key constraints

julianladisch:
No longer create duplicate foreign key constraints on upgrade.

Delete all existing duplicate foreign key constraints that
RMB < 30.0.3 has created.

Replace each case sensitive format() %I placeholder by a
case insensitive form, either %s placeholder or || concatenation,
resulting in identifier downcasing:
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

### DIFF
--- a/domain-models-runtime/src/main/resources/templates/db_scripts/foreign_keys.ftl
+++ b/domain-models-runtime/src/main/resources/templates/db_scripts/foreign_keys.ftl
@@ -52,7 +52,7 @@
       BEFORE INSERT OR UPDATE ON ${myuniversity}_${mymodule}.${table.tableName}
       FOR EACH ROW EXECUTE PROCEDURE ${myuniversity}_${mymodule}.update_${table.tableName}_references();
 
-    -- Remove duplicate foreign key constraints created by RMB before 30.0.3
+    -- Remove duplicate foreign key constraints created by RMB before 30.1.0
     -- https://issues.folio.org/browse/RMB-555
     DO $$
     DECLARE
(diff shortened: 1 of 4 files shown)


domain-models-runtime/src/main/resources/templates/db_scripts/foreign_keys.ftl
```diff
@@ -4,7 +4,21 @@
     <#list table.foreignKeys?filter(key -> key.fieldName??) as key>
       <#if key.tOps.name() == "ADD">
         ALTER TABLE ${myuniversity}_${mymodule}.${table.tableName}
-          ADD COLUMN IF NOT EXISTS ${key.fieldName} UUID REFERENCES ${myuniversity}_${mymodule}.${key.targetTable};
+          ADD COLUMN IF NOT EXISTS ${key.fieldName} UUID;
+        DO $$
+        BEGIN
+          <#-- PostgreSQL doesn't have an "ADD CONSTRAINT IF NOT EXISTS" statement.
```
craigmcnally, line 10: I'm not sure it's any cleaner, but couldn't you also do a DROP CONSTRAINT IF EXISTS, then ADD CONSTRAINT ?

craigmcnally: oh I see now that someone else made that suggestion in the stackoverflow question, so you probably already considered it.

julianladisch: Adding the constraint after it has been dropped requires a check of the constraint condition for all records. This may take long if there are many records, for example in inventory with several million instances, holdings and items that are in foreign key relation.

PR #709, opened 2020-06-18
RMB-650: Client generator support Date query parameter

wwelling:


wwelling:
@julianladisch do you have any ideas how to go about testing this?

julianladisch:
You can extract the code block of `if (AnnotationGrabber.QUERY_PARAM.equals(paramType))` into a method.
You can remove `private` from all methods you want to test. If the unit test class is in the org/folio/rest/tools package it can access the method.

wwelling:
@julianladisch I understand that much. What about assertions? How do we know these methods ran as expected.

wwelling:
I will be trying to write unit tests. Other than that would like feedback as to approach.

julianladisch:
Closed in favour of #711 (code + tests).

PR #710, opened 2020-06-19
Matching all records with =*

julianladisch:


julianladisch:
Yes, code is already there:
* https://github.com/folio-org/raml-module-builder/blob/v30.2.1/cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java#L429-L438
* https://github.com/folio-org/raml-module-builder/blob/v30.2.1/cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java#L606-L610
* https://github.com/folio-org/raml-module-builder/blob/v30.2.1/cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java#L857-L859

PR #711, opened 2020-06-19
RMB-650: Client generator should support Date query parameter

julianladisch:

PR #712, opened 2020-06-19
RMB-650: Initial naive test client is generated as expected

wwelling:
This is not a good of test as needed, but will effectively cover client generator. There should be unit tests, if unit testable, and integration tests in which compile the generated client as test it.


julianladisch:
Closed in favour of #711. We need code + test in same pull request for proper sonarcube coverage calculation.

PR #718, opened 2020-06-23
RMB-654: Use class level path annotation value before generating client method

wwelling:
https://issues.folio.org/browse/RMB-654

This fixes generated method requests to use class level path annotation value and will correctly concatenate query parameters.


wwelling:
https://github.com/folio-org/raml-module-builder/blob/master/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java#L323

With this change, URL parameter generateMethodMeta will not be null and never should be.

PR #719, opened 2020-06-24
RMB-650: Simplify query parameter date time format

wwelling:
remove second fration and zone offset

### DIFF
--- a/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
+++ b/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
@@ -430,8 +430,7 @@ public class ClientGenerator {
 
   private void formatDateParameter(JBlock b, ParameterDetails details) {
     JExpression expr = jcodeModel.ref(java.time.format.DateTimeFormatter.class)
-      .staticInvoke("ofPattern")
-        .arg("yyyy-MM-dd'T'HH:mm:ss")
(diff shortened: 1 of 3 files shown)


domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -431,7 +431,7 @@ private void encodeParameter(JBlock b, ParameterDetails details) {
   private void formatDateParameter(JBlock b, ParameterDetails details) {
     JExpression expr = jcodeModel.ref(java.time.format.DateTimeFormatter.class)
       .staticInvoke("ofPattern")
-        .arg("yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
+        .arg("yyyy-MM-dd'T'HH:mm:ss")
```
julianladisch, line 434: Can you use DateTimeFormatter.ISO_LOCAL_DATE_TIME?
Can you add a unit test to DemoRamlRestTest that has both date and time?

wwelling: Sure and maybe.

wwelling: Done. The additional test will have to suffice. The implementation, BooksDemoAPI,  does not provide any use of the datetime parameters to make any useful assertions.

PR #720, opened 2020-06-27
RMB-656: NPE at startup when an OPTIONS endpoint is defined in RAML

julianladisch:
Add org.folio.rest.jaxrs.resource.support.OPTIONS as possible HTTP
method.

Move RTFConsts.POSSIBLE_HTTP_METHOD to
AnnotationGrabber.isPossibleHttpMethod and replace substring
matching by complete string matching.

Add options API endpoint to sample.raml and a rest-assured test
for it to DemoRamlRestTest.

### DIFF
--- a/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/AnnotationGrabber.java
+++ b/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/AnnotationGrabber.java
@@ -141,8 +141,7 @@ public class AnnotationGrabber {
           for (int j = 0; j < methodAn.length; j++) {
 
             Class<? extends Annotation> type = methodAn[j].annotationType();
-            //System.out.println("Values of " + type.getName());
-            if (RTFConsts.POSSIBLE_HTTP_METHOD.contains(type.getName())) {
+            if (isPossibleHttpMethod(type.getName())) {
               // put the method - get or post, etc..
               methodObj.put(HTTP_METHOD, type.getName());
             }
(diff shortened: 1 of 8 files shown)


domain-models-interface-extensions/src/main/java/org/folio/rest/tools/AnnotationGrabber.java
```diff
@@ -289,6 +288,24 @@ public static JsonObject getParameterNames(Method method) throws Exception {
...
+    switch (method) {
+    case "javax.ws.rs.PUT":
+    case "javax.ws.rs.POST":
+    case "javax.ws.rs.DELETE":
+    case "javax.ws.rs.GET":
+    case "javax.ws.rs.OPTIONS":
+    case "javax.ws.rs.HEAD":
+    case "javax.ws.rs.TRACE":
+    case "javax.ws.rs.CONNECT":
+    case "org.folio.rest.jaxrs.resource.support.OPTIONS":
```
craigmcnally, line 301: I'm not familiar enough with this code to know if this was done intentionally or not, but OPTIONS is here twice...

Is this needed, or a mistake?

julianladisch: The new one is needed.
I don't know whether the existing one from javax.ws.rs is need, might be in use in older versions. To avoid a breaking change we keep it.

craigmcnally:
@julianladisch can we please get this merged?  If I'm going to be working from a snapshot jar I'd rather it be from the master branch than a dev branch.

PR #728, opened 2020-07-06
RMB-670: Replace log4j 1.2 MDC by log4j 2 ThreadContext

julianladisch:
log4j 1.2 and its MDC does not work in Java >= 9:
https://blogs.apache.org/logging/entry/moving_on_to_log4j_2

Wrap each RestVerticle logging statement into withRequestId to
temporarly add an reqId entry into ThreadContext. This will
add the request id to all RestVerticle log lines and allows to
calculate some basic metrics. Refinement will be done in
https://issues.folio.org/browse/RMB-669

The old code has been using this pattern in log4j2.properties:
`%d{dd MMM yyyy HH:mm:ss:SSS} %-5p %C{1} [%reqId] %m%n`

`%r` is number of milliseconds elapsed since the JVM was started
until the creation of the logging event.
https://logging.apache.org/log4j/2.x/manual/layouts.html#Patterns
This results in `[7265eqId]` and clearly is a typo.

The new pattern will be
`%d{dd MMM yyyy HH:mm:ss:SSS} %-5p %C{1} [%X{reqId}] %m%n`
resulting in `[reqId=981376281]` if reqId is defined or `[]` otherwise.


craigmcnally:
> What is your opinion, should I completely remove `ThreadContext.put("reqId", "reqId="+reqId);` because one Thread is used by multiple API request resulting in wrong reqId if the first API call has an async call after a second API has started another async call on the same thread?

If it's going to cause confusion - log the wrong reqId in some cases then it should be removed.

craigmcnally:
Adding a comment to the README/release notes would be a good idea too... it's possible that folks have some log parsing in place that will break if you rip out a field.  

One way to avoid this would be to always put some static placeholder there and keep the field in place.  Either way it should be documented.

julianladisch:
This solution using withRequestId allows logging with requestId for Goldenrod using Vert.x 3.9.
For Honeysuckle we will revert the requestId, update to Vert.x 4 and use a solution using vertx context local values.

PR #729, opened 2020-07-07
RMB-671: Reinstall PostgreSQL extensions on upgrade, drop pg_catalog.pg_trgm

julianladisch:
As an SysOp I want to run the upgrade on a system that is based on a freshly
installed backup. The backup may miss the PostgreSQL extensions that
reside in public schema.
Example: https://issues.folio.org/browse/MODORDSTOR-161

To make the life of SysOps easier be forgiving and reinstall all PostgreSQL
extensions on upgrade.

pg_trgm may have incorrectly been created in pg_catalog schema, see bug report in
https://issues.folio.org/browse/MODORDSTOR-161
Drop pg_trgm if it is in a wrong schema.
Note: A PostgreSQL extension can be loaded into one schema only.


julianladisch:
Sure, removed.

PR #731, opened 2020-07-09
RMB-673: totalRecords exact hit count when < 1000, fixing limit=0

julianladisch:
limit=0 is used to get the totalRecords number without fetching records.
This is useless if estimated count is 8924 and actual count is 0, see
https://issues.folio.org/browse/MODORDERS-407 "Cannot create POL due to POL limit bug".

Solution: Always use `count(*) ... LIMIT 1000` to get the actual count
if the actual count is 1000. Only if >= 1000 return an estimation.

Background: We have removed this in
https://issues.folio.org/browse/RMB-591 "remove or harmonize 'optimizedSql' execution path"
for performance reasons. However,
https://issues.folio.org/browse/RMB-645 'use where-only clause for the "count query"'
removed the sortBy/order by clause from the count query reducing the execution time
of the count query.

There can still be queries where calculating the count can take long, for example
`title=int* and title=and* and title=con* and title=with*` takes 700 ms for the
performance dataset with 2.7 million records. Similar queries will take 2800 ms
for a dataset with 10 million records. And it takes even longer when the query
uses fields from instance, holding and item.

These performance issues may be addressed by changing the default `exactCount=1000`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1773,7 +1773,15 @@ http://localhost:<port>/configurations/entries?query=scope.institution_id=aaa%20
 
 RMB adds a `totalRecords` field to result sets. It contains an estimation how many records were matching if paging parameters were set to `offset` = 0 and `limit` = unlimited.
 
-It uses this algorithm:
+It uses this algorithm for RMB >= 30.2.4 and for RMB < 30.0.0:
+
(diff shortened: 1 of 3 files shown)


julianladisch:
We use the other solution: #732 "RMB-673: Return exact totalRecords value for limit=0"

PR #732, opened 2020-07-09
RMB-673: Return exact totalRecords value for limit=0

julianladisch:


julianladisch:
When limit=0 then getWithOptimizedSql is not called because nothing can be optimized if no records are returned:
https://github.com/folio-org/raml-module-builder/blob/v30.2.3/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L1235

PR #739, opened 2020-07-31
RMB-687: ResponseException for TenantAPI

julianladisch:
Use a ResponseException that contains a javax.ws.rs.core.Response
when a TenantAPI call fails.

The TenantAPI methods use a Handler<AsyncResult<Response>> handler
to return success and failure. It is returned as a succeeded
AsyncResult where the value is the Response:
https://github.com/folio-org/raml-module-builder/blob/v30.2.4/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java

On success the Response has a 200 or 201 HTTP status code.
If installation or upgrade fails the Response has a 400 or 500
HTTP status code and a detailed error message in the body.
Note that AsyncResult is a succeeded result in all cases.

The documentation was wrong in that it only checked the
AsyncResult (succeeded, failed) and not the Response status
code.

This is how we fix this:
A success is returned using a succeeded AsyncResult as before.
A failure is returned using a failed AsyncResult. A failed AsyncResult
doesn't have a value where we can pass the Response. But it has an
Exception and we can create our custom ResponseException that contains
the Response.

RestVerticle extracts the Response from the ResponseException.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
@@ -51,6 +51,7 @@ import org.folio.rest.tools.AnnotationGrabber;
 import org.folio.rest.tools.ClientGenerator;
 import org.folio.rest.tools.PomReader;
 import org.folio.rest.tools.RTFConsts;
+import org.folio.rest.tools.client.exceptions.ResponseException;
 import org.folio.rest.tools.client.test.HttpClientMock2;
 import org.folio.rest.tools.codecs.PojoEventBusCodec;
 import org.folio.rest.tools.messages.MessageConsts;
@@ -769,7 +770,13 @@ public class RestVerticle extends AbstractVerticle {
    *          - request's start time, using JVM's high-resolution time source, in nanoseconds
    */
   private void sendResponse(RoutingContext rc, AsyncResult<Response> v, long start, String tenantId) {
-    Response result = ((Response) ((AsyncResult<?>) v).result());
+    Response result = v.result();
+    if (v.failed()) {
+      Throwable exception = v.cause();
+      if (exception instanceof ResponseException) {
+        result = ((ResponseException) exception).getResponse();
+      }
+    }
     if (result == null) {
       // catch all
       endRequestWithError(rc, 500, true, "Server error", new boolean[] { true });
--- a/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
@@ -759,6 +759,17 @@ public class RestVerticle extends AbstractVerticle {
     return false;
   }
 
+  static Response getResponse(AsyncResult<Response> asyncResult) {
+    if (asyncResult.succeeded()) {
+      return asyncResult.result();
+    }
+    Throwable exception = asyncResult.cause();
+    if (exception instanceof ResponseException) {
+      return ((ResponseException) exception).getResponse();
+    }
+    return null;
+  }
+
   /**
    * Send the result as response.
    *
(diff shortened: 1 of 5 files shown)


domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -770,13 +781,7 @@ private boolean isStreamed(Annotation[] annotations) {
...
    */
   private void sendResponse(RoutingContext rc, AsyncResult<Response> v, long start, String tenantId) {
-    Response result = v.result();
-    if (v.failed()) {
-      Throwable exception = v.cause();
-      if (exception instanceof ResponseException) {
-        result = ((ResponseException) exception).getResponse();
-      }
-    }
+    Response result = getResponse(v);
```
marcjohnson-kint, line 784:
```suggestion
    Response response = getResponse(v);
```

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -759,6 +759,17 @@ private boolean isStreamed(Annotation[] annotations) {
     return false;
   }
 
+  static Response getResponse(AsyncResult<Response> asyncResult) {
```
marcjohnson-kint, line 762: How does this function relate to the result from the `postTenant` method, would this method be used to interpret it?

julianladisch: Yes, I've added javadoc explaining this. This method is package-private.

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
julianladisch: This name is already used for `HttpServerResponse response` a few lines below.

marcjohnson-kint: Then maybe it could be called something like `responseFromResult`?

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -319,14 +321,23 @@ public void postTenant(TenantAttributes tenantAttributes, Map<String, String> he
...
+      handlers.handle(Future.failedFuture(new ResponseException(response)));
     })
-    .onSuccess(response -> handlers.handle(Future.succeededFuture(response)));
+    .onSuccess(response -> {
+      if (response.getStatus() >= 300) {
+        handlers.handle(Future.failedFuture(new ResponseException(response)));
+        return;
+      }
+      handlers.handle(Future.succeededFuture(response));
+    });
```
marcjohnson-kint, line 341: Does this mean that all clients will need to change their failure handling?

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
julianladisch: Yes, done.

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
julianladisch: No, clients that have a failure can pass the 400-/500-Response either as a succeeding AsyncResult or wrapped in ResponseException as a failing AsyncResult. RestVerticle#getResponse extracts it from all possibilities.

julianladisch:
@hjiebsco Good catch, I've changed all other 400s and 500s.

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
julianladisch: It is used in one place only, in the first line of sendResponse:
`Response responseFromResult = getResponse(v);`

craigmcnally:
I think these changes make sense.  Thanks for the explanation in the description

PR #740, opened 2020-08-04
RMB-682 Resolve issues with not working ClassPath util class from Guava lib

dmtkachenko:
## Purpose
`ClassPath` utility from Guava library used by RMB stops working under JDK 11: returns empty set of classes/interface for a given package although the classes are present. This results in errors during API endpoint classes generation and request processing by `RestVerticle`. Since an update to the latest version of Guava is not possible due to its incompatibility with JAXRS code generator, latest version of `ClassPath` has been taken and adopted to work w/o any other Guava related classes.

## Approach
- place the latest version of 'ClassPath' into `domain-models-interface-extensions` and make its code independent from other Guava classes;
- replace usage of Guava's version of `ClassPath` to localized (RMB) version in `AnnotationGrabber`, `PomReader`, `InterfaceToImpl`
- remove explicit dependency to Guava from RMB modules. `jaxrs-code-generator` still depends on Guava v.20.0, so this dependency remains.

#### TODOS and Open Questions
- [ ] to completely get rid of Guava usage in RMB code, several classes should be updated. Those classes mostly use different guava collection classes that can be replaced with either Java standard classes or Apache Commons Collection classes.

### Note
PR can only be **built with JDK 11**


adamdickmeiss:
Close because the work is moved to RMB-328-openjdk-11 branch

PR #741, opened 2020-08-13
RMB-693: Close prepared statements in PostgresClient stream get

julianladisch:
Closing the prepared statement releases the resources allocated
by PostgreSQL for them and also releases the name of the prepared
statement for reuse. The latter will reduce the likelihood
of failures caused by duplicate prepared statement names reported
in https://issues.folio.org/browse/MODINVSTOR-540 :
"ERROR: prepared statement "XYZ" already exists"

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -1873,10 +1873,11 @@ public class PostgresClient {
         replyHandler.handle(Future.failedFuture(prepareRes.cause()));
         return;
       }
-      PreparedStatement pq = prepareRes.result();
-      RowStream<Row> stream = pq.createStream(STREAM_GET_DEFAULT_CHUNK_SIZE, Tuple.tuple());
-      PostgresClientStreamResult<T> streamResult = new PostgresClientStreamResult(resultInfo);
-      doStreamRowResults(stream, clazz, closeConnection, queryHelper,
+      PreparedStatement preparedStatement = prepareRes.result();
+      RowStream<Row> rowStream = new PreparedRowStream(
+          preparedStatement, STREAM_GET_DEFAULT_CHUNK_SIZE, Tuple.tuple());
+      PostgresClientStreamResult<T> streamResult = new PostgresClientStreamResult<>(resultInfo);
+      doStreamRowResults(rowStream, clazz, closeConnection, queryHelper,
           streamResult, replyHandler);
     });
   }
@@ -1896,10 +1897,10 @@ public class PostgresClient {
     }
   }
 
-  <T> void doStreamRowResults(RowStream<Row> sqlRowStream, Class<T> clazz,
-    PgConnection pgConnection, QueryHelper queryHelper,
-    PostgresClientStreamResult<T> streamResult,
-    Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
+  <T> void doStreamRowResults(RowStream<Row> rowStream, Class<T> clazz,
+      PgConnection pgConnection, QueryHelper queryHelper,
+      PostgresClientStreamResult<T> streamResult,
+      Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
 
     ResultInfo resultInfo = streamResult.resultInto();
     Promise<PostgresClientStreamResult<T>> promise = Promise.promise();
@@ -1907,7 +1908,7 @@ public class PostgresClient {
     boolean isAuditFlavored = isAuditFlavored(resultsHelper.clazz);
     Map<String, Method> externalColumnSetters = new HashMap<>();
     AtomicInteger resultCount = new AtomicInteger();
-    sqlRowStream.handler(r -> {
+    rowStream.handler(r -> {
       try {
         // for first row, get column names
         if (resultsHelper.offset == 0) {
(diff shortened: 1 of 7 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1873,10 +1873,11 @@ public TotaledResults(RowSet<Row> set, Integer total) {
         replyHandler.handle(Future.failedFuture(prepareRes.cause()));
         return;
       }
-      PreparedStatement pq = prepareRes.result();
-      RowStream<Row> stream = pq.createStream(STREAM_GET_DEFAULT_CHUNK_SIZE, Tuple.tuple());
```
hjiebsco, line 1877: Would it work if adding `stream.endHandler/exceptionHandler` to close `pq` here?

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public RowStream<Row> endHandler(Handler<Void> endHandler) {
+    rowStream.endHandler(endHandler);
+    return this;
+  }
+
+  @Override
+  public void close() {
+    preparedStatement.close(close -> rowStream.close());
```
hjiebsco, line 79: Since rowStream is created from PreparedStatement, should we close rowStream first, and then PreparedStatment?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2840,14 +2840,14 @@ public void streamGetExceptionInHandler(TestContext context) throws IOException,
         });
         sr.endHandler(x -> {
           events.append("[endHandler]");
+          async.complete();
         });
         sr.exceptionHandler(x -> {
           events.append("[exception]");
```
hjiebsco, line 2846: Maybe the exceptionHandler can be moved above endHandler, so the code order aligns with the testing logic better?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: No, doStreamRowResults will overwrite endHandler and exceptionHandler.

julianladisch:
> have you tried using rowStream.endHandler/exceptionHandler to take care of closing the opened PreparedStatement? If works, it would be much simpler.

The caller of

    public void selectStream(AsyncResult<SQLConnection> conn, String sql,
        Handler<AsyncResult<RowStream<Row>>> replyHandler)

may want to cancel and close the query before the it ends, for example if some other parallel running query or task fails.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
julianladisch: Reordered.

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
julianladisch: Yes, thanks, good catch!

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public void close(Handler<AsyncResult<Void>> completionHandler) {
+    rowStream.close(close1 -> preparedStatement.close(close2 -> {
+      if (close1.failed()) {
+        completionHandler.handle(close1);
+        return;
+      }
+      completionHandler.handle(close2);
+    }));
```
hjiebsco, line 90: If rowStream failed to close, we still need to try to close preparedStatement.

domain-models-runtime/src/test/java/org/folio/rest/persist/PreparedRowStreamTest.java
```diff
@@ -0,0 +1,76 @@
...
+    doAnswer(AdditionalAnswers.answerVoid((Handler<AsyncResult<Void>> handler)
+        -> handler.handle(Future.failedFuture("fail1"))))
+        .when(mockedPreparedStatement).close(any());
+    doAnswer(AdditionalAnswers.answerVoid((Handler<AsyncResult<Void>> handler)
+        -> handler.handle(Future.failedFuture("fail2"))))
+        .when(mockedRowStream).close(any());
+    PreparedRowStream preparedRowStream = new PreparedRowStream(mockedPreparedStatement, mockedRowStream);
+    StringBuilder s = new StringBuilder();
+    preparedRowStream.close(handler -> s.append(handler.cause().getMessage()));
+    assertThat(s.toString(), is("fail1"));
```
hjiebsco, line 73: Currently this assertion fails the test/build because of the new close order (rowStream first, then preparedStatement).

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
julianladisch: preparedStatement.close is called even if close1 failed.

domain-models-runtime/src/test/java/org/folio/rest/persist/PreparedRowStreamTest.java
julianladisch: Fixed.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2840,14 +2840,14 @@ public void streamGetExceptionInHandler(TestContext context) throws IOException,
...
+          vertx.runOnContext(run -> async.complete());
         });
         sr.endHandler(x -> {
           events.append("[endHandler]");
-          async.complete();
         });
       }));
     async.await(1000);
-    context.assertEquals("[handler][exception][endHandler]", events.toString());
+    context.assertEquals("[handler][exception]", events.toString());
```
hjiebsco: Good catch. Is async.await(1000) still needed then?

julianladisch: Yes, it is still needed, because we need to wait until async.complete() has been called before we can run the assertEquals checks. Otherwise event will be empty.

PR #747, opened 2020-08-19
RMB-684 no records when offset value > estimated totalRecords

julianladisch:
Fix bug that has been introduced in v22.0.0:
https://github.com/folio-org/raml-module-builder/commit/84d969dd4d9bf20bf9360090738960652a625132

Rename total to estimatedTotal.


adamdickmeiss:
I don't think the error occurred before estimated hit counts started to occur. So at least after 29.1.0.

adamdickmeiss (approved):

julianladisch:
Using count_estimate_smart for calculating totalRecords has been introduced in v16.0.0: https://github.com/folio-org/raml-module-builder/commit/c8ddce643202795506d44bc5407c81b5ae48f24d

PR #749, opened 2020-08-20
Use JarFile to sweep through Jar RMB-699

adamdickmeiss:
So we get things in correct order.

### DIFF
--- a/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
+++ b/domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
@@ -1,19 +1,21 @@
 package org.folio.rest.tools;
 
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
 import java.io.File;
 import java.io.FileReader;
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Enumeration;
 import java.util.List;
 import java.util.Properties;
-import java.util.Set;
-
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
-import org.folio.rest.tools.utils.ClassPath;
-
-import io.vertx.core.logging.Logger;
-import io.vertx.core.logging.LoggerFactory;
-
 
 /**
  * @author shale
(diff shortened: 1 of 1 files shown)


adamdickmeiss:
TODO: PomReaderTest and clean up the class which has a weird style.

adamdickmeiss:
0% code coverage.. New bottom!

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
```diff
@@ -34,81 +38,102 @@
   @SuppressWarnings("checkstyle:methodlength")
```
julianladisch, line 38: This has been fixed and can be removed.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
```diff
@@ -34,81 +38,102 @@
...
-        version = model.getParent().getVersion();
-      }
-      else{
-        moduleName = model.getArtifactId();
-      }
+      boolean readCurrent = currentRunningJar != null && (currentRunningJar.contains("domain-models-runtime")
+          || currentRunningJar.contains("domain-models-interface-extensions"));
+      readIt(readCurrent);
+    } catch (Exception e) {
+      log.error(e.getMessage(), e);
```
julianladisch, line 47: The exception should be rethrown as a RuntimeException to fail the deployment if we cannot read the module name.

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Test
+  void readFromJar() throws IOException, XmlPullParserException {
+    PomReader pom = PomReader.INSTANCE;
+
+    pom.readIt(false);
+    // first dependency in main pom / but surefire sometimes?
+    assertThat(pom.getModuleName(), anyOf(is("vertx_parent"), is("surefire")));
+    pom.readIt(true);
```
julianladisch, line 54: Can pom.readIt(true) been removed?

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+class PomReaderTest {
+  static {
+    System.setProperty(LoggerFactory.LOGGER_DELEGATE_FACTORY_CLASS_NAME, "io.vertx.core.logging.Log4j2LogDelegateFactory");
+  }
+
+  @Test
+  void testGetModuleName() {
+    assertThat(PomReader.INSTANCE.getModuleName(), is("raml_module_builder"));
+  }
```
julianladisch, line 24: Can you add a unit test *Test.java (reading from file system) and an integration test *IT.java (reading from jar) for domain-models-runtime-it that checks that getModuleName() returns domain-models-runtime-it?

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
adamdickmeiss: No. It's a singleton. So the values stored there will be incorrect.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
adamdickmeiss: Yes. Agreed. Hesitated because that will change behavior.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
```diff
@@ -34,81 +38,102 @@
...
+        while (entries.hasMoreElements()) {
+          JarEntry entry = entries.nextElement();
+          String name = entry.getName();
+          // first pom.xml should be the right one.
+          if (name.startsWith(dirname) && !dirname.equals(name) && name.endsWith("pom.xml")) {
+            InputStream pomFile = PomReader.class.getClassLoader().getResourceAsStream(name);
+            model = mavenreader.read(pomFile);
+            break;
+          }
+        }
```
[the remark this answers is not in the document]
adamdickmeiss: yes

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
julianladisch: Can you add a comment that this is to restore the values for the other unit tests?

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
adamdickmeiss: sure

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
adamdickmeiss: Sure

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/PomReader.java
adamdickmeiss: For now, re-throw a Runtime exception. I don't know how to otherwise do it with this enum. If you think it's a big deal I think we should make a follow up issue. I'd like to get fix out.

PR #752, opened 2020-08-21
RMB-677: Close PostgreSQL connection after invalid CQL failure

julianladisch:
PostgresClient.streamGet should close the PostgreSQL connection.
Otherwise an invalid CQL will not release the connection and
further requests will wait forever to get an available connection.


adamdickmeiss (approved):
OMG. Thx

PR #753, opened 2020-08-24
RMB-702: Upgrading note: Rename {version} path variable in RAML files

julianladisch:


alb3rtino (approved):

adamdickmeiss (changes requested):
This is a general advice, not an advice for an upgrade. If it is, we have introduced a regression (and a bug) and then it should be a major change.

adamdickmeiss (approved):
ok. so something happened in 30.2 that caused this change.

PR #754, opened 2020-08-24
RMB-553, RMB-698, RMB-707: PgUtil.postSync with metadata

julianladisch:
Copy mod-inventory-storage StorageHelper.postSync to PgUtil.postSync for
reuse by other modules.

Move metadata code from RestVerticle into MetadataUtil for reuse.

Set metadata in PgUtil.postSync.

See RMB-535 why we want each API endpoint to define a maximum number
of posted entities.


adamdickmeiss (commented):
Looks like it's the same logic for decoding tokens as in okapi-common. Why repeat it here?

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/JwtUtils.java
```diff
@@ -1,17 +1,48 @@
...
+   * Get the value for key from the payload. It neither validates the
+   * token header nor the token signature.
+   * @param token  a JWT token
+   * @return the value for key as String, or null if the key does not exist or token cannot be decoded
+   */
+  public static String get(String key, String token) {
+    try {
+      if (token == null) {
+        return null;
+      }
```
hjiebsco, line 33: null check does not need to be in the try block

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/MetadataUtil.java
```diff
@@ -0,0 +1,98 @@
...
+    // we overwrite them here, the correct value will be reset via a database trigger on update.
+    if (entity == null) {
+      return;
+    }
+    Method setMetadata = getSetMetadataMethod(entity);
+    if (setMetadata == null) {
+      return;
+    }
+    setMetadata.invoke(entity, createMetadata(okapiHeaders));
+  }
```
hjiebsco, line 69: There are some duplicated logic between this method and Line 75 `public static <T> void populateMetadata(List<T> entities, Map<String, String> okapiHeaders)`. Maybe consider refactoring for one to call the other.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/JwtUtils.java
julianladisch: JwtUtils.java has been deleted.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/MetadataUtil.java
julianladisch: Good catch, changed.

julianladisch:
@adamdickmeiss Good catch, I've removed JwtUtils in favour of okapi-common OkapiToken. I've created RMB-707 to track this (very small) breaking change.

PR #755, opened 2020-08-25
FOLREL-422: Release 30.2.6

julianladisch:

PR #756, opened 2020-08-25
RMB-704: RestVerticle "singleField" is nullable

julianladisch:
Sonar fails the first pull request against a new branch
because of two bugs: https://github.com/folio-org/raml-module-builder/pull/755

This issue fixes the one in RestVerticle:

A "NullPointerException" could be thrown; "singleField" is nullable here.


julianladisch:
Thanks, good catch. To prevent this to happen again I've extended the release procedure: https://github.com/folio-org/folio-org.github.io/pull/619/files

PR #759, opened 2020-08-26
RMB-703: Full text seach doesn't match URLs containing &

julianladisch:
PostgreSQL's full text parser recognizes URLs:
https://www.postgresql.org/docs/current/textsearch-parsers.html
An URL may contain &, for example
```
SELECT to_tsvector('simple', 'http://example.com/foo?bar=1&baz=2');
                             to_tsvector
----------------------------------------------------------------------
 '/foo?bar=1&baz=2':3 'example.com':2 'example.com/foo?bar=1&baz=2':1
```

to_tsvector removes all punctuation. URLs are an exception.
RMB's tsquery_phrase, tsquery_and and tsquery_or functions
replace & by , (comma) because & is the and operator.
```
SELECT tsquery_phrase('http://example.com/foo?bar=1&baz=2');
                             tsquery_phrase
------------------------------------------------------------------------
 'example.com/foo?bar=1,baz=2' <-> 'example.com' <-> '/foo?bar=1,baz=2'
```

For details see cql2tsqueryAnd, cql2tsqueryOr and cql2tsqueryPhrase in
https://github.com/folio-org/raml-module-builder/blob/master/cql2pgjson/src/main/java/org/folio/cql2pgjson/util/Cql2SqlUtil.java

Fix:
RMB needs to replace each & by , before passing the string to ts_vector.
Replace each to_tsvector('simple', ...) by get_tsvector(...) and define
get_tsvector($1)=to_tsvector('simple', translate($1, '&', ',')).


skoczko:
@julianladisch can we use some more descriptive name rather than "my_tsvector"?

julianladisch:
In what direction should the name go? It should be similar to our tsquery function names: tsquery_and, tsquery_or, tsquery_phrase.
https://github.com/folio-org/raml-module-builder/blob/f6267e35f6b09daf472cad10dbd1593f4e4268ed/domain-models-runtime/src/main/resources/templates/db_scripts/general_functions.ftl#L110-L137

julianladisch:
@skoczko How about get_tsvector instead of my_tsvector?

adamdickmeiss:
& maps to , .. And , maps to , I presume. Does that mean that both http://myhost?a and http://myhost,a will hit the same result?

adamdickmeiss:
The bug is marked as Q2 bug fix. Are customers really willing to reindex most things? Is this error so severe it requires a Q2 fix.

julianladisch:
Within URLs all punctuation must match but ampersand and comma are considered the same character. Outside URLs all punctuation is the same as whitespace.

"foo bar" and "foo,bar" and "foo & bar" are the same and match each other.

`http://example.com/foo?bar=1&baz=2` and `http://example.com/foo?bar=1,baz=2` match each other.

`http://example.com/foo?bar=1&baz=2` and `http://example.com/foo,bar=&baz=2` do NOT match because ? and , are considered different.

PostgreSQL requires a dot in the hostname to recognize it as URL. `http://myhost?a` and `http://myhost,a` and `http://myhost&a` and `http myhost a` are considered equal. `http://localhost/`  is the same as `http localhost`.

julianladisch:
https://issues.folio.org/browse/MODDICORE-80 has high customer priority for Lehigh and Chicago.
RMB-703 is a bug that we need to fix sooner or later and this requires reindexing.
I don't see any alternative solutions.

adamdickmeiss (approved):
Thx for answers.

PR #772, opened 2020-09-03
RMB-721: NPE in PomReader while PostgresClient.getInstance

julianladisch:
PomReader cannot read parent pom's dependencyManagement when an RMB based module resides in a maven submodule like [folio-service-tools/folio-service-tools-dev/pom.xml](https://github.com/folio-org/folio-service-tools/blob/master/folio-service-tools-dev/pom.xml).
Reading the parent's pom is needed if the submodule pom.xml doesn't contain a raml-module-builder (RMB) version but need to fetch it from the patent's dependencyManagement entry.
In addition maven has a complicated dependency resolution algorithm: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html
We need to implement that algorithm if we want to fetch the RMB version from the poms like folio-service-tools/folio-service-tools-dev/pom.xml.

Solution: Let maven write the version into a RmbVersion.java when building RMB before compile phase. Then the version is always available, in all unit test and integration test phases, both in RMB and in RMB based modules.


domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
```diff
@@ -88,6 +88,6 @@ void otherPom() throws IOException, XmlPullParserException {
     pom.init("src/test/resources/pom/pom-sample.xml");
     assertThat(PomReader.INSTANCE.getModuleName(), is("mod_inventory_storage"));
     assertThat(PomReader.INSTANCE.getVersion(), is("19.4.0"));
-    assertThat(PomReader.INSTANCE.getRmbVersion(), is("30.0.0"));
+    assertThat(PomReader.INSTANCE.getRmbVersion(), is("19.4.0"));
```
hjiebsco, line 91: Is `19.4.0` expected result? The pom-sample.xml actually has rmb version `30.0.0`

hjiebsco (commented):
Should the new logic be combined with the previous implementation, for example, when parent version is null then use the new logic?

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/PomReaderTest.java
julianladisch: Thanks for pointing out this discrepancy.
Now rmbVersion is always provided by maven in RmbVersion and never read from and pom.xml or properties files.

adamdickmeiss (approved):
Nice. Just beware that the same exception was thrown in earlier versions of RMB. Logged, but ignored.

hjiebsco (commented):
Strictly speaking `RMB_VERSION` means version of `domain-models-runtime`, not `domain-models-interface-extensions`. How about remove RmbVersion detection from `POMReader`, and implement it in `domain-models-runtime` in the simplest way? It would be cleaner in my view.

julianladisch:
PomReader.getRmbVersion() is used by domain-models-runtime only. No RMB based module is affected. Therefore I made this suggested change.

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -81,7 +81,8 @@ public void deleteTenant(Map<String, String> headers,
 
             String sqlFile = null;
             try {
-              SchemaMaker sMaker = new SchemaMaker(tenantId, PostgresClient.getModuleName(), TenantOperation.DELETE, null, PomReader.INSTANCE.getRmbVersion());
+              SchemaMaker sMaker = new SchemaMaker(tenantId, PostgresClient.getModuleName(),
+                  TenantOperation.DELETE, null, RmbVersion.getRmbVersion());
```
hjiebsco, line 85: Not sure why set module previous version to null and module module version to RMB version here.

hjiebsco (approved):
Left a question about the old code logic.

adamdickmeiss (commented):
While RmbVersion.getRmbVersion() need not be public it is executed by RMB when linked with another module at run-time. So we still need to save it or fix the analysis of the POM anyway.

julianladisch:
> While RmbVersion.getRmbVersion() need not be public it is executed by RMB when linked with another module at run-time. So we still need to save it or fix the analysis of the POM anyway.

The RMB version is detected when building RMB.

It is saved in the source code of getRmbVersion() when building RMB. The version is hardcoded and then compiled and shipped with RMB in the RmbVersion class file.

In contrast the module version (mod-inventory-storage version, or folio-service-tools-dev version) need to be detected when building that module. But it is much easier to detect that from the module's pom than detecting the RMB version from the module's pom.

> Should the new logic be combined with the previous implementation, for example, when parent version is null then use the new logic?

No, this is too complicated for the reasons explained in this pull request's description.

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
julianladisch: Good catch. We neither use nor need any version on delete. I've replaced the RMB version by null. For code readability we should not pass a value for a parameter that is not used.

PR #773, opened 2020-09-07
Removed transaction created during stream get

wwelling:


skoczko:
@julianladisch I am a little confused by this, I thought the JDBC driver was used only for migrations?

wwelling:
@skoczko the driver change had no effect. Don't see an issue with it being upgraded though. Will remove from this PR.

julianladisch:
Is there a reproducer - a test that fails with the old code and succeeds with the new code?

wwelling:
No, there is not a test that reproduces the problem. This issue only began occurring for Q2 deployment attempts. 

https://issues.folio.org/browse/MODUSERS-213

The existing tests do still pass. Is there a reason for this method always starting a transaction?

wwelling:
https://vertx.io/docs/vertx-pg-client/java/#_cursors_and_streaming

// Cursors require to run within a transaction
    Transaction tx = connection.begin();

I am going to attempt to write a test to cause the same failure as we have when deploying Q2.

wwelling:
Closing for further investigation.

PR #774, opened 2020-09-09
RMB-723: Close database connection after dropping schema and role

julianladisch:
After dropping schema and role the database connection still uses
the old oid. This will fail when schema and role are recreated
with the same name but a different oid.

Solution:

Close all database connections when dropping schema and role.

Other changes:
* Fix error logging (switch from System.out to log.error).
* Use schema name when calling count_estimate; this results in
  a better error message: "permission problem" instead of "not found"
  when permissions are missing.
* Remove "REVOKE ALL PRIVILEGES ON DATABASE postgres". This is
  not needed. Even if it were needed the hard-coded database
  name "postgres" may be wrong, can be "folio".
* Use "DROP ROLE" (= SQL standard) and not the PostgreSQL synonym
  "DROP USER"; this parallels "CREATE ROLE".

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
@@ -98,7 +98,7 @@ public class TenantAPI implements Tenant {
                   try {
                     // close clients because they still use the old oid of the dropped schema/role name,
                     // they will fail when the same schema/role is recreated with a new oid.
-                    PostgresClient.closeAllClients();
+                    PostgresClient.closeAllClients(tenantId);
                     String res = "";
                     if(reply.succeeded()){
                       res = new JsonArray(reply.result()).encodePrettily();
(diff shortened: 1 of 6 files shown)


hjiebsco (approved):

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -96,12 +96,15 @@ public void deleteTenant(Map<String, String> headers,
             postgresClient(context).runSQLFile(sqlFile, true,
                 reply -> {
                   try {
+                    // close clients because they still use the old oid of the dropped schema/role name,
+                    // they will fail when the same schema/role is recreated with a new oid.
+                    PostgresClient.closeAllClients();
```
adamdickmeiss, line 101: Why close connections for all tenants? Won't the current tenant suffice?

julianladisch: Good catch! Fixed.

adamdickmeiss (approved):

PR #775, opened 2020-09-10
RMB-720: Switch DropWizard to Micrometer InfluxDB

hjiebsco:
See https://issues.folio.org/browse/RMB-720. Basically removed DropWizard dependency and related old monitoring code. Added Micrometer InfluxDB dependency similar to what was done in Okapi.


adamdickmeiss (approved):
Don't see any problems.

julianladisch (approved):

PR #776, opened 2020-09-11
RMB-725: StringUtil.cqlEncode masking CQL characters

julianladisch:
The characters * ? ^ \ " have special meaning when used in a
CQL string constant. CQL injection is possible if not properly
masked/encoded. This may be result is severe security vulnerabilities.
Write StringUtil.cqlEncode for reuse by all modules that use RMB.

References:
https://dev.folio.org/faqs/explain-cql/
https://github.com/folio-org/raml-module-builder#cql-relations
https://www.loc.gov/standards/sru/cql/spec.html
https://www.loc.gov/standards/sru/cql/contextSets/theCqlContextSet.html


adamdickmeiss (commented):
If it's up to the user of this utility to add quotes where necessary, I'm all good.

But what about tokens (whitespace)?

Eg username=first last 

Is it up to the API user to put quotes around that? You could add a utility to do just that. 

Note that the cql-java does allow two tokens like that but it's not standard.. You'll have to use
username="first last"

Note that quotes must be added just not in case of whitespace, but whenever an operator character is part of the string.. Eg username=a>b  ..


https://github.com/indexdata/yaz/blob/0b8d1700b31baba3e73f353e49b09178a11ed11e/src/rpn2cql.c#L223

julianladisch:
I've reworded the javadoc to better explain the quoting.

adamdickmeiss (approved):
Looks good. Thx

PR #778, opened 2020-09-22
Remove some dead code and tests RMB-730

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/security/AES.java
```diff
@@ -1,23 +1,20 @@
...
 import javax.crypto.spec.SecretKeySpec;
 
-
 public class AES {
 
   private static String secretKey = null;
 
+  private AES() {
+    throw new UnsupportedOperationException("Cannot instantiate AES class");
+  }
```
julianladisch, line 16: Can you add test coverage using https://github.com/folio-org/raml-module-builder/blob/v31.1.0/testing/src/main/java/org/folio/rest/testing/UtilityClassTester.java#L52-L58 , example: https://github.com/folio-org/raml-module-builder/blob/087e899fd2dc12758da09462e52844431c2b30ab/domain-models-runtime/src/test/java/org/folio/rest/tools/utils/TenantToolTest.java#L16

julianladisch (approved):

domain-models-runtime/src/main/java/org/folio/rest/security/AES.java
adamdickmeiss: Neat.. Should do that for Okapi util classes too.

PR #781, opened 2020-09-24
Upgrade to Vert.x 3.9.3 RMB-733

adamdickmeiss:


adamdickmeiss:
Breaking changes again in a what is apparently a bug fix release of Vert.x : https://github.com/eclipse-vertx/vertx-sql-client/pull/750
and yet https://github.com/eclipse-vertx/vertx-sql-client/pull/710 is stalled for 2 months.

domain-models-runtime/src/main/java/org/folio/rest/persist/helpers/LocalRowSet.java
```diff
@@ -43,6 +44,11 @@ public int rowCount() {
     return columns;
   }
 
+  @Override
+  public List<ColumnDescriptor> columnDescriptors() {
+    return new LinkedList<>();
+  }
+
```
julianladisch, line 51: I would rather use

```
  RowDesc rowDesc = new RowDesc(Collections.emptyList());
```
```
  public LocalRowSet withColumns(List<String> columns) {
    this.rowDesc = new RowDesc(columns);
    return this;
  }
```
```
  @Override
  public List<String> columnsNames() {
    return rowDesc.columnNames();
  }
```

```
  @Override
  public List<ColumnDescriptor> columnDescriptors() {
    return rowDesc.columnDescriptor();
  }
```

adamdickmeiss: sweet

julianladisch (approved):

PR #782, opened 2020-09-24
RMB-735: Reject unsupported sort modifiers in getWithOptimizedSql

julianladisch:
Sorting in CQL is expessed using the sortby keyword followed by
a list of sort-keys, each expressed as a CQL index data optionally
followed by one or more index modifiers — see
http://zing.z3950.org/cql/sorting.html#6.
For example, title/sort.descending/sort.ignorecase.

RMB handles the sort.ascending and sort.descending modifiers.
It rejects all other sort modifiers with a
CQLFeatureUnsupportedException:
https://github.com/folio-org/raml-module-builder/blob/v31.1.0/cql2pgjson/src/main/java/org/folio/cql2pgjson/model/CqlModifiers.java#L71

For optimized queries (that we use for instance.title only)
we have some lazy code that doesn't reject other sort modifiers:
https://github.com/folio-org/raml-module-builder/blob/v31.1.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L1097

This pull request fixes this inconsistency: Reject unsupported sort
modifiers in optimized queries (PgUtil.getWithOptimizedSql) as well.

Example that demonstrates the inconsistency:

((keyword=water) and (source=marc)) sortby title/sort.missingLow/sort.ascending/sort.respectCase

is accepted; but if I add another sort-key, hrid/sort.descending
— which you will note does not include a "missing-value" modified — yielding

((keyword=water) and (source=marc)) sortby title/sort.missingLow/sort.ascending/sort.respectCase hrid/sort.descending

then that sort is rejected with "Unsupported modifier sort.missinglow".
That can't be right.

Note that RMB uses schema.json to determine whether the index is
case sensitive or accents sensitive. See https://issues.folio.org/browse/RMB-734
for the feature request to allow those sort modifiers that match
the index definition in schema.json.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
@@ -34,6 +34,7 @@ import org.folio.rest.tools.utils.TenantTool;
 import org.folio.rest.tools.utils.ValidationHelper;
 import org.folio.util.UuidUtil;
 import org.folio.cql2pgjson.CQL2PgJSON;
+import org.folio.cql2pgjson.exception.CQLFeatureUnsupportedException;
 import org.folio.cql2pgjson.exception.FieldException;
 import org.folio.cql2pgjson.exception.QueryValidationException;
 import org.folio.rest.persist.cql.CQLWrapper;
@@ -1083,7 +1084,7 @@ public final class PgUtil {
     }
   }
 
-  private static String getAscDesc(ModifierSet modifierSet) {
+  private static String getAscDesc(ModifierSet modifierSet) throws CQLFeatureUnsupportedException {
     String ascDesc = "";
     for (Modifier modifier : modifierSet.getModifiers()) {
       switch (modifier.getType()) {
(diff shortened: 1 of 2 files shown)


MikeTaylor (approved):
Looks reasonable to me.

(Though I have not written Java in 20 years, so I am no expert on the details.)

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1094,7 +1095,7 @@ private static String getAscDesc(ModifierSet modifierSet) {
         ascDesc = "DESC";
         break;
       default:
-        // ignore
+        throw new CQLFeatureUnsupportedException("CQL: Unsupported modifier " + modifier.getType());
       }
```
adamdickmeiss, line 1099: Wonder why this is not covered in the tests. It looks as if it should be.

adamdickmeiss (commented):
I wonder why the throw

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
julianladisch: The wrong sortBy column caused the non-optimized code path been executed.

adamdickmeiss (approved):

hjiebsco (approved):

PR #789, opened 2020-10-12
RMB-609 [Work in progress]

MikhailFokanov:
Migrations to promises and WebClients are done. Please provide early feedback.

Use https://github.com/folio-org/raml-module-builder/pull/789 to test it out on mod-inventory-storage.


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "domain-models-runtime-it/ramls/raml-util"]
 	path = domain-models-runtime-it/ramls/raml-util
 	url = https://github.com/folio-org/raml
+[submodule "ramls/raml-util"]
+	path = ramls/raml-util
+	url = https://github.com/folio-org/raml
```
julianladisch, line 9: This creates the new submodule "ramls" and creates the raml-util submodule in it.
What purpose does it serve?

domain-models-interface-extensions/pom.xml
```diff
@@ -11,6 +11,12 @@
     <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-core</artifactId>
+      <version>4.0.0.Beta3</version>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-web-client</artifactId>
+      <version>4.0.0.Beta3</version>
```
julianladisch, line 19: Mixing version is likely to fail.
```
mvn dependency:tree -Dincludes=*vertx*

[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ domain-models-interface-extensions ---
[INFO] org.folio:domain-models-interface-extensions:jar:31.2.0-SNAPSHOT
[INFO] +- io.vertx:vertx-core:jar:4.0.0.Beta3:compile
[INFO] +- io.vertx:vertx-web-client:jar:4.0.0.Beta3:compile
[INFO] |  \- io.vertx:vertx-web-common:jar:3.9.3:compile
[INFO] \- io.vertx:vertx-web:jar:3.9.3:compile
[INFO]    +- io.vertx:vertx-auth-common:jar:3.9.3:compile
[INFO]    \- io.vertx:vertx-bridge-common:jar:3.9.3:compile
```

pom.xml
```diff
@@ -51,7 +51,7 @@
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-stack-depchain</artifactId>
-        <version>3.9.4</version>
+        <version>3.9.3</version>
```
julianladisch, line 54: Can you set this to 4.0.0.Beta3?

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/ClientGeneratorTest.java
```diff
@@ -45,204 +45,21 @@ public void doesGenerateTestResourceClient() throws Exception {
...
+    String expected = removeTrailingSpaces(FileUtils.readFileToString(
+        new File(ClientGeneratorTest.class.getResource("/clients/TestClient.txt").getFile()),
+        StandardCharsets.UTF_8));
+
     Assert.assertEquals(expected, actual);
   }
 
+
+  private String removeTrailingSpaces(String str) {
+    return str.replaceAll("(\\s+)\n","\n");
```
julianladisch, line 62: Braces are not needed.
`\n` doesn't match all possible end of line characters, and it doesn't match the last line that does not end with `\n`.

    private static final Pattern TRAINLING_SPACE_PATTERN = Pattern.compile("\\s+$", Pattern.MULTILINE);
    
    private String removeTrailingSpaces(String str) {
      return TRAINLING_SPACE_PATTERN.matcher(str).replaceAll("");
    }

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -631,7 +631,7 @@ public void startTx(Handler<AsyncResult<SQLConnection>> done) {
       }
       try {
         SQLConnection pgTransaction = new SQLConnection(res.result(),
-            res.result().begin(), null);
+            res.result().begin().result(), null);
```
julianladisch, line 634: begin() is a Future<Transaction>. It takes some time until the database has responded to the begin request.
begin().result() will be null.
You need some async handling, for example

        res.result().begin()
        .map(transaction -> new SQLConnection(res.result(), transaction, null))
        .onComplete(done);

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1894,7 +1894,7 @@ public TotaledResults(RowSet<Row> set, Integer estimatedTotal) {
                             Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
     // Start a transaction that we need to close.
     // If a transaction is already running we don't need to close it.
-    final Transaction transaction = connection.tx == null ? connection.conn.begin() : null;
+    final Transaction transaction = connection.tx == null ? connection.conn.begin().result() : null;
```
julianladisch, line 1897: begin() is a Future. It takes some time until the database has responded to the begin request.
begin().result() will be null.
You need some async handling, for example begin().map(...) or begin().compose(...)

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1922,7 +1922,7 @@ public TotaledResults(RowSet<Row> set, Integer estimatedTotal) {
 
   private void closeIfNonNull(Transaction transaction) {
     if (transaction != null) {
-      transaction.close();
+      transaction.completion();
```
julianladisch, line 1925: completion() does not do anything, it only returns a Future.
You need to call commit().

```
private Future<Void> commitIfNonNull(Transaction transaction) {
  if (transaction == null) {
  }
  return transaction.commit();
}
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3517,7 +3518,13 @@ public void dropCreateDatabase(String database) throws SQLException {
...
   public Future<List<String>> runSQLFile(String sqlFile, boolean stopOnError) {
     Promise<List<String>> promise = Promise.promise();
-    runSQLFile(sqlFile, stopOnError, promise.future());
+    runSQLFile(sqlFile, stopOnError, (res) -> {
+        if (res.failed()) {
+          log.error("Error while executing SQL :" + res.cause().getMessage());
+          promise.fail(res.cause());
+        }
+        promise.complete(res.result());
+        });
```
julianladisch, line 3527: This change is not needed because runSQLFile(String, boolean, Handler) already has logged the error.

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
```diff
@@ -75,8 +77,13 @@ public PreparedRowStream(PreparedStatement preparedStatement, int fetch, Tuple t
...
   @Override
-  public void close() {
-    rowStream.close(close -> preparedStatement.close());
+  public Future<Void> close() {
+    Promise<Void> p = Promise.promise();
+    rowStream.close(close -> {
+      preparedStatement.close();
+      p.complete();
+    });
+    return p.future();
```
julianladisch, line 86: This code doesn't propagates failures from preparedStatement.close() but it should. Simply use the existing close(Handler) code:

```
  @Override
  public Future<Void> close() {
    Promise<Void> promise = Promise.promise();
    close(promise);
    return promise.future();
  }
```

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -1,116 +1,106 @@
...
-class HTTPJsonResponseHandler implements Handler<HttpClientResponse> {
+class HTTPJsonResponseHandler implements Handler<AsyncResult<HttpResponse<Buffer>>> {
 
   private static final Logger log = LoggerFactory.getLogger(HTTPJsonResponseHandler.class);
 
   CompletableFuture<Response> cf;
   String endpoint;
   RollBackURL rollbackURL;
-  HttpClient httpClient;
+  WebClient httpClient;
```
julianladisch, line 25:
```suggestion
  WebClient webClient;
```

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -234,19 +231,20 @@ private static String getId(LoadingEntry loadingEntry, URL url, String content,
...
 
-  private static void handleException(Throwable ex, String lead, Future<Void> f) {
-    String diag = lead  + ": " + ex.getMessage();
+  private static void handleException(Throwable ex, String lead, Promise<Void> f) {
+    String diag = lead + ": " + ex.getMessage();
     log.error(diag, ex);
-    if (!f.isComplete()) {
+    if (!f.future().isComplete()) {
       f.handle(Future.failedFuture(diag));
     }
```
julianladisch, line 239:
```suggestion
    f.tryFail(diag);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
```diff
@@ -76,7 +78,9 @@ public PreparedRowStream(PreparedStatement preparedStatement, int fetch, Tuple t
 
   @Override
   public void close() {
-    rowStream.close(close -> preparedStatement.close());
+    rowStream.close(close -> {
+      preparedStatement.close();
+    });
```
julianladisch, line 83: Sonar will complain about this ("Lambdas containing only one statement should not nest this statement in a block").
https://sonarcloud.io/organizations/folio-org/rules?open=java%3AS1602&rule_key=java%3AS1602

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
MikhailFokanov: done.

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/ClientGeneratorTest.java
MikhailFokanov: done.

pom.xml
MikhailFokanov: done.

domain-models-interface-extensions/pom.xml
MikhailFokanov: done.

.gitmodules
MikhailFokanov: done.

MikhailFokanov: I don't know why, but build fails on: "Could not perform submodule update" both locally and on Jenkins without this

MikhailFokanov:
@julianladisch , @adamdickmeiss After merge from master build fails with
"Could not find artifact io.vertx:vertx-pg-client:jar:4.0.0.Beta3-FOLIO in folio-nexus"

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/ClientGeneratorTest.java
```diff
@@ -16,6 +16,8 @@
 
 public class ClientGeneratorTest {
 
+  private static final Pattern TRAINLING_SPACE_PATTERN = Pattern.compile("\\s+$", Pattern.MULTILINE);
```
julianladisch, line 19: Fix my typo
```suggestion
  private static final Pattern TRAILING_SPACE_PATTERN = Pattern.compile("\\s+$", Pattern.MULTILINE);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1894,12 +1894,28 @@ public TotaledResults(RowSet<Row> set, Integer estimatedTotal) {
                             Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
     // Start a transaction that we need to close.
     // If a transaction is already running we don't need to close it.
-    final Transaction transaction = connection.tx == null ? connection.conn.begin() : null;
+    if (connection.tx == null) {
+      connection.conn.begin().onComplete((AsyncResult<Transaction> trans) -> {
+        if (trans.failed()) {
+          log.error(trans.cause());
```
julianladisch, line 1900: The error should also been reported to replyHandler.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1894,12 +1894,28 @@ public TotaledResults(RowSet<Row> set, Integer estimatedTotal) {
                             Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler) {
     // Start a transaction that we need to close.
     // If a transaction is already running we don't need to close it.
-    final Transaction transaction = connection.tx == null ? connection.conn.begin() : null;
+    if (connection.tx == null) {
+      connection.conn.begin().onComplete((AsyncResult<Transaction> trans) -> {
```
julianladisch, line 1898: `onComplete` and `if` can be replaced by `onSuccess(trans -> ...).onFailure(e -> ...)` for readability.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3517,7 +3536,12 @@ public void dropCreateDatabase(String database) throws SQLException {
    */
   public Future<List<String>> runSQLFile(String sqlFile, boolean stopOnError) {
     Promise<List<String>> promise = Promise.promise();
-    runSQLFile(sqlFile, stopOnError, promise.future());
+    runSQLFile(sqlFile, stopOnError, (res) -> {
```
julianladisch, line 3539: This change should be reverted.

domain-models-interface-extensions/pom.xml
```diff
@@ -11,6 +11,12 @@
     <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-core</artifactId>
+      <version>4.0.0.Beta3</version>
```
julianladisch, line 14: Can you remove the version line so that is uses the version provided by the parent pom.xml?

domain-models-interface-extensions/pom.xml
julianladisch, line 19: Can you remove the version line so that is uses the version provided by the parent pom.xml?

pom.xml
```diff
@@ -59,12 +59,12 @@
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-sql-client</artifactId>
-        <version>${vertx.version}-FOLIO</version>
+        <version>${vertx.version}</version>
```
julianladisch, line 62: Please revert to use the -FOLIO version that [RMB-752](https://issues.folio.org/browse/RMB-752) has released.

pom.xml
```diff
@@ -59,12 +59,12 @@
...
       </dependency>
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-pg-client</artifactId>
-        <version>${vertx.version}-FOLIO</version>
+        <version>${vertx.version}</version>
```
julianladisch, line 67: Please revert to use the -FOLIO version that [RMB-752](https://issues.folio.org/browse/RMB-752) has released.

.gitmodules
adamdickmeiss: We need to see why this is necessary.

pom.xml
```diff
@@ -29,7 +29,7 @@
     <aspectj.version>1.9.6</aspectj.version>
     <junit-jupiter-version>5.7.0</junit-jupiter-version>
     <postgresql-embedded-version>2.10</postgresql-embedded-version>
-    <vertx.version>3.9.4</vertx.version>
+    <vertx.version>4.0.0.Beta3</vertx.version>
```
adamdickmeiss, line 33: You can use 4.0.0.CR1 now, just to be closer to final 4.0.0

adamdickmeiss (commented):
Looking good so far. I'll look more at ait.

.gitmodules
adamdickmeiss: Using
```sh
  git submodule deinit ramls/raml-util
  git rm ramls/raml-util
```
it is gone. And RMB still compiles happily on my machine.

MikhailFokanov: @adamdickmeiss it is working on your machine, could you please commit it to the branch, so that it builds successfully on Jenkins.

adamdickmeiss:
Why was the version bumped? If anything it should be 32.0.0-SNAPSHOT

.gitmodules
julianladisch: Resolved.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -209,16 +213,17 @@ private void addConstructorOkapi6Args(JFieldVar tokenVar, JFieldVar options, JFi
     conBody.assign(JExpr._this().ref(tenantId), tenantIdVar);
     conBody.assign(JExpr._this().ref(tokenVar), token);
     conBody.assign(JExpr._this().ref(okapiUrl), okapiUrlVar);
-    conBody.assign(options, JExpr._new(jcodeModel.ref(HttpClientOptions.class)));
+    conBody.assign(options, JExpr._new(jcodeModel.ref(io.vertx.ext.web.client.WebClientOptions.class)));
```
julianladisch, line 216:
```suggestion
    conBody.assign(options, JExpr._new(jcodeModel.ref(WebClientOptions.class)));
```

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -260,10 +265,10 @@ public void generateClassMeta(String className) {
       okapiUrl = jc.field(JMod.PRIVATE, String.class, OKAPI_URL);
 
       /* class variable to http options */
-      JFieldVar options = jc.field(JMod.PRIVATE, HttpClientOptions.class, "options");
+      JFieldVar options = jc.field(JMod.PRIVATE, WebClientOptions.class, "options");
 
       /* class variable to http client */
-      JFieldVar httpClient = jc.field(JMod.PRIVATE, HttpClient.class, "httpClient");
+      JFieldVar httpClient = jc.field(JMod.PRIVATE, WebClient.class, "httpClient");
```
julianladisch, line 271:
```suggestion
      JFieldVar webClient = jc.field(JMod.PRIVATE, WebClient.class, "webClient");
```
All `httpClient` usages should be renamed to `webClient` to reflect reality.

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -383,17 +388,17 @@ public void generateMethodMeta(String methodName, JsonObject params, String url,
     ifClause2._then().directStatement("request.putHeader(\"X-Okapi-Url\", okapiUrl);");
 
     /* add response handler to each function */
-    JClass handler = jcodeModel.ref(Handler.class).narrow(HttpClientResponse.class);
+    JClass handler = jcodeModel.ref(Handler.class).narrow(jcodeModel.ref(AsyncResult.class).narrow(jcodeModel.ref(HttpResponse.class).narrow(Buffer.class)));
```
julianladisch, line 391:
```suggestion
    JClass handler = jcodeModel.ref(HttpResponse.class).narrow(Buffer.class);
    handler = jcodeModel.ref(AsyncResult.class).narrow(handler);
    handler = jcodeModel.ref(Handler.class).narrow(handler);
    jmCreate.param(handler, "responseHandler");
```

domain-models-interface-extensions/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -383,17 +388,17 @@ public void generateMethodMeta(String methodName, JsonObject params, String url,
...
     jmCreate.param(handler, "responseHandler");
 
     /* if we need to pass data in the body */
     if(bodyContentExists[0]){
       body.directStatement("request.putHeader(\"Content-Length\", buffer.length()+\"\");");
-      body.directStatement("request.setChunked(true);");
-      body.directStatement("request.write(buffer);");
+      //body.directStatement("request.setChunked(true);"); //TODO: consider adding this somehow
```
julianladisch, line 397: Remove the lines with content-length and setChunked.
Then WebClient may chunk the request: https://vertx-web-site.github.io/docs/vertx-web-client/java/#_writing_request_bodies

domain-models-interface-extensions/src/test/java/org/folio/rest/tools/ClientGeneratorTest.java
```diff
@@ -45,204 +47,21 @@ public void doesGenerateTestResourceClient() throws Exception {
...
-
-    Assert.assertEquals(expected.size(), actual.size());
+    // IDEs always removes trailing spaces from edited files, but java code generator adds then,
+    // so we nee to remove them before the comparison
+
+    String actual = removeTrailingSpaces(FileUtils.readFileToString(expectedClient, StandardCharsets.UTF_8));
+
+    String expected = removeTrailingSpaces(FileUtils.readFileToString(
+        new File(ClientGeneratorTest.class.getResource("/clients/TestClient.txt").getFile()),
+        StandardCharsets.UTF_8));
```
julianladisch, line 57:
```suggestion
    String expected = removeTrailingSpaces(ResourceUtil.asString("/clients/TestClient.txt", this.getClass()));
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1894,12 +1899,28 @@ public TotaledResults(RowSet<Row> set, Integer estimatedTotal) {
...
+
+  private <T> void executeGetQuery(SQLConnection connection, QueryHelper queryHelper,
+      ResultInfo resultInfo, Class<T> clazz,
+      Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler, Transaction transaction) {
     connection.conn.prepare(queryHelper.selectQuery, prepareRes -> {
       if (prepareRes.failed()) {
-        closeIfNonNull(transaction);
+        closeIfNonNull(transaction).onComplete((voidRes) -> {
         log.error(prepareRes.cause().getMessage(), prepareRes.cause());
         replyHandler.handle(Future.failedFuture(prepareRes.cause()));
```
julianladisch, line 1922:
```suggestion
          log.error(prepareRes.cause().getMessage(), prepareRes.cause());
          replyHandler.handle(Future.failedFuture(prepareRes.cause()));
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1894,12 +1899,28 @@ public TotaledResults(RowSet<Row> set, Integer estimatedTotal) {
...
+    }
+  }
+
+  private <T> void executeGetQuery(SQLConnection connection, QueryHelper queryHelper,
+      ResultInfo resultInfo, Class<T> clazz,
+      Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler, Transaction transaction) {
     connection.conn.prepare(queryHelper.selectQuery, prepareRes -> {
       if (prepareRes.failed()) {
-        closeIfNonNull(transaction);
+        closeIfNonNull(transaction).onComplete((voidRes) -> {
```
julianladisch, line 1920:
```suggestion
        closeIfNonNull(transaction).onComplete(ignore -> {
```
Braces not needed.
`ignore` makes clear that we ignore a close failure on purpose.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1960,19 +1982,19 @@ private void closeIfNonNull(Transaction transaction) {
...
         rowStream.close();
-        closeIfNonNull(transaction);
-        streamResult.fireExceptionHandler(e);
+        closeIfNonNull(transaction).onComplete((AsyncResult<Void> voidRes) -> streamResult.fireExceptionHandler(e));
       }
     }).endHandler(v2 -> {
       rowStream.close();
-      closeIfNonNull(transaction);
-      resultInfo.setTotalRecords(
+      closeIfNonNull(transaction).onComplete((AsyncResult<Void> voidRes) -> {
```
julianladisch, line 1996:
```suggestion
      closeIfNonNull(transaction).onComplete(ignore -> {
```
Braces not needed. Type declaration not needed.
The name `ignore` makes clear that we ignore a close failure on purpose.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1985,14 +2007,16 @@ private void closeIfNonNull(Transaction transaction) {
       } catch (Exception ex) {
         streamResult.fireExceptionHandler(ex);
       }
+      });
     }).exceptionHandler(e -> {
       rowStream.close();
-      closeIfNonNull(transaction);
-      if (!promise.future().isComplete()) {
+      closeIfNonNull(transaction).onComplete((AsyncResult<Void> voidRes) -> {
```
julianladisch, line 2013:
```suggestion
      closeIfNonNull(transaction).onComplete(ignore -> {
```
Braces not needed. Type declaration not needed.
The name `ignore` makes clear that we ignore a close failure on purpose.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -1985,14 +2007,16 @@ private void closeIfNonNull(Transaction transaction) {
...
+        if (!promise.future().isComplete()) {
         promise.complete(streamResult);
         replyHandler.handle(promise.future());
       }
       streamResult.fireExceptionHandler(e);
+      });
     });
   }
```
julianladisch, line 2021: Can you fix the indentation of the lines of this method?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2827,7 +2851,7 @@ public ResultsHelper(Class<T> clazz) {
...
 
   private boolean isStringArrayType(Object value) {
-    // https://github.com/eclipse-vertx/vertx-sql-client/blob/4.0.0.Beta3/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java#L737
+    // https://github.com/eclipse-vertx/vertx-sql-client/blob/4.0.0.CR1/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java#L737
```
julianladisch, line 2854:
```suggestion
    // https://github.com/eclipse-vertx/vertx-sql-client/blob/4.0.0.CR1/vertx-sql-client/src/main/java/io/vertx/sqlclient/Tuple.java#L910
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PreparedRowStream.java
```diff
@@ -75,8 +77,10 @@ public PreparedRowStream(PreparedStatement preparedStatement, int fetch, Tuple t
   }
 
   @Override
-  public void close() {
-    rowStream.close(close -> preparedStatement.close());
+  public Future<Void> close() {
+    Promise<Void> promise = Promise.promise();
+    close(promise);
+    return promise.future();
```
julianladisch, line 83:
```suggestion
    return Future.future(promise -> close(promise));
```

domain-models-runtime/src/main/java/org/folio/rest/persist/cql/CQLWrapper.java
```diff
@@ -290,7 +289,7 @@ public String toString() {
     spaceAppend(sb, offset.toString());
     String sql = sb.toString();
     if (log.isInfoEnabled()) {
-      log.info(getType() + " >>> SQL: " + getQuery() + " >>>" + sql);
+      log.info(String.format("%s >>> SQL: %s >>>%s", getType(), getQuery(), sql));
```
julianladisch, line 292:
```suggestion
      log.info("{} >>> SQL: {} >>>{}", getType(), getQuery(), sql));
```
String.format is slow and should be avoided for logging.

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -1,116 +1,106 @@
...
+        log.error(e.getMessage(), e);
       }
+    }
 /*      if(r.error != null && rollbackURL != null){
 
       }*/
-    });
-    hcr.exceptionHandler( eh -> {
-      if(httpClient != null){
+  /* //TODO: think if it is needed
```
julianladisch, line 70: Yes, it is still needed for HttpModuleClient2 autoCloseConnections.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3602,7 +3627,7 @@ public void dropCreateDatabase(String database) throws SQLException {
    */
   public Future<List<String>> runSQLFile(String sqlFile, boolean stopOnError) {
     Promise<List<String>> promise = Promise.promise();
-    runSQLFile(sqlFile, stopOnError, promise.future());
+     runSQLFile(sqlFile, stopOnError, promise);
     return promise.future();
```
julianladisch, line 3631: return Future.future(promise -> runSQLFile(sqlFile, stopOnError, promise));

domain-models-runtime/src/test/resources/log4j2-test.properties
```diff
@@ -11,7 +12,7 @@ appenders = console
 appender.console.type = Console
 appender.console.name = STDOUT
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{dd MMM yyyy HH:mm:ss:SSS} %-5p %C{1} [%X{reqId}] %m%n
+appender.console.layout.pattern = %d{HH:mm:ss} [$${FolioLoggingContext:requestid}] [$${FolioLoggingContext:tenantid}] [$${FolioLoggingContext:userid}] [$${FolioLoggingContext:moduleid}] %-5p %-20.20C{1} %m%n
```
julianladisch, line 15: Can you add a note to the https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md that RMB based modules should copy https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/test/resources/log4j2-test.properties into their /src/test/resources directory?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -322,6 +321,13 @@ public PgConnection prepare(String s, Handler<AsyncResult<PreparedStatement>> ha
       return this;
     }
 
+    @Override
+    public Future<PreparedStatement> prepare(String s) {
+      Promise<PreparedStatement> promise = Promise.promise();
+      prepare(s, promise);
+      return promise.future();
+    }
+
```
julianladisch, line 330:
```
    @Override
    public PgConnection prepare(String s, Handler<AsyncResult<PreparedStatement>> handler) {
      handler.handle(prepare(s));
      return this;
    }

    @Override
    public Future<PreparedStatement> prepare(String s) {
      return Future.failedFuture("not implemented");
    }
```

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -462,93 +455,11 @@ public void options() {
...
-    try {
-      request.write(getFile("job_conf_post.json"));
-    } catch (IOException e) {
-      log.error(e.getMessage(), e);
-      context.fail("unable to read file");
-    }
-    request.end();
+      aClient.putAdminLoglevel(AdminLoglevelPutLevel.FINE, "org", context.asyncAssertSuccess());
+      aClient.getAdminJstack(context.asyncAssertSuccess());
+      aClient.getAdminMemory(false, context.asyncAssertSuccess());
```
julianladisch, line 462: Can you fix the indentation?

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1745,13 +1741,14 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
       }
 
       @Override
-      public void begin(Handler<AsyncResult<Transaction>> handler) {
-        handler.handle(Future.succeededFuture(null));
+      public void close(Handler<AsyncResult<Void>> handler) {
+
```
julianladisch, line 1745:
```suggestion
      handler.handle(close());
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1734,6 +1725,11 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
         handler.handle(Future.succeededFuture(null));
       }
 
+      @Override
+      public Future<SqlConnection> getConnection() {
+        return null;
+      }
+
```
julianladisch, line 1732:
```
      @Override
      public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
        handler.handle(getConnection());
      }

      @Override
      public Future<SqlConnection> getConnection() {
      }
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1773,6 +1770,11 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
         handler.handle(Future.failedFuture("postgresClientGetConnectionFails"));
       }
 
+      @Override
+      public Future<SqlConnection> getConnection() {
+        return null;
+      }
+
```
julianladisch, line 1777:
```
      @Override
      public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
        handler.handle(getConnection());
      }

      @Override
      public Future<SqlConnection> getConnection() {
        return Future.failedFuture("postgresClientGetConnectionFails");
      }
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1972,6 +2014,11 @@ public void execute(Handler<AsyncResult<RowSet<Row>>> handler) {
             handler.handle(Future.failedFuture("queryFails"));
           }
 
+          @Override
+          public Future<RowSet<Row>> execute() {
+            return null;
+          }
+
```
julianladisch, line 2021:
```
          @Override
          public void execute(Handler<AsyncResult<RowSet<Row>>> handler) {
            handler.handle(execute());
          }

          @Override
          public Future<RowSet<Row>> execute() {
            return Future.failedFuture("queryFails");
          }
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -366,6 +383,13 @@ public void execute(Handler<AsyncResult<RowSet<Row>>> handler) {
           }
         }
 
+        @Override
+        public Future<RowSet<Row>> execute() {
+          Promise<RowSet<Row>> promise = Promise.promise();
+          execute(promise::handle);
+          return promise.future();
```
julianladisch, line 390:
```suggestion
          return Future.future(promise -> execute(promise));
```

doc/upgrading.md
```diff
@@ -17,6 +18,24 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
 * [Version 20](#version-20)
 
+## Version 32.0
+
+* [RMB-609](https://issues.folio.org/browse/RMB-609) Upgrade to Vert.x 4:
+  * Replace deprecated `io.vertx.core.logging.Logger` by
+    `org.apache.logging.log4j.Logger`, for example
+    `private static final Logger LOGGER = LogManager.getLogger(Foo.class)`
+    (using `getLogger()` without argument requires Multi-Release for
+    stack walking, see below).
```
marcjohnson-kint, line 28: I didn't notice any further remarks on multi-release stack walking. What content is this referring to?

MikhailFokanov: @julianladisch ?

adamdickmeiss: Probably the Multi-Release section in 31.0.

julianladisch: Yes, line 85.

doc/upgrading.md
```diff
@@ -17,6 +18,24 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+  * Replace `HttpClient` `.getAbs`, `.postAbs`, `.putAbs`, `.deleteAbs` by
+    `WebClient.requestAbs(HttpMethod method, ...)`.
+  * Replace `Future.setHandler` by `Future.onComplete`.
+  * `Future` has been split into `Future` and `Promise`, see
+    [Futurisation in Vert.x 4](futurisation.md).
+  * In order to have structured logging for tests RMB based modules should copy [log4j2-test.properties](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/test/resources/log4j2-test.properties) into their `/src/test/resources` directory
+* [OKAPI-943](https://issues.folio.org/browse/OKAPI-943) When calling `.any`, `.all` or `.join`
+  replace `io.vertx.core.CompositeFuture` by `org.folio.okapi.common.GenericCompositeFuture`,
+  replace raw type by actual type and remove `@SuppressWarnings("rawtypes")`.
+
```
adamdickmeiss, line 38: The upgrading notes should mention that ClientGenerator functions has different responseHandler. From 
``
  Handler<HttpClientResponse>
``

to 
``
  Handler<AsyncResult<HttpResponse<Buffer>>>
``

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1745,13 +1741,14 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
...
-        handler.handle(Future.succeededFuture(null));
+      public void close(Handler<AsyncResult<Void>> handler) {
+      handler.handle(close());
       }
 
       @Override
-      public void close() {
+      public Future<Void> close() {
         // nothing to do
+        return Future.succeededFuture(null);
```
julianladisch, line 1751:
```suggestion
        return Future.succeededFuture();
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1745,13 +1741,14 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
       }
 
       @Override
-      public void begin(Handler<AsyncResult<Transaction>> handler) {
-        handler.handle(Future.succeededFuture(null));
+      public void close(Handler<AsyncResult<Void>> handler) {
+      handler.handle(close());
```
julianladisch, line 1745:
```suggestion
        handler.handle(close());
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1784,13 +1787,14 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
...
       @Override
-      public void begin(Handler<AsyncResult<Transaction>> handler) {
+      public void close(Handler<AsyncResult<Void>> handler) {
 
       }
 
       @Override
-      public void close() {
+      public Future<Void> close() {
         // nothing to do
```
julianladisch, line 1796:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1864,8 +1884,8 @@ public boolean isSSL() {
       }
 
       @Override
-      public void close() {
-
+      public Future<Void> close() {
+        return Future.succeededFuture(null);
```
julianladisch, line 1888:
```suggestion
        return Future.succeededFuture();
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1891,13 +1916,15 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
...
-      public void begin(Handler<AsyncResult<Transaction>> handler) {
+      public void close(Handler<AsyncResult<Void>> handler) {
 
       }
 
+
       @Override
-      public void close() {
+      public Future<Void> close() {
         // nothing to do
```
julianladisch, line 1926:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -2016,13 +2070,15 @@ public void getConnection(Handler<AsyncResult<SqlConnection>> handler) {
...
```
julianladisch, line 2080:
```suggestion
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
julianladisch: Returning null as a Future most likely causes a NullPointerException. The comment "nothing to do" is misleading and should be removed.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -3036,8 +3115,8 @@ public void streamGetConnectionFailed(TestContext context) throws IOException, F
     }
 
     @Override
-    public void close() {
-
+    public Future<Void> close() {
+      return Future.succeededFuture(null);
```
julianladisch, line 3119:
```suggestion
      return Future.succeededFuture();
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
MikhailFokanov: It wasn't my comment.

adamdickmeiss (approved):

julianladisch:
mod-inventory-storage compiles with this PR but there are some unit test regressions, for details see
https://issues.folio.org/browse/MODINVSTOR-624
https://github.com/folio-org/mod-inventory-storage/pull/546

doc/upgrading.md
```diff
@@ -19,6 +20,22 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+    `private static final Logger LOGGER = LogManager.getLogger(Foo.class)`
+    (using `getLogger()` without argument requires Multi-Release for
+    stack walking, see below).
+  * Replace `HttpClient` `.getAbs`, `.postAbs`, `.putAbs`, `.deleteAbs` by
+    `WebClient.requestAbs(HttpMethod method, ...)`.
+  * Replace `Future.setHandler` by `Future.onComplete`.
+  * `Future` has been split into `Future` and `Promise`, see
+    [Futurisation in Vert.x 4](futurisation.md).
+  * In order to have structured logging for tests RMB based modules should copy [log4j2-test.properties](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/test/resources/log4j2-test.properties) into their `/src/test/resources` directory
+  * The responseHandler function in ClientGenerator should be migrated from Handler<HttpClientResponse> to Handler<AsyncResult<HttpResponse<Buffer>>>
```
julianladisch, line 35: RMB based modules usually call ClientGenerator from pom.xml. One examle is mod-configuration. I don't see a change in https://github.com/folio-org/mod-configuration/pull/102/files , can you be more verbose what developers of RMB based modules should change?

MikhailFokanov: done

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -181,84 +179,70 @@ private static void endWithXHeaders(HttpClientRequest req, Map<String, String> h
...
-    if (f.isComplete()) {
-      return;
-    }
-    String id = getId(loadingEntry, url, content, f);
-    if (f.isComplete()) {
-      return;
-    }
+    return getContent(url, loadingEntry)
+        .compose(content -> getId(loadingEntry, url, content)
+            .compose(id -> loadURL(headers, content, id, httpClient, loadingEntry, endPointUrl)));
```
julianladisch, line 241:
```suggestion
        .compose(content -> getId(loadingEntry, url, content))
        .compose(id -> loadURL(headers, content, id, httpClient, loadingEntry, endPointUrl));
```

adamdickmeiss: Can't do: content not found. Am I blind here?

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -373,11 +332,21 @@ private static void loadURL(Map<String, String> headers, URL url,
...
-  }
+    WebClient httpClient = WebClient.create(vertx);
+    Future<Integer> future = Future.succeededFuture(recordsLoaded);
+    for (LoadingEntry entry : loadingEntries) {
+      if (ta != null) {
+        final String okapiUrlFinal = okapiUrl;
+        for (Parameter parameter : ta.getParameters()) {
+          if (entry.key.equals(parameter.getKey()) && "true".equals(parameter.getValue())) {
+            future = future.compose(x -> loadData(okapiUrlFinal, headers, entry, httpClient)
+                .compose(y -> Future.succeededFuture(x + y)));
```
julianladisch, line 343: Rename x and y, and use map():
```suggestion
            future = future.compose(sum -> loadData(okapiUrlFinal, headers, entry, httpClient)
                .map(newRecords -> sum + newRecords));
```

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
julianladisch: You are right, content is needed by loadURL.

julianladisch (approved):
Looks good, thanks!

PR #796, opened 2020-10-26
RMB-736: Use metrics util from Okapi-common

hjiebsco:
Note, once Okapi 4.3.0 is released, we can change POM `okapi-common` dependency from `4.3.0-SNAPSHOT` to `4.3.0`.


adamdickmeiss (approved):
ok. Not fond of it (with snapshots).. IMHO, you can develop RMB-736 completely (on your own machine) before needing this. (And even make a PR for Okapi in case you need another detail from the new common feature).

hjiebsco:
> ok. Not fond of it (with snapshots).. IMHO, you can develop RMB-736 completely (on your own machine) before needing this. (And even make a PR for Okapi in case you need another detail from the new common feature).

Just to clarify, this is the PR for RMB-736. My thinking is to let reviewers review the PR (in a green passed state, not red with build failure state), but we will wait till Okapi 4.3.0 is released before merging this PR.

julianladisch (approved):

PR #799, opened 2020-10-29
RMB-746: Release 31.1.4 with RMB-744 (REVOKE/tuple concurrently updated)

julianladisch:


hjiebsco (approved):
Might not be related. In Jason's test, there were some deadlock for some modules. Any idea why?

PR #804, opened 2020-11-04
RMB-755: Locks and ignore exceptions for REVOKE during upgrade

julianladisch:
mod-users and mod-circulation-storage have an ExpirationTool that regulary run a query against pg_catalog.pg_namespace.

This may interfere with RMB's REVOKE command that writes to pg_catalog.pg_namespace during module upgrades.

https://github.com/folio-org/mod-users/blob/v17.1.2/src/main/java/org/folio/rest/utils/ExpirationTool.java#L40

```
select nspname from pg_catalog.pg_namespace where nspname LIKE '%_mod_users'
```

https://github.com/folio-org/mod-circulation-storage/blob/v12.1.1/src/main/java/org/folio/support/ExpirationTool.java#L56

```
select nspname from pg_catalog.pg_namespace where nspname LIKE '%_mod_circulation_storage';
```

Solution:

We combine the locks of RMB-744 and RMB-750. Ignore any errors, the REVOKE will run again on next module upgrade.

### DIFF
--- a/domain-models-runtime/src/main/resources/templates/db_scripts/main.ftl
+++ b/domain-models-runtime/src/main/resources/templates/db_scripts/main.ftl
@@ -61,12 +61,27 @@ CREATE TABLE IF NOT EXISTS rmb_internal_analyze (
 </#if>
 
 DO $$
+DECLARE
+  n text;
 BEGIN
-  -- use advisory lock to prevent "tuple concurrently updated"
+  -- Use lock to prevent "tuple concurrently updated".
+  -- In AWS Aurora env, the db admin user cannot acquire row level lock on pg_catalog.pg_namespace
+  -- table due to permission issue and needs to use an advisory lock.
   -- https://issues.folio.org/browse/RMB-744
-  PERFORM pg_advisory_xact_lock(20201101, 1234567890);
-  REVOKE ALL PRIVILEGES ON SCHEMA public FROM ${myuniversity}_${mymodule};
-  REVOKE CREATE ON SCHEMA public FROM PUBLIC;
+  -- https://issues.folio.org/browse/RMB-750
+  -- https://issues.folio.org/browse/RMB-755
+  BEGIN
+    SELECT nspname INTO n FROM pg_catalog.pg_namespace WHERE nspname = 'public' FOR UPDATE;
+  EXCEPTION WHEN OTHERS THEN
+    PERFORM pg_advisory_xact_lock(20201101, 1234567890);
+  END;
+  BEGIN
+    REVOKE ALL PRIVILEGES ON SCHEMA public FROM ${myuniversity}_${mymodule};
+    REVOKE CREATE ON SCHEMA public FROM PUBLIC;
+  EXCEPTION WHEN OTHERS THEN
+    -- ignore; this happens if we use an advisory lock but the other query doesn't
+    NULL;
+  END;
 END $$;
 
 <#-- Loop over all tables that need updating / adding / deleting -->


hjiebsco (commented):
Curious why would select statements interfere with revoke statements?

julianladisch:
REVOKE statements make an update on the pg_catalog.pg_namespace row, even if the update doesn't change anything.

julianladisch:
```
GRANT ALL PRIVILEGES ON SCHEMA public TO diku_mod_users;
GRANT CREATE ON SCHEMA public TO PUBLIC;
SELECT * FROM pg_catalog.pg_namespace WHERE nspname = 'public';
```

```
 nspname | nspowner |                                         nspacl
---------+----------+----------------------------------------------------------------------------------------
 public  |       10 | {postgres=UC/postgres,=UC/postgres,diku_mod_users=UC/postgres}
```

```
REVOKE ALL PRIVILEGES ON SCHEMA public FROM diku_mod_users;
REVOKE CREATE ON SCHEMA public FROM PUBLIC;
SELECT * FROM pg_catalog.pg_namespace WHERE nspname = 'public';
```

```
 nspname | nspowner |                           nspacl
---------+----------+------------------------------------------------------------
 public  |       10 | {postgres=UC/postgres,=U/postgres}
```

hjiebsco:
> REVOKE statements make an update on the pg_catalog.pg_namespace row, even if the update doesn't change anything.

True but the select statements do not hold a lock and do not update pg_namespace, so those should not impact the revoke. The "tuple concurrently updated" error implies to me that multiple revoke statements interfere each other.

julianladisch:
Yes, I also don't completely understand what is going on.

This SELECT ... FOR UPDATE
`SELECT nspname INTO n FROM pg_catalog.pg_namespace WHERE nspname = 'public' FOR UPDATE;`
locks the record for all queries. This works even if we don't need to know which query interferes with our REVOKE.

hjiebsco:
How about we simplify it to this?
```
REVOKE ALL PRIVILEGES ON SCHEMA public FROM ${myuniversity}_${mymodule};

DO $$
BEGIN
  REVOKE CREATE ON SCHEMA public FROM PUBLIC;
EXCEPTION WHEN OTHERS THEN
  -- ignore; one of transactions has done the job
  NULL;
END $$;
```

julianladisch:
We also need to run
`REVOKE ALL PRIVILEGES ON SCHEMA public FROM ${myuniversity}_${mymodule};`
This is missing from your proposal.

We don't know whether the other transaction runs
`REVOKE CREATE ON SCHEMA public FROM PUBLIC;`
or some other query that affects the pg_catalog.pg_namespace. The other transaction has done some job, but it might be a different job.

julianladisch:
The logs posted on https://issues.folio.org/browse/RMB-744 show which of the two REVOKEs is affected:
```
REVOKE ALL PRIVILEGES ON SCHEMA public FROM diku_mod_users;
REVOKE ALL PRIVILEGES ON SCHEMA public FROM diku_mod_source_record_storage;
REVOKE ALL PRIVILEGES ON SCHEMA public FROM diku_mod_template_engine;
REVOKE ALL PRIVILEGES ON SCHEMA public FROM diku_mod_notify;
```
It is always the first one with `FROM ${myuniversity}_${mymodule}`, never the second one with `FROM public`
Wrapping the second into an exception-catch does not fix the issue with the first one.

julianladisch:
Always creating the advisory lock makes this compatible with the previous version released in RMB 31.1.5.

hjiebsco (commented):
Do we still need this PR? If yes, I am wondering if we can just re-try the `revoke` a few times with exponential backoff so it will have a better chance to succeed? Any fail-fast exception will put a burden on sysadmin to re-apply the module upgrade.

julianladisch:
This is a clean-up task. All modules properly work even if any REVOKE fails.

I suggest to merge this fix.

hjiebsco (commented):
I am fine with the change if tests have been done in both environments.

PR #806, opened 2020-11-11
RMB-388: Add getById tx and lock variants

hjiebsco:
See https://issues.folio.org/browse/RMB-388

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -2379,26 +2379,42 @@ public class PostgresClient {
 
   /**
    * Get the jsonb by id.
+   * @param conn  if provided, the connection on which to execute the query on.
+   * @param lock  whether to use SELECT FOR UPDATE to lock the selected row.
    * @param table  the table to search in
    * @param id  the value of the id field
    * @param function  how to convert the (String encoded) JSON
    * @param replyHandler  the result after applying function
    */
-  private <R> void getById(String table, String id,
+  private <R> void getById(final AsyncResult<SQLConnection> conn, boolean lock, String table, String id,
                            FunctionWithException<String, R, Exception> function,
                            Handler<AsyncResult<R>> replyHandler) {
-    getConnection(res -> {
-      if (res.failed()) {
-        replyHandler.handle(Future.failedFuture(res.cause()));
-        return;
+    Promise<PgConnection> promise = Promise.promise();
+    if (conn != null) {
+      if (conn.failed()) {
+        promise.fail(conn.cause());
+      } else {
+        promise.complete(conn.result().conn);
       }
-      PgConnection connection = res.result();
+    } else {
+      getConnection(res -> {
+        if (res.failed()) {
+          promise.fail(res.cause());
+        } else {
+          promise.complete(res.result());
+        }
+      });
+    }
+    promise.future().onFailure(ex -> replyHandler.handle(Future.failedFuture(ex)))
+    .onSuccess(connection -> {
       String sql = SELECT + DEFAULT_JSONB_FIELD_NAME
           + FROM + schemaName + DOT + table
           + WHERE + ID_FIELD + "= $1";
+      if (lock) {
+        sql += " FOR UPDATE";
+      }
       try {
         connection.preparedQuery(sql).execute(Tuple.of(UUID.fromString(id)), query -> {
-          connection.close();
           if (query.failed()) {
             replyHandler.handle(Future.failedFuture(query.cause()));
             return;
@@ -2418,6 +2434,10 @@ public class PostgresClient {
         });
       } catch (Exception e) {
         replyHandler.handle(Future.failedFuture(e));
+      } finally {
+        if (conn == null) {
+          connection.close();
+        }
       }
     });
   }
@@ -2429,7 +2449,34 @@ public class PostgresClient {
    * @param replyHandler  the result; the JSON is encoded as a String
    */
   public void getByIdAsString(String table, String id, Handler<AsyncResult<String>> replyHandler) {
-    getById(table, id, string -> string, replyHandler);
+    getByIdAsString(null, table, id, replyHandler);
+  }
+
+  /**
+   * Get the jsonb by id and return it as a String. Query will be executed on
+   * <code>conn</code> if provided.
+   * @param conn  if provided, the connection on which to execute the query on.
+   * @param table  the table to search in
+   * @param id  the value of the id field
+   * @param replyHandler  the result; the JSON is encoded as a String
+   */
+  public void getByIdAsString(AsyncResult<SQLConnection> conn,
+      String table, String id, Handler<AsyncResult<String>> replyHandler) {
+    getById(conn, false, table, id, string -> string, replyHandler);
+  }
+
+  /**
+   * Get the jsonb by id and return it as a String. Query will be executed on
+   * <code>conn</code> if provided. Selected row will be locked using
+   * <code>select ... for update<code> clause.
+   * @param conn  if provided, the connection on which to execute the query on.
+   * @param table  the table to search in
+   * @param id  the value of the id field
+   * @param replyHandler  the result; the JSON is encoded as a String
+   */
+  public void getByIdAsStringForUpdate(AsyncResult<SQLConnection> conn,
+      String table, String id, Handler<AsyncResult<String>> replyHandler) {
+    getById(conn, true, table, id, string -> string, replyHandler);
   }
 
   /**
@@ -2439,7 +2486,34 @@ public class PostgresClient {
    * @param replyHandler  the result; the JSON is encoded as a JsonObject
    */
   public void getById(String table, String id, Handler<AsyncResult<JsonObject>> replyHandler) {
-    getById(table, id, JsonObject::new, replyHandler);
+    getById(null, table, id, replyHandler);
+  }
+
+  /**
+   * Get the jsonb by id and return it as a JsonObject. Query will be executed on
+   * <code>conn</code> if provided.
+   * @param conn  if provided, the connection on which to execute the query on.
+   * @param table  the table to search in
+   * @param id  the value of the id field
+   * @param replyHandler  the result; the JSON is encoded as a JsonObject
+   */
+  public void getById(AsyncResult<SQLConnection> conn,
+      String table, String id, Handler<AsyncResult<JsonObject>> replyHandler) {
+    getById(conn, false, table, id, JsonObject::new, replyHandler);
+  }
+
+  /**
+   * Get the jsonb by id and return it as a JsonObject. Query will be executed on
+   * <code>conn</code> if provided. Selected row will be locked using
+   * <code>select ... for update<code> clause.
+   * @param conn  if provided, the connection on which to execute the query on.
+   * @param table  the table to search in
+   * @param id  the value of the id field
+   * @param replyHandler  the result; the JSON is encoded as a JsonObject
+   */
+  public void getByIdForUpdate(AsyncResult<SQLConnection> conn,
+      String table, String id, Handler<AsyncResult<JsonObject>> replyHandler) {
+    getById(conn, true, table, id, JsonObject::new, replyHandler);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2379,26 +2379,42 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
...
-    getConnection(res -> {
-      if (res.failed()) {
-        replyHandler.handle(Future.failedFuture(res.cause()));
-        return;
+    Promise<PgConnection> promise = Promise.promise();
+    if (conn != null) {
+      if (conn.failed()) {
+        promise.fail(conn.cause());
+      } else {
+        promise.complete(conn.result().conn);
```
julianladisch, line 2397:
```suggestion
      promise.handle(conn.map(sqlConnection -> sqlConnection.conn));
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2379,26 +2379,42 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
...
       }
-      PgConnection connection = res.result();
+    } else {
+      getConnection(res -> {
+        if (res.failed()) {
+          promise.fail(res.cause());
+        } else {
+          promise.complete(res.result());
+        }
+      });
```
julianladisch, line 2406:
```suggestion
      getConnection(promise);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2379,26 +2379,42 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
...
+    }
+    promise.future().onFailure(ex -> replyHandler.handle(Future.failedFuture(ex)))
```
julianladisch, line 2408:
```suggestion
    promise.future()
    .onFailure(ex -> replyHandler.handle(Future.failedFuture(ex)))
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2379,26 +2379,42 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
 
   /**
    * Get the jsonb by id.
+   * @param conn  if provided, the connection on which to execute the query on.
```
julianladisch, line 2382: The description should explain all three cases:
* if provided, the connection to use
* if null, a new connection is created and automatically released afterwards
* if failed, the replyHandler is failed

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2379,26 +2379,42 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
...
+      });
+    }
+    promise.future().onFailure(ex -> replyHandler.handle(Future.failedFuture(ex)))
+    .onSuccess(connection -> {
       String sql = SELECT + DEFAULT_JSONB_FIELD_NAME
           + FROM + schemaName + DOT + table
           + WHERE + ID_FIELD + "= $1";
+      if (lock) {
+        sql += " FOR UPDATE";
+      }
```
julianladisch, line 2415:
```suggestion
          + WHERE + ID_FIELD + "= $1"
          + (lock ? " FOR UPDATE" : "");
```
Creating the String in one go allows the compiler to implement it using a StringBuilder.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2429,7 +2449,34 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
    * @param replyHandler  the result; the JSON is encoded as a String
    */
   public void getByIdAsString(String table, String id, Handler<AsyncResult<String>> replyHandler) {
-    getById(table, id, string -> string, replyHandler);
+    getByIdAsString(null, table, id, replyHandler);
+  }
+
+  /**
+   * Get the jsonb by id and return it as a String. Query will be executed on
+   * <code>conn</code> if provided.
```
julianladisch, line 2457: The sentence `Query will be executed on <code>conn</code> if provided.` duplicates the information provided in the `@paran` entry and should be removed. The javadoc should be concise.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2429,7 +2449,34 @@ private FacetManager buildFacetManager(CQLWrapper wrapper, QueryHelper queryHelp
...
+   */
+  public void getByIdAsString(AsyncResult<SQLConnection> conn,
+      String table, String id, Handler<AsyncResult<String>> replyHandler) {
+    getById(conn, false, table, id, string -> string, replyHandler);
+  }
+
+  /**
+   * Get the jsonb by id and return it as a String. Query will be executed on
+   * <code>conn</code> if provided. Selected row will be locked using
+   * <code>select ... for update<code> clause.
```
julianladisch, line 2471: The key information should be in the beginning:

`Lock the row using <code>select ... for update</code> and return jsonb as String.`

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -709,5 +715,96 @@ public void getTotalRecordsTest() {
...
+    when(mockPgConnection.preparedQuery(anyString())).thenReturn(mockPreparedQuery);
+    PgPool mockPgPool = mock(PgPool.class);
+    doAnswer(AdditionalAnswers.answerVoid((Handler<AsyncResult<PgConnection>> handler)
+        -> handler.handle(Future.succeededFuture(mockPgConnection))))
+        .when(mockPgPool).getConnection(any());
+    when(pc.getClient()).thenReturn(mockPgPool);
+    SQLConnection mockSQLConnection = new SQLConnection(mockPgConnection, null, null);
+    AsyncResult<SQLConnection> mockConn = Future.succeededFuture(mockSQLConnection);
+    // tests
+    pc.getByIdAsString(table, id, ar -> assertTrue(ar.succeeded()));
```
julianladisch, line 745: If the method calls the callback then `ar -> assertTrue(ar.succeeded())` will run.
If the method doesn't complete and doesn't call the callback then `ar -> assertTrue(ar.succeeded())` will NOT run but the unit test will succeed and does NOT report this incomplete execution.
Please use https://vertx.io/docs/vertx-unit/java/ or https://vertx.io/docs/vertx-junit5/java/ that provide test methods for such tests, for example testContext.asyncAssertSuccess() or vertxTestContext.succeeding().

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTest.java
```diff
@@ -709,5 +715,96 @@ public void getTotalRecordsTest() {
...
+        (Tuple tuple, Handler<AsyncResult<RowSet<Row>>> handler)
+        -> handler.handle(Future.succeededFuture(mockRowSet))))
+        .when(mockPreparedQuery).execute(any(Tuple.class), any());
+    when(mockRow.getValue(anyInt())).thenThrow(new RuntimeException("fail"));
+    pc.getByIdAsString(mockConn, table, id, ar -> assertTrue(ar.failed()));
+    pc.getByIdAsString(mockConn, table, "1", ar -> assertTrue(ar.failed()));
+  } 
+  
+  private void assertGetByIdAsString(AsyncResult<String> ar) {
+    assertTrue(ar.succeeded());
```
julianladisch, line 796: assertTrue(ar.succeeded())
should be handled by testContext.asyncAssertSuccess() or vertxTestContext.succeeding().
They also unwrap the result.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch, line 2415: Can you write a test with a real PostgreSQL (not a mock) to ensure that "FOR UPDATE" actually works:
On first connection run "SELECT ... FOR UPDATE".
On second connection run "SELECT" for that row, this will block until the first connection closes the transaction.
On first connection run "SELECT pg_sleep(0.1)", "UPDATE" the record, and COMMIT the transaction.
The second connection should return the new value from the UPDATE, not the old value.

hjiebsco:
@julianladisch, thanks for the feedback. Please review it again. I moved the mock test to a new file. I also added an IT test case. I tweaked your proposed IT test logic a tiny bit because `select ... for update` prevents concurrent update, not select.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1483,6 +1483,57 @@ public void getByIdAsString(TestContext context) {
...
+                new CQLWrapper(criterion), true, rows -> {
+              assertSuccess(context, rows);
+              context.assertFalse(async2.isCompleted());
+              Row row = rows.result().iterator().next();
+              context.assertEquals(id, row.getUUID("id").toString());
+              async1.complete();
+              postgresClient.endTx(conn, done -> {
+                postgresClient.getByIdAsString(FOO, id, get1 -> {
+                  assertSuccess(context, get1);
+                  context.assertTrue(get1.result().contains("\"x1\""), get1.result());
```
julianladisch, line 1522: This getByIdAsString should go before the endTx to avoid any race condition.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1483,6 +1483,57 @@ public void getByIdAsString(TestContext context) {
...
+          });
+          // update first because it holds the lock
+          postgresClient.execute(conn, "SELECT pg_sleep(0.5)", delayRs -> {
+            assertSuccess(context, delayRs);
+            postgresClient.update(conn, FOO, new StringPojo("x1"),
+                new CQLWrapper(criterion), true, rows -> {
+              assertSuccess(context, rows);
+              context.assertFalse(async2.isCompleted());
+              Row row = rows.result().iterator().next();
+              context.assertEquals(id, row.getUUID("id").toString());
```
julianladisch, line 1517: This assertEquals is not needed, we can assume that postgresClient.update works correctly.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1483,6 +1483,57 @@ public void getByIdAsString(TestContext context) {
...
+                  assertSuccess(context, get1);
+                  context.assertTrue(get1.result().contains("\"x1\""), get1.result());
+                });
+              });
+            });
+          });
+          // before update
+          postgresClient.getByIdAsString(FOO, id, get0 -> {
+            assertSuccess(context, get0);
+            context.assertTrue(get0.result().contains("\"x0\""), get0.result());
```
julianladisch, line 1530: This getByIdAsString with assert should be removed, we can assume that save works correctly.

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
```diff
@@ -1483,6 +1483,57 @@ public void getByIdAsString(TestContext context) {
...
+          postgresClient.update(FOO, new StringPojo("x2"), id, rows -> {
+            context.assertTrue(async1.isCompleted());
+            postgresClient.getByIdAsString(FOO, id, get2 -> {
+              assertSuccess(context, get2);
+              context.assertTrue(get2.result().contains("\"x2\""));
+              async2.complete();
+            });
+          });
+          // update first because it holds the lock
+          postgresClient.execute(conn, "SELECT pg_sleep(0.5)", delayRs -> {
```
julianladisch, line 1510: I don't think that we need the pg_sleep. Instead we can simply run "SELECT 1" to make sure that the update x2 has reached the database and is blocked before we run update x1.

julianladisch (approved):
Looks good, thanks!

domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
hjiebsco: Can you clarify how to use `SELECT 1` to make sure that the update x2 has reached the db? We could query system tables to check locks (if that is what you mean) but it is probably overkill than `pg_sleep`.

PR #807, opened 2020-11-17
Async Tenant API RMB-754

adamdickmeiss:
Property generate_routing_context property is evil.


domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -43,99 +45,16 @@
...
-          .respond500WithTextPlain(e.getMessage())));
-      }
-    });
-  }
-
   Future<Boolean> tenantExists(Context context, String tenantId) {
     Promise<Boolean> promise = Promise.promise();
-    tenantExists(context, tenantId, promise.future());
+    tenantExists(context, tenantId, promise::handle);
     return promise.future();
```
julianladisch, line 58: `::handle` is not needed:

    tenantExists(context, tenantId, promise);

This short form is also possible:

    return Future.future(promise -> tenantExists(context, tenantId, promise));

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -309,54 +201,115 @@ public String sqlFile(String tenantId, boolean tenantExists, TenantAttributes te
...
-      log.debug("upgrade from " + tenantAttributes.getModuleFrom() + " to " + tenantAttributes.getModuleTo());
+    String id = UUID.randomUUID().toString();
+    TenantJob job = new TenantJob();
+    job.setId(id);
+    job.setTenant(tenantId);
+    job.setTenantAttributes(tenantAttributes);
+    job.setComplete(false);
+
+    jobs.put(id, job);
+    String location = (routingContext != null ? routingContext.request().uri() : "") + "/" + id;
```
julianladisch, line 221: POST https://example.com:9876/_/tenant
with a
Location: /12345
reply will result in the wrong
https://example.com:9876/12345
```suggestion
    String location = (routingContext != null ? routingContext.request().uri() : "/_/tenant") + "/" + id;
```
https://tools.ietf.org/html/rfc7231#section-7.1.2
https://tools.ietf.org/html/rfc3986#section-4.2
https://tools.ietf.org/html/rfc3986#section-5.4

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -309,54 +201,115 @@ public String sqlFile(String tenantId, boolean tenantExists, TenantAttributes te
...
+          }
+        })
+        .compose(res -> {
+          if (tenantAttributes == null) {
+            return Future.succeededFuture();
+          } else if (Boolean.TRUE.equals(tenantAttributes.getPurge())) {
+            PostgresClient.closeAllClients(tenantId);
+            return Future.succeededFuture();
+          } else {
+            return loadData(tenantAttributes, tenantId, headers, context);
```
julianladisch, line 243:
```suggestion
          }
          if (tenantAttributes == null) {
            return Future.succeededFuture();
          }
          if (Boolean.TRUE.equals(tenantAttributes.getPurge())) {
            PostgresClient.closeAllClients(tenantId);
            return Future.succeededFuture();
          }
          return loadData(tenantAttributes, tenantId, headers, context);
```

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
adamdickmeiss: ok. Thx

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
adamdickmeiss: No. the uri() returns ` /_/tenant`.

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
adamdickmeiss: Thx. That's easier to read.

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
adamdickmeiss: The routingContext == null case is only for non-HTTP testing. There is obviously no test for Location in that case.

adamdickmeiss: Actually, I'll see if I can pass some (non-null) routingContext in tests. No need to have production code that checks for null routingContext anyway.

julianladisch: Yes, the only change that I suggest is the `routingContext == null` case.

adamdickmeiss: ok. Since you insist on hard-coding the path in the case that is not used in production code, I'll just switch to not use RoutingContext. That also avoids the generate_routing context property which, as I've mentioned before, is evil :-)

README.md
```diff
@@ -1327,54 +1331,24 @@ public class MyTenantAPI extends TenantAPI {
...
-    }
-    ...
-
-    handler.handle(res);  // HTTP status: 200 for upgrade, 201 for install
-  }, context);
+Future<Void> loadData(TenantAttributes attributes, String tenantId, Map<String, String> headers,
+  Context verxContext) {
+  Future<Void> future = super.loadData(attributes, tenantId, headers, vertxContext).compose(res -> {
+    return Future...;
+  });
```
julianladisch, line 1351:
```suggestion
  return super.loadData(attributes, tenantId, headers, vertxContext).compose(res -> {
    // load data
    return Future...;
  });
}
```

README.md
```diff
@@ -1387,36 +1361,32 @@ import org.folio.rest.tools.utils.TenantLoading;
...
-public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers,
-    Handler<AsyncResult<Response>> handler, Context context) {
-  super.postTenant(tenantAttributes, headers, res -> {
-    if (res.failed()) {
-      handler.handle(res);
-      return;
-    }
+Future<Void> loadData(TenantAttributes attributes, String tenantId, Map<String, String> headers,
+  Context verxContext) {
+  Future<Void> future = super.loadData(attributes, tenantId, headers, vertxContext).compose(res -> {
```
julianladisch, line 1366:
```suggestion
  return super.loadData(attributes, tenantId, headers, vertxContext).compose(res -> {
```

README.md
```diff
@@ -1387,36 +1361,32 @@ import org.folio.rest.tools.utils.TenantLoading;
...
-        handler.handle(Future.succeededFuture(PostTenantResponse
-          .respond500WithTextPlain(res1.cause().getMessage())));
+        promise.fail(res1.cause().getMessage());
         return;
       }
-      handler.handle(res);  // HTTP status: 200 for upgrade, 201 for install
+      promise.complete(); // OK
     });
-  }, context);
+    return promise.future();
```
julianladisch, line 1383: More concise when using Future.future(promise -> ...):
```
    return Future.future(promise -> tl.perform(tenantAttributes, headers, vertx, promise)
        .mapEmpty();
```

README.md
```diff
@@ -1624,7 +1594,13 @@ When upgrading a module via the Tenant API, an index is deleted if either `"tOps
...
 ##### Posting information
 
-Posting a new tenant must include a body. The body should contain a JSON conforming to the [moduleInfoSchema](https://github.com/folio-org/raml/blob/master/schemas/moduleInfo.schema) schema. The `module_to` entry is mandatory, indicating the version module for this tenant. The `module_from` entry is optional and indicates an upgrade for the tenant to a new module version.
+Posting a new tenant must include a body. The body should contain a JSON
+conforming to the
+[tenantAttributes](https://github.com/folio-org/raml/blob/master/schemas/tenantAttributes.schema)
+schema. The `module_to` entry is optional, indicating the version module for
+this tenant. If `module_to` is omitted, module is disabled for the tenant. The
+`module_from` entry is optional and indicates an upgrade for the tenant to a
+new module version or disable if `module_to` is omitted.
```
julianladisch, line 1603:
```suggestion
schema.

To enable a new module indicate the version module in `module_to` and omit `module_from`.
To upgrade a module indicate the existing version module in `module_from` and the new version module in `module_to`.
To disable a module indicate the existing version module in `module_from` and omit `module_to`.
```

README.md
adamdickmeiss: Doesn't work for me.. Is this correct?

```
   Promise<Integer> promise = Promise.promise();
   tl.perform(attributes, headers, vertx, promise);
   return promise.future().mapEmpty();
```

How can that be turned into a one-liner?

julianladisch: Yes, done.

julianladisch (approved):
Looks good, thanks!

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
```diff
@@ -392,9 +392,8 @@ private static void loadURL(Map<String, String> headers, URL url,
    * @param vertx Vertx handle to be used (for spawning HTTP clients)
    * @return Success or failure
    */
-  public Future<Void> perform(TenantAttributes ta, Map<String, String> headers,
-      Context context) {
-    return perform0(ta, headers, context.owner()).mapEmpty();
+  public Future<Void> perform(TenantAttributes ta, Map<String, String> headers, Vertx vertx) {
+    return perform0(ta, headers, vertx).mapEmpty();
```
julianladisch, line 396: The `Future<Void> perform` method should be designed to match the types used in TenantRef loadData method:
```
Future<Void> loadData(TenantAttributes attributes, String tenantId,
                      Map<String, String> headers, Context vertxContext)
```
The Future type should be Void and not Integer.
It should take a Context and not a Vertx. (The javadoc in line 392 should be changed.)
This will result in the most straightforward usage as explained in the documentation:
https://github.com/folio-org/raml-module-builder/pull/807/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R1377
https://github.com/folio-org/raml-module-builder/pull/807/files#diff-95cca51db9a9a6cf4c286410f08fe5eec3c68456869faf08739d8b0235fdedd9R71
We don't want all the modules using the TenantLoading loadData interface to duplicate the code that extracts the Vertx from vertxContext. This task should be handled by the perform method. The other perform method that has been used for postTenant is legacy and is not changed to avoid a breaking change: it takes a `Promise<Integer>` (not `Promise<Void>`) and takes a Vertx (not a vertx Context).

adamdickmeiss: OK. Fine as far as Context goes.. I'll change it to take that.. But isn't it bad if we throw away the Integer in the result. At least it allows a user of the API to determine the number of items loaded.

julianladisch (approved):

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
julianladisch: I've made the requested change: Not throwing away the number of records loaded but returning them as `Future<Integer>`. There isn't any module that uses this number yet, though.

PR #811, opened 2020-11-30
RMB-696: maxPoolSize parallel run test

julianladisch:


adamdickmeiss (approved):
I tested with the RMB-609 branch. Passes there too.

PR #812, opened 2020-12-03
Default log4j2 logging is line based; update documentation RMB-768

adamdickmeiss:
Notice in properties:

```
packages = org.folio.okapi.common.logging
```

it was also part of RMB-609 merge. Is that OK?


julianladisch:
The https://wiki.folio.org/display/DD/Folio+logging+solution says:
> For any logging aggregation solution json format of logs is preferable over plain text in terms of performance and simplicity of parsing. All Folio modules should have the same json format for logs

My interpretation is:
The JSON log format should be the default for running the main code.
For unit tests line based logs can still be used to help developers during development.

adamdickmeiss:
Is that still what the DevOps guys prefers? I most certainly heard no objection when we changed Okapi to use the line based logging by default. Then again, Okapi is not a regular module.

adamdickmeiss:
This what it looks like now - as a developer. https://jenkins-aws.indexdata.com/blue/organizations/jenkins/folio-org%2Fmod-configuration/detail/MODCONF-68-upgrade-to-vertx-4-0-0/7/pipeline/49

I think that's not going to be productive going forward - as a developer. Folio platform can of course start modules in another fashion .. For example by just pointing to the JSON properties file.

julianladisch:
https://jenkins-aws.indexdata.com/blue/organizations/jenkins/folio-org%2Fmod-configuration/detail/MODCONF-68-upgrade-to-vertx-4-0-0/7/pipeline/49 shows running unit tests.
/src/test/resources/log4j2-test.properties has precedence over /src/main/resources/log4j2.properties:
https://logging.apache.org/log4j/2.x/manual/configuration.html
I suggest to use JSON log format in log4j2.properties for production and line base log format in log4j2-test.properties for unit tests.

adamdickmeiss:
I wouldn't suggest that. But then again.. I'm not running production. I guess we just introduce structure logging and then see what the programmers and production guys will react. So far, no feedback.

adamdickmeiss:
FWIW, MODINVSTOR-624 branch you exactly did what I think we should be doing..
https://github.com/folio-org/mod-inventory-storage/pull/546/commits/e350ce9a794e525b5360c7b85d630db1f95963a7
log4j2.properties with default properties always..
and log4j2-json.properties for JSON structured logging.

julianladisch (approved):

hjiebsco (approved):

craigmcnally:
I think we should update [the wiki](https://wiki.folio.org/display/DD/Folio+logging+solution) to reflect the decision made here

PR #816, opened 2020-12-10
RMB-727 Initial support of optimistic locking

hjiebsco:
Initial optimistic locking implementation based on RMB-727. Things that deserve more thinking and discussion:
- The proposed "version" was changed to "_version" during implementation because some modules already defined version property. If not OK, it can be easily changed back.
- The proposed changes to PgUtils require 409 response to be defined. The PR will check 409 first, if not available, fall back to use 400.


domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1022,6 +1041,8 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
       return;
     }
 
+    final Method respond409 = getRespond409(responseClass);
```
julianladisch, line 1044: This should go into the following try block, sorted between respond201 and respond413.
Only if all code is in try blocks we can assure that all exceptions are properly returned via the AsyncResult Handler.

hjiebsco: > This should go into the following try block, sorted between respond201 and respond413.
> Only if all code is in try blocks we can assure that all exceptions are properly returned via the AsyncResult Handler.

Done.

domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
```diff
@@ -0,0 +1,49 @@
...
+        <#else>
+          NEW.jsonb = jsonb_set(NEW.jsonb, '{${ol_version}}', to_jsonb(1));
+        </#if>
+      WHEN 'UPDATE' THEN
+        <#if table.withOptimisticLocking.name() == "OFF">
+          NEW.jsonb = NEW.jsonb - '${ol_version}';
+        <#else>
+          IF NEW.jsonb->'${ol_version}' IS DISTINCT FROM OLD.jsonb->'${ol_version}' THEN
+            <#if table.withOptimisticLocking.name() == "FAIL">
+              <#assign ol_notice_level = "EXCEPTION">
```
julianladisch, line 22: The two lines with `<#assign ol_notice_level` should be merged into one if-else clause for readability.

domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
```diff
@@ -0,0 +1,49 @@
...
+            </#if>
+            RAISE ${ol_notice_level} 'Cannot update record % because it has been changed: '
```
julianladisch, line 24: Can you also set some error code, for example 23F09 (23 for class 23 — Integrity Constraint Violation, F for FOLIO, 09 for 409 HTTP status code), to make it more easy to detect this error or warning?
https://www.postgresql.org/docs/current/errcodes-appendix.html

domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerIT.java
```diff
@@ -389,4 +391,46 @@ public void replacePublicSchemaFunctions(TestContext context) throws Interrupted
...
+    assertThat("tab_ol_fail has version 1", selectInteger(context, String.format(sql, olVersion, "tab_ol_fail")), is(1));
+    assertThat("tab_ol_none has no version", selectInteger(context, String.format(sql, olVersion, "tab_ol_none")), is(0));
+    // test update
+    tables.forEach(table -> {
+      String updateSql = "UPDATE " + table + " SET jsonb=jsonb_set(jsonb, '{" + olVersion + "}', to_jsonb('2'::text))";
+      if (table.equals("tab_ol_fail")) {
+        // fail on conflict
+        executeAndExpectFailure(context, updateSql, "Cannot update record", "because it has been changed");
+      } else {
+        execute(context, "UPDATE " + table + " SET jsonb=jsonb_set(jsonb, '{" + olVersion + "}', to_jsonb('2'::text))");
```
julianladisch, line 418: Can you replace it by `updateSql`?

domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
hjiebsco: > The two lines with `<#assign ol_notice_level` should be merged into one if-else clause for readability.

Done.

domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerIT.java
hjiebsco: > Can you replace it by `updateSql`?

Done.

domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
hjiebsco: > Can you also set some error code, for example 23F09 (23 for class 23 — Integrity Constraint Violation, F for FOLIO, 09 for 409 HTTP status code), to make it more easy to detect this error or warning?
> https://www.postgresql.org/docs/current/errcodes-appendix.html

Good suggestion. Done.

README.md
```diff
@@ -1480,6 +1480,10 @@ For each **table** in `tables` property:
 14. `deleteFields` / `addFields` - delete (or add with a default value), a field at the specified path for all JSON entries in the table
 15. `populateJsonWithId` - This schema.json entry and the disable option is no longer supported. The primary key is always copied into `jsonb->'id'` on each insert and update.
 16. `pkColumnName` - No longer supported. The name of the primary key column is always `id` and is copied into `jsonb->'id'` in each insert and update. The method PostgresClient.setIdField(String) no longer exists.
+17. `withOptimisticLocking` - Creates insert/update trigger to auto populate/update `_version` field in the json
+    * `off` Optimistic Locking is disabled. The trigger does nothing
+    * `logOnConflict` Optimistic Locking is enabled. Version conflict info is logged as a warning message
+    * `failOnConflict` Optimistic Locking is enabled. Version conflict will fail the transaction
```
julianladisch, line 1486: We have this paragraph and the paragraph at line 1818 below about optimistic locking. Both are lacking some information. To avoid repetition I suggest to only list the options here

    17. `withOptimisticLocking` - `off` (default), `logOnConflict`, or `failOnConflict`, for details see [optimistic locking section](#optimistic-locking) below

and explain all details below.

README.md
```diff
@@ -1809,6 +1813,10 @@ Replace 1000 by `exactCount` if configured differently.
 
 RMB is aware of the [metadata.schema](https://github.com/folio-org/raml/blob/raml1.0/schemas/metadata.schema). When a request (POST / PUT / PATCH) comes into an RMB module, RMB will check if the passed-in JSON's schema declares a reference to the metadata schema. If so, RMB will populate the JSON with a metadata section with the current user and the current time. RMB will set both update and create values to the same date/time and to the same user, as accepting this information from the request may be unreliable. The module should persist the creation date and the created by values after the initial POST. For an example of this using SQL triggers see [metadata.ftl](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/resources/templates/db_scripts/metadata.ftl). Add [withMetadata to the schema.json](https://github.com/folio-org/raml-module-builder#the-post-tenant-api) to create that trigger.
 
+## Optimistic Locking
```
julianladisch, line 1816: Can you add this to the table of contents?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgExceptionUtil.java
```diff
@@ -65,6 +66,17 @@ public static boolean isInvalidTextRepresentation(Throwable throwable) {
     return INVALID_TEXT_REPRESENTATION.equals(get(throwable, 'C'));
   }
 
+  /**
+   * Check for version conflict.
```
julianladisch, line 70:
```suggestion
   * Check for optimistic locking version conflict.
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -971,7 +973,12 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
       PostgresClient postgresClient = postgresClient(vertxContext, okapiHeaders);
       postgresClient.update(table, entity, id, reply -> {
         if (reply.failed()) {
-          asyncResultHandler.handle(response(table, id, reply.cause(), clazz, respond400, respond500));
+          if (PgExceptionUtil.isVersionConflict(reply.cause())) {
+            Method method = respond409 == null ? respond400 : respond409;
+            asyncResultHandler.handle(response(reply.cause().getMessage(), method, method));
```
julianladisch, line 978:
```suggestion
            asyncResultHandler.handle(response(reply.cause().getMessage(), method, respond500));
```
If creating a respond400 or respond409 throws an exception we need to fall-back to respond500.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -1036,8 +1056,13 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
       PostgresClient postgresClient = postgresClient(vertxContext, okapiHeaders);
       Handler<AsyncResult<RowSet<Row>>> replyHandler = result -> {
         if (result.failed()) {
-          asyncResultHandler.handle(response(table, /* id */ "", result.cause(),
-              responseClass, respond500, respond500));
+          if (PgExceptionUtil.isVersionConflict(result.cause())) {
+            Method method = respond409 == null ? respond400 : respond409;
+            asyncResultHandler.handle(response(result.cause().getMessage(), method, method));
```
julianladisch, line 1061:
```suggestion
            asyncResultHandler.handle(response(result.cause().getMessage(), method, respond500));
```

README.md
```diff
@@ -1480,6 +1480,10 @@ For each **table** in `tables` property:
 14. `deleteFields` / `addFields` - delete (or add with a default value), a field at the specified path for all JSON entries in the table
 15. `populateJsonWithId` - This schema.json entry and the disable option is no longer supported. The primary key is always copied into `jsonb->'id'` on each insert and update.
 16. `pkColumnName` - No longer supported. The name of the primary key column is always `id` and is copied into `jsonb->'id'` in each insert and update. The method PostgresClient.setIdField(String) no longer exists.
+17. `withOptimisticLocking` - Creates insert/update trigger to auto populate/update `_version` field in the json
+    * `off` Optimistic Locking is disabled. The trigger does nothing
```
julianladisch, line 1484:
```suggestion
    * `off` Optimistic Locking is disabled. The trigger is removed if it existed before.
```

domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
```diff
@@ -0,0 +1,50 @@
...
+  RETURNS trigger AS $$
+  BEGIN
+    CASE TG_OP
+      WHEN 'INSERT' THEN
+        <#if table.withOptimisticLocking.name() != "OFF">
+          NEW.jsonb = jsonb_set(NEW.jsonb, '{${ol_version}}', to_jsonb(1));
+        </#if>
+      WHEN 'UPDATE' THEN
+        <#if table.withOptimisticLocking.name() == "OFF">
+          NEW.jsonb = NEW.jsonb - '${ol_version}';
```
julianladisch, line 16: Removing the version does not match the documentation.
If optimistic locking is off the trigger should be removed if it exists.

domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerIT.java
```diff
@@ -389,4 +391,50 @@ public void replacePublicSchemaFunctions(TestContext context) throws Interrupted
...
+    tables.forEach(table -> {
+      String updateSql = "UPDATE " + table + " SET jsonb=jsonb_set(jsonb, '{" + olVersion + "}', to_jsonb('2'::text))";
+      if (table.equals("tab_ol_fail")) {
+        // fail on conflict
+        executeAndExpectFailure(context, updateSql, "Cannot update record", "because it has been changed");
+      } else {
+        execute(context, updateSql);
+      }
+    });
+    assertThat("tab_ol_off has no version", selectInteger(context, String.format(sql, olVersion, "tab_ol_off")), is(0));
```
julianladisch, line 421: This does not match the documentation.

domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerIT.java
```diff
@@ -389,4 +391,50 @@ public void replacePublicSchemaFunctions(TestContext context) throws Interrupted
...
+    assertThat("tab_ol_log has version 1", selectInteger(context, String.format(sql, olVersion, "tab_ol_log")), is(2));
+    assertThat("tab_ol_fail has version 1", selectInteger(context, String.format(sql, olVersion, "tab_ol_fail")), is(1));
+    // update version as provided if table has no optimistic locking configuration
+    assertThat("tab_ol_none has no version", selectInteger(context, String.format(sql, olVersion, "tab_ol_none")), is(2));
+    // test insert with OFF
+    execute(context, "DELETE from tab_ol_off");
+    execute(context, "INSERT INTO tab_ol_off SELECT md5('abc')::uuid, jsonb_build_object('" + olVersion + "', 5)");
+    assertThat("tab_ol_off ignore version", selectInteger(context, String.format(sql, olVersion, "tab_ol_off")), is(5));
+  }
```
julianladisch, line 430: You may want to add a `runSchema(context, TenantOperationUPDATE, "schemaWithOptimisticLocking2.json")` that sets optimistic locking to off for "tab_ol_fail" and check that it is really off.

README.md
```diff
@@ -1480,6 +1480,10 @@ For each **table** in `tables` property:
...
+    * `logOnConflict` Optimistic Locking is enabled. Version conflict info is logged as a warning message
```
julianladisch, line 1485: logOnConflict documentation needs some explanation how RMB based modules need to configure slf4j/log4j2 to get this logged. You can handle this within this PR or create a follow-up Jira for this.

julianladisch: The SQL error code 23F09 will be logged, this should be mentioned here to help filtering the logs.

README.md
hjiebsco: > We have this paragraph and the paragraph at line 1818 below about optimistic locking. Both are lacking some information. To avoid repetition I suggest to only list the options here
> 
> ```
> 17. `withOptimisticLocking` - `off` (default), `logOnConflict`, or `failOnConflict`, for details see [optimistic locking section](#optimistic-locking) below
> ```
> 
> and explain all details below.

Done.

README.md
hjiebsco: > Can you add this to the table of contents?

Done.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
hjiebsco: > If creating a respond400 or respond409 throws an exception we need to fall-back to respond500.

Done.

domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
hjiebsco: > Removing the version does not match the documentation.
> If optimistic locking is off the trigger should be removed if it exists.

Did not know we should remove the trigger when config is changed to `off`. Made the change as suggested now. Technically speaking, since this `off` is the same as the config is not defined, maybe we should simplify it by removing this `off` option.

domain-models-runtime/src/test/java/org/folio/rest/persist/ddlgen/SchemaMakerIT.java
hjiebsco: > You may want to add a `runSchema(context, TenantOperationUPDATE, "schemaWithOptimisticLocking2.json")` that sets optimistic locking to off for "tab_ol_fail" and check that it is really off.

Added test as suggested.

README.md
hjiebsco: > logOnConflict documentation needs some explanation how RMB based modules need to configure slf4j/log4j2 to get this logged. You can handle this within this PR or create a follow-up Jira for this.

Not sure what config needs to be tweaked, but I was able to see the WARN message. I added a sample to the doc.

hjiebsco:
Based on the new PR feedback, looks like we should remove `off` option. It has the same behavior as removing `withOptimisticLocking`.

README.md
```diff
@@ -1809,6 +1811,18 @@ Replace 1000 by `exactCount` if configured differently.
...
+
+RMB supports optimistic locking. By default it is disabled. Module developer can enable it by adding attribute `withOptimisticLocking` to the table definition in schema.json. The available options are listed below. When either `failOnConflict` or `logOnConflict` attribute are specified, a database trigger will be created to auto populate/update `_version` field in json on insert/update. Note, `_version` field has to be defined in json to make this work. The version conflict error will be reported as 419 HTTP response code if 419 is defined in RAML, otherwise, it will fall back to use 400 HTTP response code.
+
+    * `off` Optimistic Locking is disabled. The trigger is removed if it existed before.
+    * `failOnConflict` Optimistic Locking is enabled. Version conflict will fail the transaction
+    * `logOnConflict` Optimistic Locking is enabled. Version conflict info is logged as a warning message with a customized SQL error code 23F09. A sample log entry looks like below:
+
+```
+Backend notice: severity='WARNING', code='23F09', message='Cannot update record 57db089f-18e4-7815-55d5-4cc6607e9059 because it has been changed: Stored _version is 1, _version of request is "2"' ...
```
julianladisch, line 1823: Showing an example log entry is a good idea.
The version should show the common case where the request version is older than the stored version:
```suggestion
Backend notice: severity='WARNING', code='23F09', message='Cannot update record 57db089f-18e4-7815-55d5-4cc6607e9059 because it has been changed: Stored _version is 2, _version of request is "1"' ...
```

julianladisch (approved):
Looks good, thanks!

julianladisch:
We should keep the `off` option. At some later time and after testing we might want to set the default to warn or fail. It is a common configuration pattern that there is a possibility to explicitly write the default configuration option.

README.md
hjiebsco: > Showing an example log entry is a good idea.
> The version should show the common case where the request version is older than the stored version:

Good catch. Done.

PR #817, opened 2020-12-11
RMB-389 PgUtil.delete by CQL

MikhailFokanov:
Test will be implemented in future commits.


adamdickmeiss:
What's the 200 message content going to be? Should it be 204 instead?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -745,6 +745,88 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
     }
   }
 
+   /**
+   * Delete records by CQL.
+   * @param table  the table that contains the records
+   * @param cql  the CQL query for filtering and sorting the records
```
julianladisch, line 751: SQL DELETE doesn't have an ORDER BY clause.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -745,6 +745,88 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+   * @param offset number of records to skip, use 0 or negative number for not skipping
```
julianladisch, line 752: SQL DELETE doesn't have an OFFSET clause.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -745,6 +745,88 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+   * @param limit maximum number of records to return, use a negative number for no limit
```
julianladisch, line 753: SQL DELETE doesn't have a LIMIT clause.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -745,6 +745,88 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+   * @param okapiHeaders  http headers provided by okapi
+   * @param vertxContext  the current context
+   * @param responseDelegateClass  the ResponseDelegate class generated as defined by the RAML file,
+   *    must have these methods: respond200(C), respond400WithTextPlain(Object), respond500WithTextPlain(Object).
```
julianladisch, line 757: Can you explain in the javadoc what content the JSON will have that is returned along with the HTTP 200 status code on success?
If we don't want to return anything we can use the 204 status code.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -745,6 +745,88 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+    final Method respond500;
+    final Method respond400;
+    try {
+      respond500 = responseDelegateClass.getMethod(RESPOND_500_WITH_TEXT_PLAIN, Object.class);
+      respond400 = responseDelegateClass.getMethod(RESPOND_400_WITH_TEXT_PLAIN, Object.class);
+    } catch (Exception e) {
+      logger.error(e.getMessage(), e);
+      asyncResultHandler.handle(response(e.getMessage(), null, null));
+      return;
+    }
```
julianladisch, line 776: This code (respond500, respond400) is not needed and should be removed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -745,6 +745,88 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+    } catch (FieldException e) {
+      logger.error(e.getMessage(), e);
+      asyncResultHandler.handle(response(e.getMessage(), respond400, respond500));
+    }
+  }
+
+    static void delete(PreparedCQL preparedCql,
+      Map<String, String> okapiHeaders, Context vertxContext,
+      Class<? extends ResponseDelegate> responseDelegateClass,
+      Handler<AsyncResult<Response>> asyncResultHandler) {
```
julianladisch, line 792: These two delete methods should be merged.
The 3 lines with cql2pgJson,  cqlWrapper and preparedCql can be moved just before PostgresClient line.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,14 @@ public void optimizedSQLwithNo500(TestContext testContext) {
         }));
   }
 
+  @Test
+  public void deleteByCQLwithNo500(TestContext testContext) {
+    PgUtil.delete("users",  "username=delete_test ",
+        okapiHeaders, vertx.getOrCreateContext(), ResponseWithout500.class, testContext.asyncAssertFailure(e -> {
+          assertThat(e, is(instanceOf(NullPointerException.class)));
+        }));
+  }
```
julianladisch, line 1116: Unit tests for the other parts of the new code are missing.

adamdickmeiss (changes requested):
You should be able to increase code coverage a bit. I'm sure there are unit tests already you can use as inspiration which provoke all kinds of exceptions. (deleteById ... ones)

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -751,6 +751,68 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
+          asyncResultHandler.handle(response(e.getMessage(), respond500, respond500));
+        }
+      } catch (FieldException e) {
+        logger.error(e.getMessage(), e);
+        asyncResultHandler.handle(response(e.getMessage(), respond400, respond500));
+      }
+    } catch (Exception e) {
+      logger.error(e.getMessage(), e);
+      asyncResultHandler.handle(response(e.getMessage(), null, null));
+    }
```
julianladisch, line 813: Nesting try-catch clauses should be avoided. See the other methods how this can be accomplished.
The second and third try clause can be combined by using one try with two catch clauses.

julianladisch: See also https://sonarcloud.io/organizations/folio-org/rules?languages=java&open=java%3AS1141&q=try

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
         }));
   }
 
+   @Test
```
julianladisch, line 1110:
```suggestion
  @Test
```

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
...
+  public void deleteByCQLwithNo500(TestContext testContext) {
```
julianladisch, line 1111: All the deleteByCQL tests should go to line 404, right after the deleteById tests, and not between the optimized sql tests.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
...
+    PostgresClient pg = PostgresClient.getInstance(vertx, "testtenant");
+    insert(testContext, pg, "delete_test",  1);
+    PgUtil.delete("users",  "username=delete_test",
+        okapiHeaders, vertx.getOrCreateContext(), ResponseWithout500.class, testContext.asyncAssertFailure(e -> {
+          assertThat(e, is(instanceOf(NullPointerException.class)));
+        }));
+  }
+
+  @Test
+  public void deleteByCQLwithNo204(TestContext testContext) {
```
julianladisch, line 1121: The name doesn't match the test.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
         }));
   }
 
+   @Test
+  public void deleteByCQLwithNo500(TestContext testContext) {
+    PostgresClient pg = PostgresClient.getInstance(vertx, "testtenant");
+    insert(testContext, pg, "delete_test",  1);
```
julianladisch, line 1113: This setup is not needed. See deleteByIdResponseWithout500 for an example.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
...
+    PgUtil.delete("users",  "username=delete_test",
+        okapiHeaders, vertx.getOrCreateContext(), ResponseWithout500.class, testContext.asyncAssertFailure(e -> {
+          assertThat(e, is(instanceOf(NullPointerException.class)));
```
julianladisch, line 1116: The assert should check that the message contains an indication the that respond500WithTextPlain method is missing. See deleteByIdResponseWithout500 for an example.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
...
+
+  @Test
+  public void deleteByCQLwithNo204(TestContext testContext) {
+    String uuid = randomUuid();
+    post(testContext, "Pippilotta", uuid, 201);
+
+    PgUtil.delete("users", "username=delete_test",
+        okapiHeaders, vertx.getOrCreateContext(),
+        Users.PutUsersByUserIdResponse.class,
+        asyncAssertSuccess(testContext, 204));
```
julianladisch, line 1128: This only checks that the response code is 204.
It should also check that PgUtil.delete correctly deletes the records filtered by CQL.
The test case should start with at least 3 records and use a CQL that deletes at least 2 records but not all records.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1107,6 +1107,49 @@ public void optimizedSQLwithNo500(TestContext testContext) {
...
+            testContext.fail("Expected status 500, got "
+                + response.getStatus() + " " + response.getStatusInfo().getReasonPhrase());
+            async.complete();
+            return;
+          }
+
+          async.complete();
+    }));
+    async.awaitSuccess(10000 /* ms */);
+  }
```
julianladisch, line 1150: Please remove all unnecessary test setup. See deleteByIdResponseWithout500 and deleteByIdResponseWithout204 for examples.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
julianladisch: https://issues.folio.org/browse/DEBT-3 requests 11 APIs where this PgUtil.delete will be used, more to come.
Therefore we should have 100 % code coverage. Examples how to do this are the other unit tests in PgUtilIT.

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
adamdickmeiss: Yep. Done already.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
adamdickmeiss: Yes.. Much better! Done now

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
adamdickmeiss: done

adamdickmeiss (approved):

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -402,6 +402,109 @@ public void deleteByIdResponseWithout204(TestContext testContext) {
         asyncAssertSuccess(testContext, 500, "respond204"));
   }
 
+  @Test
+  public void deleteByCQLwithNo500(TestContext testContext) {
+    PostgresClient pg = PostgresClient.getInstance(vertx, "testtenant");
+    //  insert(testContext, pg, "delete_test",  1);
```
julianladisch, line 408: These two lines should be removed.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -402,6 +402,109 @@ public void deleteByIdResponseWithout204(TestContext testContext) {
...
+    insert(testContext, pg, "delete_test",  1);
+    PgUtil.delete("users",  "username==delete_test",
+        null, vertx.getOrCreateContext(), Users.DeleteUsersByUserIdResponse.class,
+        asyncAssertSuccess(testContext, 400, "null"));
+  }
+
+  @Test
+  public void deleteByCQLOK(TestContext testContext) {
+    PostgresClient pg = PostgresClient.getInstance(vertx, "testtenant");
+    insert(testContext, pg, "delete_test",  1);
```
julianladisch, line 440: This test reuses the name "delete_test". This makes this test fail if the test deleteByCQLNullHeaders runs before this test:
```
[ERROR] deleteByCQLOK  Time elapsed: 0.009 s  <<< ERROR!
io.vertx.pgclient.PgException: { "message": "duplicate key value violates unique constraint \"users_pkey\"", "severity": "ERROR", "code": "23505", "detail": "Key (id)=(4d318819-8675-07ab-d95b-e877c5789410) already exists.", "file": "nbtinsert.c", "line": "427", "routine": "_bt_check_unique", "schema": "testtenant_raml_module_builder", "table": "users", "constraint": "users_pkey" }
```

A different name should be used.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -402,6 +402,109 @@ public void deleteByIdResponseWithout204(TestContext testContext) {
...
+    String cql = "username=delete_test";
```
julianladisch, line 441: This doesn't test the key feature of this method: Deleting more than one record at a time.
Why don't you create "delete_a", "delete_b1", "delete_b2" and delete by CQL "username=delete_b*"?

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
adamdickmeiss: Deleting that name in use for NullHeaders

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
adamdickmeiss: yes

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
adamdickmeiss: Done, though I'm really not too worried .. PostgresClient.delete does the job.

adamdickmeiss:
https://github.com/folio-org/raml-module-builder/pull/829

PR #819, opened 2020-12-14
RMB-727: Revert requiring 400 status code for PgUtil.postSync

julianladisch:
Use 500 (not 400) as fall-back if 409 method is not available.

This avoids a breaking change. 500 method has been required before.


hjiebsco (approved):

hjiebsco (commented):
Also we can remove line 1045 now.

PR #831, opened 2020-12-22
Remove jdbc driver from runtime RMB-642

adamdickmeiss:
The driver is used for testing in cql2pgjson but that's not a problem.

These public methods were removed from PostgresClient: importFile and dropCreateDatabase

dropCreateDatabase is used by mod-vendors in a number of places, but that module is archived and deprecated.

importFile is also only used by deprecated-mod-acquisitions-postgres

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -3673,7 +3673,7 @@ public class PostgresClient {
   }
 
   /**
-   * Execute multiple SQL commands ine one transaction as superuser.
+   * Execute multiple SQL commands in a transaction as superuser.
    *<p>
    *   Currently this function always succeeds. Failure is indicated by
    *   the lines returned (empty on no errors).
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -3703,17 +3703,25 @@ public class PostgresClient {
                         log.info("Successfully executed {}", stmt);
                         return Future.succeededFuture();
                       }, res -> {
+                        log.error(res.getMessage(), res);
                         results.add(stmt);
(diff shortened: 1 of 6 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3702,155 +3672,64 @@ public void runSQLFile(String sqlFile, boolean stopOnError,
...
-   * \.
-   * 2. Copy from a data file packaged in the jar
-   * COPY config_data (jsonb) FROM 'data/locales.data' ENCODING 'UTF8';
-   * 3. Copy from a file on disk (absolute path)
-   * COPY config_data (jsonb) FROM 'C:\\Git\\configuration\\mod-configuration-server\\src\\main\\resources\\data\\locales.data' ENCODING 'UTF8';
-
-   * @param copyInStatement
-   * @param connection
-   * @throws Exception
+   * Execute multiple SQL commands ine one transaction as superuser.
```
hjiebsco, line 3676: Typo `ine`

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3702,155 +3672,64 @@ public void runSQLFile(String sqlFile, boolean stopOnError,
...
-            }
+    List<String> results = new LinkedList<>();
+
+    getInstance(vertx).getClient().getConnection()
+        .compose(conn -> conn.begin()
+            .compose(tx -> {
+              Future<Void> future = Future.succeededFuture();
+              for (int i = 0; i < sql.length; i++) {
+                String stmt = sql[i];
+                int fi = i;
```
hjiebsco, line 3698: not used variable

hjiebsco (approved):

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: yep

PR #832, opened 2020-12-23
Rmb 780 npe in http json response handler

adamdickmeiss:


adamdickmeiss:
Made another PR with correct base and further clean up here: https://github.com/folio-org/raml-module-builder/pull/833

PR #833, opened 2020-12-23
Fix NPE in HTTPJsonResponseHandle RMB-780

adamdickmeiss:
Close WebClient (previous HttpClient) whether failure or success (as in RMB 31).
Do not check for "Accept" header. It's useless, because there's no Accept header
in response headers!

Add more tests.


domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -64,21 +62,6 @@ public void handle(AsyncResult<HttpResponse<Buffer>> res) {
         log.error(e.getMessage(), e);
       }
     }
-      /* if(r.error != null && rollbackURL != null){
-
-      }*/
-
-    if (webClient != null) {
```
grabsefx, line 71: I don't see any place where `webclient` class field have been initialized or used. I beleive it could be removed from class definition as well

grabsefx (approved):

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
adamdickmeiss: It's set in HttpModuleClient2 .. when autoCloseConnections = true.

PR #834, opened 2021-01-05
Make postTenantSync public RMB-782

adamdickmeiss:


doc/upgrading.md
```diff
@@ -63,6 +67,8 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
        } ]
     } ]
 ```
+Tenant interface is supported by Okapi 4.5.0 and later. RMB 32, thus, will
```
hjiebsco, line 70: Maybe `Tenant interface 2.0` would be more clear?

hjiebsco (approved):

doc/upgrading.md
adamdickmeiss: Right. I'll change that. Thx

PR #841, opened 2021-01-14
Relax on Accept and Content-Type RMB-519

adamdickmeiss:
Accept default is `"*/*"`. Content-Type default is `"application/json"`.
The existing implementation hard-coded content-types for which
RMB would buffer the incoming request. This is now simplified and
no longer hard-coded (and default operation is to buffer the
incoming request). So this fixes the problem for RMB support all
content types for regular operations.

This PR also removes "final" pieces of multipart upload, which is considered broken since July 2018.


julianladisch (approved):
Looks good, thanks!

adamdickmeiss:
Thx Julian. I added one more commit, but merging now.

PR #842, opened 2021-01-15
Futurise PostgresClient.getConnection

julianladisch:


adamdickmeiss:
--- a/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
+++ b/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientIT.java
@@ -1928,7 +1928,7 @@ public class PostgresClientIT {
 
       @Override
       public Future<SqlConnection> getConnection() {
-        return null;
+        return Future.succeededFuture(pgConnection);
       }
 
       @Override
@@ -2080,7 +2080,7 @@ public class PostgresClientIT {
 
       @Override
       public Future<SqlConnection> getConnection() {
-        return null;
+        return Future.succeededFuture(pgConnection);
       }
 
       @Override
@@ -2208,7 +2208,7 @@ public class PostgresClientIT {
 
       @Override
       public Future<SqlConnection> getConnection() {
-        return null;
+        return Future.succeededFuture(pgConnection);
       }

adamdickmeiss (approved):

hjiebsco (approved):

PR #846, opened 2021-01-28
RMB-483 testContainers for testing with Postgres

adamdickmeiss:
This is an extension to RMB that allows Postgres testing with TestContainers -  https://issues.folio.org/browse/RMB-483.

The existing postgres embedder is still enabled by default if "host" config is absent or if embed_postgres=true on the command-line. This PR makes RMB create a new artifact postgres-testing that should be included for modules in scope test. A module enables container based testing with:
```java
  PostgresClient.setPostgresTester(new PostgresTesterContainer());
```

Next major RMB could go one step further and not the include old tester. This would be a trivial, but will require modules to include postgres-testing artifact and enable postgres containers with the above construct.

Have tested with branch https://github.com/folio-org/mod-inventory-storage/tree/use-test-containers-from-rmb-483


domain-models-runtime-it/src/test/java/org/folio/rest/ApiTestBase.java
```diff
@@ -67,13 +70,16 @@ static void beforeAll() {
         contentType(ContentType.JSON);
 
     // delete tenant (schema, tables, ...) if it exists from previous tests, ignore errors.
```
julianladisch, line 72:
```suggestion
    // delete tenant (schema, tables, ...) if it exists from previous tests.
```
to match the code that asserts status code 204.

adamdickmeiss: thx. Will update.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -252,8 +258,11 @@ public static void setIsEmbedded(boolean embed){
    *
    * <p>-1 is the default.
    *
-   * @param port  the port for embedded PostgreSQL, or -1 to not overwrite the port
+   * @param port  the port for embedded PostgreSQL, or -1 to not overwrite the port.
+   * @deprecated will be deprecated in a future release
```
julianladisch, line 262:
```suggestion
   * @deprecated will be removed in a future release, use {@link #setPostgresTester(PostgresTester)} and {@link org.folio.postgres.testing.PostgresTesterContainer} instead that picks a free port.
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -262,10 +271,13 @@ public static void setEmbeddedPort(int port){
    * @return the port number to use for embedded PostgreSQL, or -1 for not overwriting the
    *         port number of the configuration.
    * @see #setEmbeddedPort(int)
+   * @deprecated will be depracted in a future release (system assigns port on its own)
```
julianladisch, line 274:
```suggestion
   * @deprecated will be removed in a future release, use {@link #setPostgresTester(PostgresTester)} and {@link org.folio.postgres.testing.PostgresTesterContainer} instead that picks a free port.
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresTesterEmbedded.java
```diff
@@ -0,0 +1,66 @@
...
+import org.folio.rest.tools.utils.NetworkUtils;
+import org.folio.util.PostgresTester;
+import ru.yandex.qatools.embed.postgresql.EmbeddedPostgres;
+import ru.yandex.qatools.embed.postgresql.distribution.Version;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Arrays;
+
+public class PostgresTesterEmbedded implements PostgresTester {
```
julianladisch, line 14: This class should be deprecated. The deprecation note should link to the replacement (PostgresTesterContainer) and mention that this class will be removed in the next major RMB version.

julianladisch (approved):
Elegant solution, thanks!

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresTesterEmbedded.java
adamdickmeiss: Yes. Will fix that.

hjiebsco (approved):
Very nice! Wondering if `cql2pgjson` can be tweaked to use the new `PostgresTesterContainer` as well. Also, I assume removing `postgres_import_path` stuff has no impact to other modules.

adamdickmeiss:
> Very nice! Wondering if `cql2pgjson` can be tweaked to use the new `PostgresTesterContainer` as well. Also, I assume removing `postgres_import_path` stuff has no impact to other modules.

It could use it. But we wouldn't save much in code size (if any). I'll leave it as is.

Good spot on `postgres_import_path`. It's not documented and might not even work (there's no unit test). I can only see it mentioned in GitHub in deprecated-mod-acquisitions-postgres . So I think we can get by.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresTesterEmbedded.java
adamdickmeiss: Done now. I also made it package private.. I don't see why anybody would use it!!

adamdickmeiss: Done now. I also made it package private.. I don't see why anybody would use it!!

PR #847, opened 2021-01-31
RMB-786 change OL notice level

hjiebsco:
Change trigger notice level so https://github.com/folio-org/mod-inventory-storage/pull/570 can pass unit tests with embedded PostgreSQL. I was able to verify it locally. Let's try it on b32.0 branch first.


adamdickmeiss (approved):
Excellent. Is that just by coincidence ? Or have you read somewhere that it could matter?

hjiebsco:
> Excellent. Is that just by coincidence ? Or have you read somewhere that it could matter?

I tried and found that out. It works on my local machine. hopefully it works on Jenkins as well. We will see soon.

PR #853, opened 2021-02-07
GenerateRunner and ClientGenerator as a maven plugin RMB-785

adamdickmeiss:
Convert GenerateRunner and ClientGenerator into the maven plugin "domain-models-maven-plugin" that runs at compile time only and is not shipped with domain-models-runtime.


adamdickmeiss:
As far as I can tell this PR does not break anything.. I've tried with mod-copycat and that seems fine.

I think we should fix the code smells.

There are also definitions that are duplicated .. they shouldn't be..  For example `INTERFACE_PACKAGE` is now defined in no less than 3 places. There are others like this. I think they should be defined in RTFConsts, and that class should be moved to domain-models-api-aspects (common for both plugin and runtime).

julianladisch:
The code smells existed before but they are reported now because we move around the .java files.
The code smells should be fixed, but not in this pull request.

adamdickmeiss:
> The code smells existed before but they are reported now because we move around the .java files.
> The code smells should be fixed, but not in this pull request.

Yes.. But we added some redundant definitions that weren't there before. I'll push shortly.

adamdickmeiss:
mod-inventory-storage and mod-copycat passes tests with this PR - at this point (when they have been updated to use the plugin of course).

julianladisch (approved):
Looks good, thanks!

PR #856, opened 2021-02-15
RMB-799: Update Vert.x from 4.0.0 to 4.0.2

julianladisch:
Vertx 4.0.2 has many bug fixes, see release notes:

https://github.com/vert-x3/wiki/wiki/4.0.2-Release-Notes


adamdickmeiss (approved):
We also need the 4.0.2 versions of vertx-{pg,sql}-client.

julianladisch:
RMB doesn't build if the vertx-{pg,sql}-client of that version is not available:
https://github.com/folio-org/raml-module-builder/blob/master/pom.xml#L54-L68
Our forked vertx-{pg,sql}-clients have been available with version 4.0.2 since Sunday.

PR #857, opened 2021-02-15
GeneratedClient should not throw Generic Exception RMB-800

adamdickmeiss:
Json encode errors gives runtime exception IllegalArgumentExcption.
URLDecoder.encode uses StandardCharsets.UTF_8 thereby avoiding throwing
UnsupportedEncodingException.


julianladisch (approved):
Looks good, thanks!

PR #859, opened 2021-02-18
RMB-803: Fix URL encoding in BuildCQL preventing CQL injection

julianladisch:
BuildCQL.buildCQL() doesn't URL encode cqlStatementOperator
and operatorBetweenArgs. This may result in CQL injection.

Solution:

URL encode the complete CQL query.


adamdickmeiss (approved):
thx

PR #860, opened 2021-02-22
TenantInit async utility RMB-806

adamdickmeiss:
Add TenantInit.{purge,exec} to ease unit testing.


domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantInit.java
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Perform tenant purge, wait for result and delete job.
+   * @param client Tenant Client.
+   * @param waitMs number of milliseconds to wait for job completion (0=no wait).
+   * @return async result.
+   */
+  public static Future<Void> purge(TenantClient client, int waitMs) {
+    TenantAttributes at = new TenantAttributes().withPurge(true);
+    return exec(client, at, waitMs);
+  }
```
hjiebsco, line 27: Curious, is `module id` not needed for `purge`?

adamdickmeiss: Not really.

hjiebsco (approved):

PR #861, opened 2021-02-22
RestVerticle checks for implementation at startup RMB-805

adamdickmeiss:
Big refactor; all user requests handled in static class RestRouting.
Use Vert.x route for each of the routes offered, instead of
one big fat route function that does everything (slow and big).
This has been tested with mod-inventory-storage and all tests pass.
RMB returns 404 on path/method not found (actually this is thrown by Vert.x
itself now that only supported routes are registered).

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
@@ -129,7 +129,6 @@ public class RestVerticle extends AbstractVerticle {
           return Future.succeededFuture();
         })
         .onFailure(cause -> {
-          String reason = cause.getMessage();
           log.error(cause.getMessage(), cause);
           startPromise.fail(cause);
         })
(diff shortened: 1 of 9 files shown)


domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -155,338 +62,78 @@ public void start(Promise<Void> startPromise) throws Exception {
...
-            v -> withRequestId(rc, () ->
-              LogUtil.formatLogMessage(className, "start", " invoking " + method2Run))
-        );
-        endRequestWithError(rc, 400, true, "unable to upload file " + event.getMessage(), validRequest);
-      }
-    });
+          return Future.succeededFuture();
+        })
+        .onFailure(cause -> {
+          String reason = cause.getMessage();
```
julianladisch, line 132: Unused variable.

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -155,338 +62,78 @@ public void start(Promise<Void> startPromise) throws Exception {
...
+          log.error(cause.getMessage(), cause);
+          startPromise.fail(cause);
+        })
+        .onSuccess(x -> startPromise.complete());
```
julianladisch, line 135: Should be combined:
```
        .onFailure(cause -> log.error(cause.getMessage(), cause))
        .onComplete(startPromise);
```

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -155,338 +62,78 @@ public void start(Promise<Void> startPromise) throws Exception {
...
+      port = config().getInteger(HTTP_PORT_SETTING, 8081);
+    }
+    RestRouting.populateRoutes(router, packageOfImplementations)
+        .compose(x -> runHook())
+        .compose(x -> {
+          log.info("Listening port {}", port);
+          return server.requestHandler(router).listen(port);
+        })
+        .compose(ret -> {
+          try {
```
julianladisch, line 106: The 2 try-catch clauses within the .compose are not needed because .compose catches all exceptions.
This is possible if runPeriodicHook() and runPostDeployHook(Handler) no longer throw (checked) Exception - simply add
```
    } catch (IOException|IllegalArgumentException|ReflectiveOperationException e) {
      throw new RuntimeException(e);
     }
```
```
    } catch (IOException|ReflectiveOperationException e) {
      throw new RuntimeException(e);
    }
```
to them to throw unchecked exceptions only.

julianladisch (approved):
Looks good, thanks!

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
adamdickmeiss: done

PR #870, opened 2021-02-26
Make AnnotationGrabber.METHOD_URL a path (NOT a regex) RMB-808

adamdickmeiss:


domain-models-maven-plugin/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -334,13 +334,13 @@ public void generateMethodMeta(String methodName, JsonObject params, String url,
     /* Handle place holders in the URL
       * replace {varName} with "+varName+" so that it will be replaced
       * in the url at runtime with the correct values */
-    Matcher m = Pattern.compile("\\{.*?\\}").matcher(url);
-    while(m.find()){
+    Matcher m = Pattern.compile("\\{.*?\\}").matcher("^" + url);
```
julianladisch, line 337: "^" is not needed here.

adamdickmeiss: Good point!

domain-models-maven-plugin/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -334,13 +334,13 @@ public void generateMethodMeta(String methodName, JsonObject params, String url,
...
+    while (m.find()) {
       String varName = m.group().replace("{","").replace("}", "");
       url = url.replace("{"+varName+"}", "\"+"+varName+"+\"");
     }
 
-    url = "\""+url.substring(1)+"\"+queryParams.toString()";
+    url = "\"" + url + "\"+queryParams.toString()";
```
julianladisch, line 343:
```suggestion
    url = "\"" + 
        Pattern.compile("\\{([^{}/]+)\\}").matcher(url).replaceAll("\"+$1+\"") + 
        "\"+queryParams.toString()";
```

adamdickmeiss: This is supposed to replace the preceding for loop.. That results in failures.. We need a better test for clientGenerator itself --  the TestResource.java does not expose enough features!

julianladisch: This should replace all the other Matcher and url code above.

julianladisch (approved):

domain-models-maven-plugin/src/main/java/org/folio/rest/tools/ClientGenerator.java
adamdickmeiss: OK.. I'll try again.. I like the simplication.

adamdickmeiss: OK.. I'll try again.. I like the simplification.

adamdickmeiss: OK. Good now.

PR #872, opened 2021-03-01
Fix form parameters lost RMB-809

adamdickmeiss:
Make application/x-www-form-urlencoded work again.


domain-models-api-interfaces/pom.xml
```diff
@@ -8,7 +8,7 @@
   <artifactId>domain-models-api-interfaces</artifactId>
   <properties>
     <jaxrsVersion>2.0</jaxrsVersion>
-    <generate_routing_context>/rmbtests/test</generate_routing_context>
+    <generate_routing_context>/rmbtests/test,/rmbtests/testForm</generate_routing_context>
```
hjiebsco, line 11: Do we need this line? I removed this whole line and `mvn install` still works for me

domain-models-api-interfaces/ramls/sample.raml
```diff
@@ -160,3 +160,14 @@ traits:
...
+    post:
+      body:
+        application/x-www-form-urlencoded:
+      responses:
+        200:
+          body:
+            application/json:
+        500:
+          body:
+            text/plain:
```
hjiebsco, line 173: No `400`?

hjiebsco (commented):
Left two minor inline comments. Also, can we remove `rules` folder as part of this PR? Not related but `rules` has been an empty folder for a while.

domain-models-api-interfaces/pom.xml
adamdickmeiss: Good question!

```sh
cd domain-models-api-interfaces
mvn -B install
```

and you'll notice that Rmbtests.java do not have any routingContext calls! So this is needed or something
else that makes it take property from parent pom.

domain-models-api-interfaces/ramls/sample.raml
adamdickmeiss: OK, will add it.

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -345,6 +348,17 @@ public void invalidCqlClosesConnection(TestContext context) {
...
+  @Test
+  public void testForm(TestContext context) {
+    JsonPath jsonPath = given().spec(tenant)
+        .formParam("form1name", "form1value")
+        .formParam("form2name", "form2value")
+        .post("/rmbtests/testForm")
+        .then()
+        .statusCode(200).extract().body().jsonPath();
+    context.assertEquals("form1value", jsonPath.getString("[0].form1name"));
+    context.assertEquals("form2value", jsonPath.getString("[1].form2name"));
```
julianladisch, line 360:
```
        .then()
        .statusCode(200)
        .body("[0].form1name", is("form1value"),
              "[1].form2name", is("form2value"));
```
Rest-assured has built-in JSON path testing.

julianladisch (approved):
Looks good, thanks!

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
adamdickmeiss: Thx @julianladisch .. I new there was a better way!

adamdickmeiss:
Jenkins failed.. IT has happened a couple of times in the last week.. Sometimes test containers, just won't start:
```
16:29:17 [] [] [] [] INFO  PostgresClient       No DB configuration found, switching to embedded mode
16:33:40 [] [] [] [] ERROR PostgresClient       Container startup failed
org.testcontainers.containers.ContainerLaunchException: Container startup failed
        at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:330) ~[testcontainers-1.15.1.jar:?]
        at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:311) ~[testcontainers-1.15.1.jar:?]
        at org.folio.postgres.testing.PostgresTesterContainer.start(PostgresTesterContainer.java:44) ~[postgres-testing-33.0.0-SNAPSHOT.jar:?]
        at org.folio.rest.persist.PostgresClient.startEmbeddedPostgres(PostgresClient.java:3790) ~[domain-models-runtime-33.0.0-SNAPSHOT.jar:?]
        at org.folio.rest.persist.PostgresClient.init(PostgresClient.java:499) ~[domain-models-runtime-33.0.0-SNAPSHOT.jar:?]
        at org.folio.rest.persist.PostgresClient.<init>(PostgresClient.java:159) ~[domain-models-runtime-33.0.0-SNAPSHOT.jar:?]
```

domain-models-api-interfaces/pom.xml
hjiebsco: > Good question!
> 
> ```shell
> cd domain-models-api-interfaces
> mvn -B install
> ```
> 
> and you'll notice that Rmbtests.java do not have any routingContext calls! So this is needed or something
> else that makes it take property from parent pom.

Do you mean `cd domain-models-runtime`? The `Rmbtests.java` is in `domain-models-runtime` not in `domain-models-api-interfaces`. Anyway, without that line, I can still see below in the console.
```
12:22:31 [] [] [] [] INFO  RestRouting          invoking postRmbtestsTestForm
12:22:31 [] [] [] [] INFO  LogUtil              127.0.0.1:62209 POST /rmbtests/testForm null HTTP_1_1 200 70 11 tid=folio_shared OK
```

PR #873, opened 2021-03-01
RMB-770 Add PERSONAL_DATA_DISCLOSURE form

hjiebsco:
Please see https://issues.folio.org/browse/RMB-770 

Note: below would mean any RMB based modules will need to check both `userId` and `IP` in the form.

- RMB uses okapi-common logging to log `userId`
- `IP` is used in https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/LogUtil.java#L25 and https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/LogUtil.java#L35

### DIFF
--- a/PERSONAL_DATA_DISCLOSURE.md
+++ b/PERSONAL_DATA_DISCLOSURE.md
@@ -1,0 +1,58 @@
+
+
+## Overview
+The purpose of this form is to disclose the types of personal data stored by each module.  This information enables those hosting FOLIO to better manage and comply with various privacy laws and restrictions, e.g. GDPR.
+
+It's important to note that personal data is not limited to that which can be used to identify a person on it's own (e.g. Social security number), but also data used in conjunction with other data to identify a person (e.g. date of birth + city + gender).
+
+For the purposes of this form, "store" includes the following:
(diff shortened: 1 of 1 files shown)


adamdickmeiss (approved):

PERSONAL_DATA_DISCLOSURE.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] Passport number / National identification numbers
+- [ ] Driver’s license number
+- [ ] Social security number
+- [ ] Email address
+- [ ] Web cookies
+- [x] IP address
+- [ ] Geolocation data
+- [ ] Financial information
+- [ ] Logic or algorithms used to build a user/profile
+<!--- - [ ] Other personal data - Please list as needed -->
```
julianladisch, line 40: - [x] Date and time of the request

hjiebsco: > * [x]  Date and time of the request

Per core platform standup discussion, data and time are considered as personal data.

PR #876, opened 2021-03-05
db_connection LoadConfs code and documentation cleanup

julianladisch:
* Remove undocumented configuration loading from URL to close
  this security hole

* Expand db_connection documentation

* LoadConfs code cleanup and unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,12 +257,18 @@ or `org.folio.rest.RestVerticle` for a specific class.
 
 - `embed_postgres=true` Optional -- enforces starting an embedded postgreSQL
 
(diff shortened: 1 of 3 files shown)


adamdickmeiss (approved):
LGTM, although I'd rather see this feature be removed completely.

hjiebsco (approved):
Should those `info` log level be a bit higher, maybe `warning`?

julianladisch:
> Should those `info` log level be a bit higher, maybe `warning`?

No, because not specifying db_connection uses the default `/postgres-conf.json` and will emit these log messages when the file and the resource is not found.

PR #881, opened 2021-03-07
Mention pluginRepositories section

adamdickmeiss:
Rewrite introduction section so that it no longer mentions
exec-maven-plugin, but instead the domain-models-maven-plugin
and repositories. This a follow-up to RMB-810.


adamdickmeiss:
Making this work in another branch.

PR #882, opened 2021-03-08
Fix HttpClientFactory mock can never be unset RMB-811

adamdickmeiss:
Not per Verticle yet, but at least the property is honored.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/tools/client/HttpClientFactory.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/tools/client/HttpClientFactory.java
@@ -8,53 +8,47 @@ import org.folio.rest.tools.client.test.HttpClientMock2;
  *
  */
 public class HttpClientFactory {
-
-
-  private static boolean mock = false;
-
-  static {
-    if(System.getProperty(HttpClientMock2.MOCK_MODE) != null ){
-      mock = true;
-    }
+  private static boolean isMock() {
+    return System.getProperty(HttpClientMock2.MOCK_MODE) != null;
   }
 
   public static HttpClientInterface getHttpClient(String host, int port, String tenantId, boolean keepAlive, int connTO,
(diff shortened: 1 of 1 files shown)


hjiebsco (approved):

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HttpClientFactory.java
```diff
@@ -8,53 +8,47 @@
...
-
-
-  private static boolean mock = false;
-
-  static {
-    if(System.getProperty(HttpClientMock2.MOCK_MODE) != null ){
-      mock = true;
-    }
+  private static boolean isMock() {
+    return System.getProperty(HttpClientMock2.MOCK_MODE) != null;
```
julianladisch, line 12: Calling System.getProperty is slow. That's the reason why the `mock` variable has been introduced in the first place. Please keep the `mock` variable.
You may add a setMock(boolean) method.

julianladisch (changes requested):
Please keep the mock variable for performance reasons.

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HttpClientFactory.java
adamdickmeiss: Slow? I doubt you even notice the difference.. This is called when a client is constructed . A client  may be used for multiple HTTP requests.. So my immediate thinking is that we are talking 3 orders of magnatude.

Having said that, I should have marked this as draft. Using a property to signal "singleton" like things from one part of RMB to another is just plain bad design. Eg RestVerticle sets it and it's then picked up here.

Beyond this, this factory, is just plain broken.. One webclient per client. Much worse for performance.. If that is opened and closed for each request, that is much slower than keepint the connection alive (so the connect handshake is saved).

Anyway, I'll see if I can get this to work without that property at all.

adamdickmeiss:
New PR in #883

PR #884, opened 2021-03-09
RMB-812: Make PostgresClientStreamResult public

julianladisch:
PostgresClient.streamGet methods have this reply parameter:

Handler<AsyncResult<PostgresClientStreamResult<T>>> replyHandler

This can only been used from all packages if the visibility of PostgresClientStreamResult is changed from package-private to public.


adamdickmeiss (approved):
OK

PR #888, opened 2021-03-10
RMB-602: PostgresClient futurisation of saveBatch/upsertBatch

julianladisch:


domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
```diff
@@ -297,4 +306,133 @@ static String durationMsg(String descriptionKey, String sql, long startNanoTime)
...
+    return saveBatch(true, table, entities);
+  }
+
+  <T> Future<RowSet<Row>> saveBatch(boolean upsert, String table, List<T> entities) {
+
+    try {
+      List<Tuple> batch = new ArrayList<>();
+      if (entities == null || entities.isEmpty()) {
+        return Future.succeededFuture(emptyRowSetOfId());
+      }
```
hjiebsco, line 397: This does not need to be in the try block

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
```diff
@@ -297,4 +306,133 @@ static String durationMsg(String descriptionKey, String sql, long startNanoTime)
...
+      // We must use reflection, the POJOs don't have a interface/superclass in common.
+      Method getIdMethod = entities.get(0).getClass().getDeclaredMethod("getId");
+      for (Object entity : entities) {
+        Object obj = getIdMethod.invoke(entity);
+        UUID id = obj == null ? UUID.randomUUID() : UUID.fromString((String) obj);
+        batch.add(Tuple.of(id, PostgresClient.pojo2JsonObject(entity)));
+      }
+      return saveBatchInternal(upsert, table, batch);
```
hjiebsco, line 405: Would it work (and could be simpler) if you convert `List<T>` to `JsonArray` and call `saveBatch(boolean upsert, String table, JsonArray entities)` directly?

hjiebsco (approved):

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
julianladisch: entities.isEmpty() may throw an exception because it may be any custom List implementation. Putting it into the try-catch block ensures that the exception is wrapped into a failed Future.

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
julianladisch: `Conn#saveBatch(boolean upsert, String table, JsonArray entities)` calls
`Conn#saveBatchInternal(boolean upsert, String table, List<Tuple> batch)`.

`Conn#saveBatchInternal(boolean upsert, String table, List<Tuple> batch)` calls
`[PreparedQuery#executeBatch(List<Tuple> batch)](https://vertx.io/docs/apidocs/io/vertx/sqlclient/PreparedQuery.html#executeBatch-java.util.List-)`.

We should directly convert to `List<Tuple>`. I don't see an advantage in converting to `JsonArray` that requires another conversion to `List<Tuple>`.

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
hjiebsco: Fine. Can you switch the order of `batch` declaration and `entities` check? There is no reason to declare the `batch` if `entities` check could fail.

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
hjiebsco: My point is why do you need the reflection call? If you can assume there is a `getId` method you can also convert the object to json directly and assume there is an `id` field.

julianladisch: The code has been moved from https://github.com/folio-org/raml-module-builder/blob/v32.2.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L1181 to Conn.java.
PostgresClient.pojo2JsonObject uses Jackson ObjectMapper that probably also uses reflection.
The change you suggest requires careful investigation and is out of scope of RMB-602. Please create a new Jira for it.

PR #895, opened 2021-03-23
RMB-814: PostgresClient.selectStream futurisation

julianladisch:


adamdickmeiss (approved):

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
```diff
@@ -866,4 +867,21 @@ static String durationMsg(String descriptionKey, String sql, long startNanoTime)
...
     return streamGet(table, clazz, PostgresClient.DEFAULT_JSONB_FIELD_NAME, wrapper, false, null, null, streamHandler);
   }
+
+  /**
+   * Get a stream of the results of the {@code sql} query.
+   *
+   * @param params arguments for {@code $} placeholders in {@code sql}
+   * @param chunkSize cursor fetch size
+   */
+  public Future<RowStream<Row>> selectStream(String sql, Tuple params, int chunkSize) {
```
hjiebsco, line 877: Will there be a futurized version of `selectStream` in `PostgresClient`?

julianladisch:
No.
When selectStream returns a RowStream the connection is still open.
You always need a connection and should use Conn if you want to use futurised methods:

```
PostgresClient.withConn(conn -> {
  return conn.selectStream(...)
  .compose ...
});
```

hjiebsco:
> No.
> When selectStream returns a RowStream the connection is still open.
> You always need a connection and should use Conn if you want to use futurised methods:
> 
> ```
> PostgresClient.withConn(conn -> {
>   return conn.selectStream(...)
>   .compose ...
> });
> ```

So consumer has to use `PostgresClient` for some methods but `Conn` for others? Is it possible to simplify it by just exposing `PostgresClient` but not `Conn`?

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
julianladisch: I've added it.

julianladisch:
I've changed the implementation with better futurisation.
PostgresClient.selectStream should be used if RMB should create the connection and the transaction.
Conn.selectStream should be used if the query should run on an existing connection in an existing transaction.

hjiebsco (approved):

PR #896, opened 2021-03-25
RMB-814: PostgresClient.execute futurisation

julianladisch:


adamdickmeiss (approved):

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
```diff
@@ -866,4 +868,88 @@ static String durationMsg(String descriptionKey, String sql, long startNanoTime)
...
+   * @return the reply from the database
+   */
+  public Future<RowSet<Row>> execute(String sql, Tuple params) {
+    try {
+      long start = log.isDebugEnabled() ? System.nanoTime() : 0;
+      // more than optimization.. preparedQuery does not work for multiple SQL statements
+      if (params.size() == 0) {
+        return pgConnection.query(sql).execute()
+            .onComplete(x -> log.debug(() -> durationMsg("execute", sql, start)));
+      }
```
hjiebsco, line 886: This means every call of this method will be checked for empty `params`. The alternative would be not reuse this method for `execute(sql)`.

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
```diff
@@ -866,4 +868,88 @@ static String durationMsg(String descriptionKey, String sql, long startNanoTime)
...
+   */
+  public Future<List<RowSet<Row>>> execute(String sql, List<Tuple> params) {
+
+    try {
+      long start = log.isDebugEnabled() ? System.nanoTime() : 0;
+      Promise<List<RowSet<Row>>> promise = Promise.promise();
+      List<RowSet<Row>> results = new ArrayList<>(params.size());
+      Iterator<Tuple> iterator = params.iterator();
+      PreparedQuery<RowSet<Row>> preparedQuery = pgConnection.preparedQuery(sql);
+      Runnable task = new Runnable() {
```
hjiebsco, line 925: Wondering if `executeBatch` is not used here is because sequential execution is desired. If so, maybe the method name and doc can be more specific about that.

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
julianladisch: Calling `params.size()` is very fast: https://github.com/eclipse-vertx/vertx-sql-client/blob/4.0.3/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ArrayTuple.java#L71
Moving the code for empty params into `execute(String sql)` requires to duplicate the try-catch and the `long start` code.
This empty params check is taken from old code in PostgresClient:
* https://github.com/folio-org/raml-module-builder/blob/v32.2.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L3362
* https://github.com/folio-org/raml-module-builder/blob/v32.2.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java#L3388

The performance degregation of the if clause is unnoticeable and doesn't justifiy deduced source code readability.

domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
julianladisch: Good catch!
I've created https://issues.folio.org/browse/RMB-819 for this change.

PR #897, opened 2021-03-26
RMB-819: Atomic batch PostgresClient.execute with List<Tuple>

julianladisch:
Futurised methods return a Future<RowSet<Row>>.

A List<RowSet<Row>> is not needed because a
RowSet<Row> is an Iterable<Row>.

To avoid a breaking change existing methods with a
`Handler<AsyncResult<List<RowSet<Row>>>> replyHandler`
parameter are not changed.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
@@ -8,7 +8,6 @@ import io.vertx.core.Promise;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.pgclient.PgConnection;
-import io.vertx.sqlclient.PreparedQuery;
 import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowSet;
 import io.vertx.sqlclient.Tuple;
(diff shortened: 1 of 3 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/Conn.java
```diff
@@ -906,46 +904,25 @@ static String durationMsg(String descriptionKey, String sql, long startNanoTime)
...
+   *                    parameters for the {@code $} placeholders.
+   * @return the reply from the database, one RowSet per params Tuple. null if params.size() == 0.
    */
-  public Future<List<RowSet<Row>>> execute(String sql, List<Tuple> params) {
+  public Future<RowSet<Row>> execute(String sql, List<Tuple> params) {
 
     try {
+      if (params.size() == 0) {
+        return Future.succeededFuture();
+      }
```
hjiebsco, line 920: Do nothing? Or should it fall back to `execute(String sql)`?

julianladisch: "params - there is one list entry for each SQL invocation"

`sql` is run two times if params.size() == 2.
`sql` is run one time if params.size() == 1.
`sql` is run zero times if params.size() == 0.

I've added javadoc and unit tests for these cases.

hjiebsco: Thanks for updating the javadoc to make it explicit.

hjiebsco (approved):

PR #901, opened 2021-04-08
Fix ConnIT.getByIdForUpdate race condition

julianladisch:
In about 10% of all builds ConnIT.getByIdForUpdate fails because
of a race condition. The value is still b but should be c.

It happens on Jenkins and locally.

This fix avoids the race condition by polling until success.


adamdickmeiss:
Is this related to https://issues.folio.org/browse/MODINVSTOR-690 ?

adamdickmeiss (approved):
ok. Fix for test, not actual bug.

hjiebsco (approved):

PR #904, opened 2021-04-12
RMB-796: ClientGenerator: add new HttpClient constructor

julianladisch:


domain-models-maven-plugin/src/main/java/org/folio/rest/tools/ClientGenerator.java
```diff
@@ -291,11 +320,12 @@ public void generateClassMeta(String className) {
 
   public void generateCloseClient(){
     JMethod jmCreate = method(JMod.PUBLIC, void.class, "close");
-    jmCreate.javadoc().add("Close the client. Closing will close down any "
-        + "pooled connections. Clients should always be closed after use.");
+    jmCreate.javadoc().add("Close the client. Closing will close the WebClient and the HttpClient. "
+        + "Only close if WebClient and HttpClient will not be re-used.");
+    jmCreate.javadoc().addDeprecated().append("A Verticle should keep the WebClient/HttpClient and reuse it.");
+    jmCreate.annotate(Deprecated.class);
     JBlock body = jmCreate.body();
```
adamdickmeiss, line 327: Looks like this method is not even called.. Perhaps just remove it?

julianladisch: Good catch, removed.

adamdickmeiss (approved):

hjiebsco (approved):
LGTM

PR #906, opened 2021-04-16
RMB-586 Fix for foreign key sub-field query causes missing FROM

MikhailFokanov:
RMB needs to replace . by _ in the fieldName when generating SQL.


cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -212,4 +212,16 @@ public void searchItemByHoldingsPermanentLocationId() throws Exception {
         + "WHERE permanentLocationId='53cf956f-c1df-410b-8bea-27f712cca7c0')";
     assertEquals(expected, sql);
   }
+
+
+    @Test
+  public void searchItemByHoldingsPermanentLocationName() throws Exception {
+      CQL2PgJSON cql2PgJSON = cql2pgJson("coursereserves_reserves", "courseReservesSchema.json");
+      String sql = cql2PgJSON
+        .toSql("(courseListingId = 1bd04c24-8f8c-4e33-b69c-359146e43808) AND copyrightStatus.name==cc").toString();
```
julianladisch, line 221: The test should be as concise as possible, `courseListingId = 1bd04c24-8f8c-4e33-b69c-359146e43808` is not needed in the CQL query.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -212,4 +212,16 @@ public void searchItemByHoldingsPermanentLocationId() throws Exception {
         + "WHERE permanentLocationId='53cf956f-c1df-410b-8bea-27f712cca7c0')";
     assertEquals(expected, sql);
   }
+
+
+    @Test
+  public void searchItemByHoldingsPermanentLocationName() throws Exception {
```
julianladisch, line 218: The method name is misleading. What about `fieldNameContainsDot`?

cql2pgjson/src/test/resources/templates/db_scripts/courseReservesSchema.json
```diff
@@ -0,0 +1,187 @@
+{
```
julianladisch, line 1: The test should be as concise as possible. Adding the complete courseReservesSchema.json is not needed and makes it difficult to understand the key point of the test.

For foreign key tests we already have https://github.com/folio-org/raml-module-builder/blob/master/cql2pgjson/src/test/resources/templates/db_scripts/foreignKey.json that can be extended by an entry in a"foreignKeys" list where "fieldName" contains a dot because it is a sub-field.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
julianladisch, line 218: This is a unit test only.
Please also add an integration test to https://github.com/folio-org/raml-module-builder/blob/master/cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationIT.java that ensures that PostgreSQL can correctly process the generated SQL. Simply extend https://github.com/folio-org/raml-module-builder/blob/master/cql2pgjson/src/test/resources/foreignKey.sql as needed.

julianladisch (changes requested):

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
MikhailFokanov: fixed

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
MikhailFokanov: done

cql2pgjson/src/test/resources/templates/db_scripts/courseReservesSchema.json
MikhailFokanov: done

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
MikhailFokanov: done

cql2pgjson/src/test/resources/templates/db_scripts/courseReservesSchema.json
julianladisch: courseReservesSchema.json should be removed if it is no longer in use.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -154,4 +154,8 @@ public void sqlInjection2() {
     //if the test passes it means it was not successful
     assertThat(cqla("tableb.prefix == \"x0')  or 1=1)--\""), is(empty()) );
   }
+
+   public void fieldNameContainsDot() throws Exception {
+     assertThat(cqlb("copyrightStatus.name==cc"), is(empty()) );
```
julianladisch, line 159: The test data should be adjusted so that the condition `copyrightStatus.name==cc` selects one out of (at least) two records.
Testing the returned record using `containsInAnyOrder(...)` proves that the code actually works.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationIT.java
```diff
@@ -154,4 +154,8 @@ public void sqlInjection2() {
     //if the test passes it means it was not successful
     assertThat(cqla("tableb.prefix == \"x0')  or 1=1)--\""), is(empty()) );
   }
+
+   public void fieldNameContainsDot() throws Exception {
```
julianladisch, line 158: `@Test` is missing.

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationTest.java
```diff
@@ -212,4 +212,16 @@ public void searchItemByHoldingsPermanentLocationId() throws Exception {
         + "WHERE permanentLocationId='53cf956f-c1df-410b-8bea-27f712cca7c0')";
     assertEquals(expected, sql);
   }
+
+
+    @Test
```
julianladisch, line 217:
```suggestion
  @Test
```

cql2pgjson/src/test/java/org/folio/cql2pgjson/ForeignKeyGenerationIT.java
julianladisch: Simply extend https://github.com/folio-org/raml-module-builder/blob/master/cql2pgjson/src/test/resources/foreignKey.sql as needed.

julianladisch (approved):
Looks good, thanks!

MikhailFokanov:
Thanks, @julianladisch

PR #907, opened 2021-04-16
Only use postgresTester on empty db configuration RMB-826

adamdickmeiss:


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -414,7 +414,7 @@ private void init() throws Exception {
 
     postgreSQLClientConfig = getPostgreSQLClientConfig(tenantId, schemaName, Envs.allDBConfs());
 
-    if (isEmbedded()) {
+    if (isEmbedded() && postgreSQLClientConfig.containsKey("testing")) {
```
hjiebsco, line 417: Technically it should check for `true` and there is no unit test to cover the new condition.

hjiebsco (approved):

julianladisch (changes requested):
I cannot review this because there is no use case on RMB-826 that shows why this is needed.
This looks like a breaking change, but documentation in README is missing.

adamdickmeiss:
> I cannot review this because there is no use case on RMB-826 that shows why this is needed.
> This looks like a breaking change, but documentation in README is missing.

I have added some material there now.. There's not a good place to put it in current manual.. But I agree that *something* somewhere should be added. .. TestContainers only mentioned once (in a sentence that really doesn't mean much to me)

doc/upgrading.md
```diff
@@ -44,6 +44,10 @@ Enable testing with postgres by calling
 before any calls to PostgresClient or the Verticle. It is also
 possible to provide your own by implemeting the `PostgresTester` interface.
 
+If database configuration is provided, the PostgresTester instance is
+*not used for testing. This allows testing to be performed on local
```
julianladisch, line 48:
```suggestion
*not* used for testing. This allows testing to be performed on local
```

julianladisch (approved):
Looks good, thanks!

doc/upgrading.md
adamdickmeiss: right

PR #908, opened 2021-04-19
RMB-829: Suppress "Table audit_... NOT FOUND" in loadDbSchema()

julianladisch:
Also search a table name in the auditingTableName field
and suppress a "Table ... NOT FOUND" warning if found to
be an auditing table.

### DIFF
--- a/cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
+++ b/cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
@@ -183,10 +183,15 @@ public class CQL2PgJSON {
   enum InitDbTableResult {
     TABLE_FOUND,
     AUDIT_TABLE_FOUND,
+    NO_PRIMARY_TABLE_NAME,
     NOT_FOUND,
   };
 
   InitDbTableResult initDbTable() {
+    if (jsonField == null) {
+      logger.log(Level.SEVERE, "loadDbSchema(): No primary table name, can not load");
+      return InitDbTableResult.NO_PRIMARY_TABLE_NAME;
+    }
     // Remove the json blob field name, usually ".jsonb", but in tests also
     // ".user_data" etc.
     String tname = this.jsonField.replaceAll("\\.[^.]+$", "");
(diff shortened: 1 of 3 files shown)


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -180,27 +180,41 @@ private void loadDbSchema(String schemaPath) {
...
+    boolean isAuditTable = false;
+    for (Table table : dbSchema.getTables()) {
+      if (tname.equalsIgnoreCase(table.getTableName())) {
+        dbTable = table;
+        return InitDbTableResult.TABLE_FOUND;
       }
-      if (dbTable == null) {
-        logger.log(Level.SEVERE, "loadDbSchema loadDbSchema(): Table {0} NOT FOUND", tname);
+      if (tname.equalsIgnoreCase(table.getAuditingTableName())) {
+        isAuditTable = true;
```
hjiebsco, line 210: Would it be simpler to use `return InitDbTableResult.AUDIT_TABLE_FOUND;` here?

hjiebsco (approved):

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
julianladisch: Yes, changed.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -180,27 +180,36 @@ private void loadDbSchema(String schemaPath) {
...
+  InitDbTableResult initDbTable() {
+    if (jsonField == null) {
+      logger.log(Level.SEVERE, "loadDbSchema(): No primary table name, can not load");
+      return InitDbTableResult.NO_PRIMARY_TABLE_NAME;
+    }
+    // Remove the json blob field name, usually ".jsonb", but in tests also
+    // ".user_data" etc.
+    String tname = this.jsonField.replaceAll("\\.[^.]+$", "");
+    for (Table table : dbSchema.getTables()) {
+      System.out.println(table.getTableName() + " " + table.getMode());
```
hjiebsco, line 199: Is this intended to keep?

hjiebsco (approved):

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
julianladisch: No, good catch!
Thanks!

PR #916, opened 2021-04-27
raml plugin no generated clients RMB-831

adamdickmeiss:
This is a clarification for client generated code, not a fix.

For earlier versions client code also required depending the artifact with generated classes.. Since the client generator is using annotations as "input" (not RAML).

 But also a bug fix for finding base path of RAML source (seen with mod-configuration).

### DIFF
--- a/domain-models-maven-plugin/src/main/java/org/folio/rest/tools/DomainModelsMojo.java
+++ b/domain-models-maven-plugin/src/main/java/org/folio/rest/tools/DomainModelsMojo.java
@@ -100,14 +100,13 @@ public class DomainModelsMojo extends AbstractMojo {
     setRootLogLevel(Level.WARNING);
     try {
       if (generateInterfaces) {
-        System.err.println("generate interfaces");
         generateInterfaces();
         ModuleNameWriter.writeModuleNameClass(project);
       }
-      System.err.println("generate clients " + generateClients);
-      ClientGenerator.generate(project.getBasedir().getAbsolutePath(), generateClients);
-      System.err.println("generate clients done");
-
+      if (generateClients) {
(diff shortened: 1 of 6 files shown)


domain-models-maven-plugin/src/main/java/org/folio/rest/tools/DomainModelsMojo.java
```diff
@@ -103,7 +103,10 @@ public void execute() {
         generateInterfaces();
         ModuleNameWriter.writeModuleNameClass(project);
       }
-      ClientGenerator.generate(project.getBasedir().getAbsolutePath(), generateClients);
+      if (generateClients) {
+        int sz = ClientGenerator.generate(project.getBasedir().getAbsolutePath());
+        System.err.println("Client classes generated: " + sz);
```
hjiebsco, line 108: Why `System.err` not `System.out`?

hjiebsco (approved):
LGTM

domain-models-maven-plugin/src/main/java/org/folio/rest/tools/DomainModelsMojo.java
adamdickmeiss: Other print statements in the plugin was err. Can you comment on that @julianladisch ? Is print statements the best to report info for a plugin?

julianladisch: `System.err` should be used for error logging, `System.out` should be used for non-error logging.
This might be used for different colors when displaying the output of the build process.
The existing statements (10 x System.err, 3 x System.out) are inconsistent.

julianladisch (approved):
Looks good, thanks!

PR #917, opened 2021-04-29
RMB-832: Implement upsert, updateBatch for optimistic locking

julianladisch:
Upsert using INSERT ... ON CONFLICT (id) DO UPDATE fails because the INSERT
trigger overwrites the _version property that the UPDATE trigger uses to
detect optimistic locking conflicts.

Solution:

Implement upsert using a transaction and executing two SQL commands:

UPDATE ...;
IF NOT FOUND THEN INSERT ...;

Adding updateBatch allows to replace upsertBatch depending on the use case:
insertBatch or updateBatch. When optimistic locking is enabled the client
must know in advance whether it is insert or update.


adamdickmeiss (approved):
LGTM, but I am no SQL expert.

PR #919, opened 2021-05-04
Fix form key is lost for large value RMB-834

adamdickmeiss:
Requires Vert.x 4.0.3. so RMB using that now, but not vertx-sql-client and
vertx-pg-client as they aren't published yet. Update Netty to 4.1.60.Final
as that's what Vert.x 4.0.3 uses anyway. Update to Okapi 4.7.2.

### DIFF
--- a/domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
+++ b/domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
@@ -341,20 +341,17 @@ public class DemoRamlRestTest {
 
   @Test
   public void testForm(TestContext context) {
-    StringBuilder v = new StringBuilder();
-    for (int i = 0; i < 1025; i++) { // 1024 is max size for Vert.x 4.0.2
-      v.append("x");
-    }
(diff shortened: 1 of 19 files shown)


hjiebsco (approved):
Thanks for the quick fix.

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -341,14 +341,20 @@ public void invalidCqlClosesConnection(TestContext context) {
 
   @Test
   public void testForm(TestContext context) {
+    StringBuilder v = new StringBuilder();
+    for (int i = 0; i < 1025; i++) { // 1024 is max size for Vert.x 4.0.2
+      v.append("x");
+    }
```
julianladisch, line 347: `String v = "x".repeat(1025);`

julianladisch (approved):
Looks good, thanks!

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
adamdickmeiss: Thx

PR #920, opened 2021-05-07
Explain DB_ env variables

julianladisch:


doc/upgrading.md
```diff
@@ -44,7 +44,8 @@ Enable testing with postgres by calling
 before any calls to PostgresClient or the Verticle. It is also
 possible to provide your own by implemeting the `PostgresTester` interface.
 
-If database configuration is provided, the PostgresTester instance is
+If [DB\_\* environment variable database configuration](../README.md#environment-variables)
+is provided, the PostgresTester instance is
 *not* used for testing. This allows testing to be performed on local
 Postgres instance. See [RMB-826](https://issues.folio.org/browse/RMB-826).
```
adamdickmeiss, line 51: OK.. The original wording was used because you can use either DB-environment variables or PostgreSQL JSON configuration (db_connection on command line, or API for that).

julianladisch: `DB_*` works out of the box, the JSON config requires to call the test in a different way.
Therefore `DB_*` is the preferred way. Is anyone using JSON config for testing?
This is the upgrade guide, the full details are in README.md.

adamdickmeiss: Yes. No problem.

adamdickmeiss (approved):

PR #923, opened 2021-05-07
RMB-839: Upgrade to Vert.x 4.1.0.Beta1 

julianladisch:


adamdickmeiss (approved):

adamdickmeiss (commented):
We should not merge this until mod-permissions problem is fully understood. See https://github.com/folio-org/mod-permissions/tree/testing-with-vertx-4-1-0-beta1
and https://issues.folio.org/browse/MODPERMS-147

PR #930, opened 2021-05-18
RMB-843: Use percent encoding, not url encoding, for HTTP requests

julianladisch:
RFC 3986 mandates that only percent encoding is used for
HTTP requests. Using a plus to encode a space is wrong:
https://datatracker.ietf.org/doc/html/rfc3986

Wrong encoding of status == open:
`?query=status+%3D%3D+open`

Correct encoding of status == open:
`?query=status%20%3D%3D%20open`

Approach:

Implement PercentCodec.

Put a warning into the StringUtil.urlEncode javadoc.

Replace StringUtil.urlEncode calls by PercentCodec calls.

### DIFF
--- a/util/src/main/java/org/folio/util/PercentCodec.java
+++ b/util/src/main/java/org/folio/util/PercentCodec.java
@@ -1,0 +1,134 @@
+package org.folio.util;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Percent encoding of UTF-8 strings, conforming to RFC 3986.
+ *
+ * <p>These are the characters that are never encoded: The letters a-z and A-Z and the digits 0-9, and the
+ * {@code -._~} characters. They are the
+ * <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-2.3">Unreserved Characters (RFC 3986 Section 2.3)</a>.
+ *
+ * <p>All other characters are always percent encoded.
+ *
+ * <p>Space is always encoded as {@code %20} to conform with RFC 3986.
+ * <a href="https://datatracker.ietf.org/doc/html/rfc1630#page-7">RFC 1630</a> allows to encode space as {@code +}
+ * but RFC 1630's category is "Informational" only, whereas RFC 3986's category is "Standards Track".
+ *
+ * <p>The only encoding difference between {@link java.net.URLEncoder} and this class is the space character:
+ * URLEncoder encodes it as {@code +} and this class encodes it as {@code %20}.
+ */
+public final class PercentCodec {
+  private static final char [] UNRESERVED = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~".toCharArray();
+  private static final char [] HEX = "0123456789ABCDEF".toCharArray();
+  private static final char [][] ENCODE = getEncodeArray();
+
+  private PercentCodec() {
+    throw new UnsupportedOperationException();
+  }
+
+  private static char [][] getEncodeArray() {
+    char [][] encode = new char [256][];
+    for (char c : UNRESERVED) {
+      char [] cc = { c };
+      encode[c] = cc;
+    }
+    for (int i=0; i<256; i++) {
+      if (encode[i] != null) {
+        continue;
+      }
+      char [] cc = { '%', HEX[i >> 4], HEX[i & 0xf] };
+      encode[i] = cc;
+    }
+    return encode;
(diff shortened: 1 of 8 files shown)


hjiebsco (commented):
Is there any existing url encode library we can use? If not, will it work if we just replace `+` with `%20` after the call to `URLEncode`? BTW, have we seen issues due to `+` in practice?

adamdickmeiss (commented):
A lot of work no existing utility? But hopefully it's been fun. ClientGenerator.java is also broken then.. As it generates clients using URLEncoder.encode for query parameter values. Should use the new utility too.

util/src/main/java/org/folio/util/PercentCodec.java
```diff
@@ -0,0 +1,134 @@
...
+    throw new UnsupportedOperationException();
+  }
+
+  private static char [][] getEncodeArray() {
+    char [][] encode = new char [256][];
+    for (char c : UNRESERVED) {
+      char [] cc = { c };
+      encode[c] = cc;
+    }
+    for (int i=0; i<256; i++) {
```
adamdickmeiss, line 40: This wouldn't be accepted Google Code Style. (blanks around binary ops)

util/src/main/java/org/folio/util/PercentCodec.java
```diff
@@ -101,7 +103,7 @@ public static Appendable encode(Appendable appendable, CharSequence charSequence
    * // yields "https://example.com/users?query=status%20%3D%3D%20open"
    * </pre>
    *
-   * @param charSequence the input in UTF8 encoding; if null or empty then "" is returned
+   * @param charSequence in UTF8 encoding; if null or empty then "" is returned
```
adamdickmeiss, line 106: The parameter is called `in` but you are referring to it as `charSequence`.

util/src/main/java/org/folio/util/PercentCodec.java
```diff
@@ -126,7 +128,7 @@ public static CharSequence encode(CharSequence in) {
    * <p>All characters except these get percent encoded: The letters a-z and A-Z and the digits 0-9, and the
    * {@code -._~} characters.
    *
-   * @param charSequence the input in UTF8 encoding; if null or empty then "" is returned
+   * @param charSequence in UTF8 encoding; if null or empty then "" is returned
```
adamdickmeiss, line 131: Same here. javadoc and parameter inconsistent.

adamdickmeiss (approved):

adamdickmeiss (approved):
Thx

julianladisch:
https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/net/PercentCodec.html
is a library that provides PercentCodec.
But the method signature `byte[] encode(byte[] bytes)` is not really useful.
The library comes with lots of other classes that we don't need, total size 345 KB.
The library is not in common use, only two FOLIO modules use it:
* mod-users-bl uses it for urlencode only, this can be replaced by RMB PercentCodec
* data-import-processing-core uses it for Base64, this should be replaced by java.util.Base64 that Java ships with since version 1.8.

util/src/main/java/org/folio/util/PercentCodec.java
julianladisch: Thanks, fixed.

util/src/main/java/org/folio/util/PercentCodec.java
julianladisch: Thanks, fixed.

util/src/main/java/org/folio/util/PercentCodec.java
```diff
@@ -0,0 +1,134 @@
...
+    return out;
+  }
+
+  /**
+   * Return charSequence with percent encoding (RFC 3986) applied.
+   *
+   * <p>All characters except these get percent encoded: The letters a-z and A-Z and the digits 0-9, and the
+   * {@code -._~} characters.
+   *
+   * @param charSequence the input in UTF8 encoding; if null or empty then "" is returned
```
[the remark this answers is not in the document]
julianladisch: Thanks, fixed.

util/src/main/java/org/folio/util/PercentCodec.java
```diff
@@ -0,0 +1,134 @@
...
+   * <p>Example usage:
+   *
+   * <pre>
+   * String baseUrl = "https://example.com/users";
+   * String query =  "status == open";
+   * String url = baseUrl + "?query=" + PercentCodec.encode(query);
+   * // yields "https://example.com/users?query=status%20%3D%3D%20open"
+   * </pre>
+   *
+   * @param charSequence the input in UTF8 encoding; if null or empty then "" is returned
```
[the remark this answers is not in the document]
julianladisch: Thanks, fixed.

PR #932, opened 2021-05-18
RMB-844: Properly report POST tenant with null exception message

julianladisch:
If a POST tenant fails TenantAPI should report the failure by setting the job error.

However, if the failure message is null the job error is set to null, this indicates no error.

Approach:

If failure message is null report the exception class name.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
@@ -283,10 +283,12 @@ public class TenantAPI implements Tenant {
           }
           return loadData(tenantAttributes, job.getTenant(), headers, context);
         })
-        .onComplete(res -> {
-          if (res.failed()) {
-            job.setError(res.cause().getMessage());
+        .onFailure(cause -> {
+          String message = cause.getMessage();
+          if (message == null) {
+            message = cause.getClass().getName();
           }
+          job.setError(message);
         }).mapEmpty();
   }
 


adamdickmeiss (approved):

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -283,10 +283,12 @@ private void completeJob(TenantJob job, Context context) {
...
         })
-        .onComplete(res -> {
-          if (res.failed()) {
-            job.setError(res.cause().getMessage());
+        .onFailure(cause -> {
+          String message = cause.getMessage();
+          if (message == null) {
+            message = cause.getClass().getName();
           }
+          job.setError(message);
```
hjiebsco, line 291:
```suggestion
          job.setError(cause.getClass().getName() + " : " + cause.getMessage());
```

hjiebsco (approved):

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
adamdickmeiss: The class name is not very useful in most cases, so I prefer Julian's approach (not to mention it preserves behavior to what it it is now -- some unit tests could break of this).

julianladisch: This fixes the problem, but always adds unnecessary information: If getMessage() is nul it adds "null", otherwise the message gives good information and the exception class name doesn't add any relevant information.

PR #934, opened 2021-05-20
Update to Vert.x 4.1.0.CR1 (Netty 4.1.65)

adamdickmeiss:


hjiebsco (approved):

adamdickmeiss:
We need the FOLIO vertx-{sql,pg}-client project updated too :-(

pom.xml
```diff
@@ -61,7 +61,7 @@
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-bom</artifactId>
-        <version>4.1.60.Final</version>
+        <version>4.1.64.Final</version>
```
julianladisch, line 64: Vert.x ships with Netty 4.1.65.Final.
Is there a reason to downgrade?
I would rather expect that the lines
```
      <!-- https://nvd.nist.gov/vuln/detail/CVE-2021-21290
           https://github.com/netty/netty/security/advisories/GHSA-5mcr-gq6c-3hq2
      -->
      <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-bom</artifactId>
        <version>4.1.64.Final</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
```
get removed from pom.xml because all vulnerabilities are fixed since 4.1.61.Final: https://github.com/netty/netty/security/advisories

julianladisch (approved):

pom.xml
adamdickmeiss: Thx .. Will remove. Missed the 65.

PR #936, opened 2021-05-25
RMB-848: Always close PostgresTester

julianladisch:
Add javadoc and a unit test to ensure that PostgresTester is always closed, even when the same PostgresTester instance is used twice in sequence.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -199,22 +199,14 @@ public class PostgresClient {
    * <p>If database configuration is already provided (DB_* env variables or JSON config file),
    * this call is ignored and testing is performed against the database instance given by configuration.
    *
-   * <p>Setting the same PostgresTester instance that is already set does nothing. This allows to reuse
-   * a database that some other test has already created, for example to copy a database with sample
-   * data using the <a href="https://stackoverflow.com/a/876565">TEMPLATE method</a>. See also
-   * <a href="https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers">
-   * singleton testcontainers</a>.
-   *
-   * <p>Setting a different PostgresTester instance closes the old PostgresTester instance.
+   * <p>Setting the same or a different PostgresTester instance invokes the close method of the
+   * old PostgresTester instance.
    *
    * <p>See {@link org.folio.postgres.testing.PostgresTesterContainer#PostgresTesterContainer()}
    *
    * @param tester instance to use for testing.
    */
   public static void setPostgresTester(PostgresTester tester) {
-    if (tester == postgresTester) {
-      return;
-    }
     stopPostgresTester();
     postgresTester = tester;
   }
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -55,7 +55,6 @@ import org.folio.rest.persist.interfaces.Results;
 import org.folio.rest.security.AES;
 import org.folio.rest.tools.utils.Envs;
 import org.folio.rest.tools.utils.MetadataUtil;
-import org.folio.rest.tools.utils.ModuleName;
 import org.folio.dbschema.ObjectMapperTool;
 import org.folio.util.PostgresTester;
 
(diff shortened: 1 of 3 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -194,13 +193,27 @@ private void statsTracker(String descriptionKey, String sql, long startNanoTime)
...
+   *
+   * <p>See {@link org.folio.postgres.testing.PostgresTesterContainer#PostgresTesterContainer()}
+   *
    * @param tester instance to use for testing.
    */
   public static void setPostgresTester(PostgresTester tester) {
+    if (tester == postgresTester) {
+      return;
+    }
     stopPostgresTester();
```
adamdickmeiss, line 217: I don't like this. starting the tester guarantees a new tester-- with empty data. Whether an old one is closed is irrelevant

We are merely saying that closing the old is not necessary as the system does it.

marcjohnson-kint:
> There a valid use cases where different test files use the same PostgresTester instance, for example to reuse a database with sample data.

What use cases are folks considering?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
marcjohnson-kint: This seems rather unlikely to happen in practice. If each test class is going to invoke this method (in order to allow for independent running) then to utilise this would require a shared instance of `PostgresTester` across all test classes. Have I understood that correctly?

marcjohnson-kint:
> Guys.. Let's make it possible to control this and leave as much hanging in the air for random behavior. 

What do you mean by `leave as much hanging in the air`?

> So you might have a static startup and start containers once in your test or you may do it for each test.. Nothing should be reused between test classes.. 

I agree that things shouldn't be reused across test classes where possible. (In effect, the PostgresTester is shared because it is static state of PostgresClient).

If you are advocating for an explicit API, then I agree. I find the current API confusing because it allows a mixture of shared and non-shared state and implicit and explicit control over the state of the tester (e.g. implicit start with explicit stop). I would prefer an API that either managed this entirely transparently for the client or required these things to be explicit.

Where does this leave the advice for folks upgrading. Should they have explicit `setPostgresTester`, `startPostgresTester` and `stopPostgresTester`?

adamdickmeiss:
> > Guys.. Let's make it possible to control this and leave as much hanging in the air for random behavior.
> 
> What do you mean by `leave as much hanging in the air`?

Tenant init half complete or complete or not at all.. So your further processing could be in trouble if you happen to use the same tenant name.. reference data loaded (or not). All things that persist.

> 
> > So you might have a static startup and start containers once in your test or you may do it for each test.. Nothing should be reused between test classes..
> 
> I agree that things shouldn't be reused across test classes where possible. (In effect, the PostgresTester is shared because it is static state of PostgresClient).
> 
> If you are advocating for an explicit API, then I agree. I find the current API confusing because it allows a mixture of shared and non-shared state and implicit and explicit control over the state of the tester (e.g. implicit start with explicit stop). I would prefer an API that either managed this entirely transparently for the client or required these things to be explicit.
> 
> Where does this leave the advice for folks upgrading. Should they have explicit `setPostgresTester`, `startPostgresTester` and `stopPostgresTester`?

They can - in 99% of the cases just get by with setPostgresTester. (It will restart or start). I don't think we should change behavior now. It seems to work fine. I'm all for improving documentation.

marcjohnson-kint:
> They can - in 99% of the cases just get by with setPostgresTester. (It will restart or start). I don't think we should change behavior now. It seems to work fine. I'm all for improving documentation.

I wasn't proposing changing the API at this point. I agree that it's too late in the cycle for that.

I think improving the documentation is a good idea. Making it explicit that most of the time no explicit control is needed, only the `setPostgresTester` in each test class, would be good.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: ok. I see that it is unlike to "cross" boundaries.. But what's the use case, then??

julianladisch: The if-clause has been removed.

adamdickmeiss (approved):

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -4101,8 +4107,16 @@ public void startPostgresTester() {
   }
 
   /**
-   * Stop Postgres Tester.
-   * Does nothing, if postgresTester is not enabled.
+   * Stop the PostgresTester instance.
+   *
+   * <p>Does nothing, if postgresTester is not enabled or stopPostgresTester() has already been called.
```
hjiebsco, line 4112: This line does not seem to match code 100% accurately. Do we even need this line?

julianladisch: postgresTester == null if postgresTester is not enabled or stopPostgresTester() has already been called.

hjiebsco: > postgresTester == null if postgresTester is not enabled or stopPostgresTester() has already been called.

Technically, `stopPostgresTester()` does not make `postgresTester == null` immediately.

hjiebsco (approved):

PR #944, opened 2021-06-14
RMB-849: Remove ModuleName class from domain-models-api-interfaces

julianladisch:
domain-models-api-interfaces (and domain-models-runtime
that depends on it) ships with a ModuleName class with
getModuleName() returning "raml_module_builder".

An RMB based module creates a correct ModuleName class
when running domain-models-maven-plugin.

If we are lucky the latter is included in the final jar.
For mod-feesfines I sometimes have integration tests failure
because of wrong order resulting in the
wrong "raml_module_builder" where it should be "mod_feesfines".

Removing ModuleName class from domain-models-api-interfaces jar
resolves this issue; it requires reflection to load
ModuleName class, though.

### DIFF
--- a/domain-models-runtime/src/test/java/org/folio/rest/tools/utils/ModuleName.java
+++ b/domain-models-runtime/src/test/java/org/folio/rest/tools/utils/ModuleName.java
@@ -1,0 +1,10 @@
+package org.folio.rest.tools.utils;
+
+public class ModuleName {
+  /**
+   * The module name with minus replaced by underscore, for example {@code mod_foo_bar}.
+   */
+  public static String getModuleName() {
+    return "raml_module_builder";
+  }
+}
--- a/domain-models-api-interfaces/pom.xml
+++ b/domain-models-api-interfaces/pom.xml
@@ -129,6 +129,25 @@
         </executions>
       </plugin>
 
+      <plugin>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>3.0.0</version>
+        <executions>
+          <execution>
+            <id>delete-modulename-class</id>
+            <phase>process-sources</phase>
+            <goals>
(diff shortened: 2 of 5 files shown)


domain-models-runtime/src/test/java/org/folio/rest/tools/utils/ModuleName.java
```diff
@@ -0,0 +1,10 @@
+package org.folio.rest.tools.utils;
+
+public class ModuleName {
+  /**
+   * The module name with minus replaced by underscore, for example {@code mod_foo_bar}.
+   */
+  public static String getModuleName() {
```
adamdickmeiss, line 7: I don't like that this class and method is bundled with the runtime.. Can't this just be in test?

julianladisch: For me `domain-models-runtime/src/test` is in test only.
Can you elaborate?

adamdickmeiss: Brain damage on my side.. Excellent.

adamdickmeiss (approved):
LGTM. Should be tested with one real module before merging (if not done already).

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/ModuleName.java
julianladisch: ModuleName is not in the fat jar:
```
unzip -l domain-models-runtime/target/domain-models-runtime-fat.jar|grep -i ModuleName
```

julianladisch:
Successfully tested on mod-inventory-storage, mod-feesfines and mod-login-saml.

PR #947, opened 2021-06-15
RMB-854: Increase FORM_ATTRIBUTE_SIZE_MAX from 16384 to 32768

julianladisch:
Some tenants exceed the current limit of 16384.

Solution: Double the limit.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
@@ -43,7 +43,7 @@ public class RestVerticle extends AbstractVerticle {
 
   public static final Map<String, String> MODULE_SPECIFIC_ARGS  = new HashMap<>(); //NOSONAR
 
-  private static final int FORM_ATTRIBUTE_SIZE_MAX = 16384;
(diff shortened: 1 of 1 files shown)


hjiebsco (approved):
Can we go higher (10x or 5x) to be safe?

adamdickmeiss (approved):

julianladisch:
> Can we go higher (10x or 5x) to be safe?

This limit exists for security reasons and should be as small as possible.

Can you report what your tenants need?

For a clean solution we were making this configurable so that only mod-login-saml uses 32768 and all other modules use the default 2048.

hjiebsco:
> > Can we go higher (10x or 5x) to be safe?
> 
> This limit exists for security reasons and should be as small as possible.
> 
> Can you report what your tenants need?
> 
> For a clean solution we were making this configurable so that only mod-login-saml uses 32768 and all other modules use the default 2048.

The case I encountered was just over 16K. I have no objection to 32K.

steveellis (approved):
Doubling it seems fine since it seems like we know already that 16 may not be enough, especially if we're doing other validation besides size.

PR #948, opened 2021-06-15
RMB-855: Mention that loading db conf from url was removed

julianladisch:


adamdickmeiss (approved):

doc/upgrading.md
```diff
@@ -123,6 +123,12 @@ Command-line option `embed_postgres=true` is no longer supported.
 Remove `<groupId>ru.yandex.qatools.embed</groupId>`
 `<artifactId>postgresql-embedded</artifactId>` from pom.xml.
 
+#### [RMB-855](https://issues.folio.org/browse/RMB-855) Remove loading db conf from url
```
hjiebsco, line 126: Are we going to hijack RMB-855 title from `Cannot load db conf from url anymore` to `Remove loading db conf from url`? Please do not do that. You can list `Remove loading db conf from url` as a bullet item and mention something like `see discussion of RMB-855` or do not even need to mention RMB-855.

PR #958, opened 2021-08-10
RMB-865: Ensure that PgUtil returns 400 (not 500) on invalid CQL

julianladisch:
Add unit tests for invalid CQL that checks that 400 is returned.

No code change to PgUtil because it already returns 400 on invalid CQL:
* get
* getWithOptimizedSql
* streamGet
* delete

However, this pull request makes a PgUtil code change to fix the HTTP status on invalid table name that
should return 500 but has returned 400.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
@@ -736,9 +736,16 @@ public final class PgUtil {
       PreparedCQL preparedCql = new PreparedCQL(table, cqlWrapper, okapiHeaders);
       return get(preparedCql, clazz, collectionClazz, okapiHeaders, vertxContext, responseDelegateClass);
     } catch (Exception e) {
+      final Method respond500;
+      try {
+        respond500 = responseDelegateClass.getMethod(RESPOND_500_WITH_TEXT_PLAIN, Object.class);
+      } catch (Exception e2) {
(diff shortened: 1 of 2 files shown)


skoczko:
@julianladisch you say that there are not code changes but there are clearly changes to PgUtil. I assume this is related to the wrong table name? Shouldn't that be a seperate PR though?

domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java
```diff
@@ -724,30 +724,28 @@ private static void streamTrailer(HttpServerResponse response, ResultInfo result
...
       return get(preparedCql, clazz, collectionClazz, okapiHeaders, vertxContext, responseDelegateClass);
-    } catch (FieldException e) {
+    } catch (Exception e) {
+      final Method respond500;
+      try {
+        respond500 = responseDelegateClass.getMethod(RESPOND_500_WITH_TEXT_PLAIN, Object.class);
+      } catch (Exception e2) {
+        logger.error(e2.getMessage(), e2);
+        return response(e2.getMessage(), null, null);
+      }
```
hjiebsco, line 745: Should this part be moved after Line 746 so the actual `e` won't be lost from the log?

julianladisch: Yes, good catch, fixed.

hjiebsco (approved):

PR #962, opened 2021-08-16
Explain optional fromModuleVersion

julianladisch:


steveellis (approved):
Proof read text and looks good to me. Only noticed that some of numbered items in Readme.md (the text after the '-') start with a capital letter and some do not. Also some end with a period and some do not as well.

adamdickmeiss (approved):

PR #964, opened 2021-09-16
RMB-850: Fix sonarqube bugs + vulnerabilities + security hot spots

SinghAdes:


domain-models-runtime/src/main/java/org/folio/rest/security/AES.java
```diff
@@ -26,26 +26,28 @@ public static void setSecretKey(String key) {
...
     aesCipherForEncryption.init(Cipher.ENCRYPT_MODE, secretKey);
     byte[] byteDataToEncrypt = password.getBytes();
     byte[] byteCipherText = aesCipherForEncryption
         .doFinal(byteDataToEncrypt);
     return byteCipherText;
-  }
+  }*/
 
   /** encrypt a password with the secret key and get back a base64 representation of the password */
+  @SuppressWarnings("all")
```
julianladisch, line 48: Please do not use `@SuppressWarnings("all")`. Instead suppress only one specific warning, for example `@SuppressWarnings("java:S5542")` so that other warnings pop up when we introduce other issues in the code in future.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -114,15 +114,15 @@ private void createFromSnippet() {
       from = new From();
       from.setSnippet(ARRAY_FROM_CLAUSE + "("
         + PostgresClient.DEFAULT_JSONB_FIELD_NAME + GET_JSON_FIELD + arrayField + ")");
-      from.setAsValue(field.get(0).replaceAll("^'|'$", "")); //remove ''
+      from.setAsValue(field.get(0).replace("'", "")); //remove ''
```
julianladisch, line 117: Can you create a separate pull request (mentioning the same RMB-850 issue) for each file when the changes are completely unrelated?

Then we can review and merge them separately.

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -64,6 +64,8 @@ public void handle(AsyncResult<HttpResponse<Buffer>> res) {
         HttpModuleClient2.cache.put(endpoint, cf.get());
       } catch (Exception e) {
         log.error(e.getMessage(), e);
+        //InterruptedExceptions should never be ignored in the code.
+        Thread.currentThread().interrupt();
```
julianladisch, line 68: Please don't interrupt the Thread if e is NOT InterruptedException.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/NetworkUtils.java
```diff
@@ -16,6 +16,10 @@
  */
 public class NetworkUtils {
 
+  /**
+ *  ThreadLocalRandom has a Random instance per thread and safeguards against contention.
+ */
+ @SuppressWarnings("all")
```
julianladisch, line 22: If @SuppressWarnings("all") is replaced by `@SuppressWarnings("java:S2245")` other warnings will still be flagged when we introduce other issues in the code in future.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/NetworkUtils.java
```diff
@@ -16,6 +16,10 @@
  */
 public class NetworkUtils {
 
+  /**
+ *  ThreadLocalRandom has a Random instance per thread and safeguards against contention.
```
julianladisch, line 20: Is this a comment about the method or about the tag `@SuppressWarnings`? If the latter it is more easy to understand for me if moved it into a `//` comment into the tag line.

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
SinghAdes: Sure, i will create a different branch then & this one i will keep it for  reference . thanks for the update.

julianladisch: Thanks, I see you've created multiple branches/pull requests for this.

Some files are in several of the new branches. This is probably not intended.

SinghAdes:
Issue has been resolved.

PR #966, opened 2021-09-20
"replaceAll" modified to replace in Criteria.class

SinghAdes:
As per my understanding,  Line:-117 & 125  "field.get(0).replaceAll("^'|'$", "")" is used for removing single quotes from the field. That can be done with "field.get(0).replace("'", "")" also.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
@@ -114,7 +114,7 @@ public class Criteria {
       from = new From();
       from.setSnippet(ARRAY_FROM_CLAUSE + "("
         + PostgresClient.DEFAULT_JSONB_FIELD_NAME + GET_JSON_FIELD + arrayField + ")");
-      from.setAsValue(field.get(0).replaceAll("^'|'$", "")); //remove ''
+      from.setAsValue(field.get(0).replace("'", "")); //remove ''
     }
   }
 
@@ -122,7 +122,7 @@ public class Criteria {
     if (isArray()) {
       select = new Select();
       //replace surrounding '' from the field name
-      select.setSnippet(field.get(0).replaceAll("^'|'$", ""));
+      select.setSnippet(field.get(0).replace("'", ""));
     }
   }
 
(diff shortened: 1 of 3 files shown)


SinghAdes:
@julianladisch, Raising separate PR for similar issues

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -114,15 +114,15 @@ private void createFromSnippet() {
       from = new From();
       from.setSnippet(ARRAY_FROM_CLAUSE + "("
         + PostgresClient.DEFAULT_JSONB_FIELD_NAME + GET_JSON_FIELD + arrayField + ")");
-      from.setAsValue(field.get(0).replaceAll("^'|'$", "")); //remove ''
+      from.setAsValue(field.get(0).replace("'", "")); //remove ''
     }
```
adamdickmeiss, line 118: Not the same .. Eg the original would not remove quote from `a'b` , but the new one will. I believe all quotes should be removed, so OK by me.

adamdickmeiss (approved):
OK by me (even with the change of behavior).

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
SinghAdes: Ok thanks.

domain-models-runtime/src/main/java/org/folio/rest/persist/facets/FacetField.java
```diff
@@ -23,6 +23,7 @@ public FacetField(String path2facet, int amountOfValues2return){
     this(path2facet, amountOfValues2return, null);
   }
 
+  @SuppressWarnings("java:S5850")//Line:34 (in regex '$' has higher precedence than '|' , but it is a small regex so we can keep it)
```

```

    this.fieldPath = path2facet;
    }
    while (m.find()) {
      this.alias = m.group(1);
    }
  }
```

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -64,6 +64,11 @@ public void handle(AsyncResult<HttpResponse<Buffer>> res) {
         HttpModuleClient2.cache.put(endpoint, cf.get());
       } catch (Exception e) {
         log.error(e.getMessage(), e);
+
+        //InterruptedExceptions should never be ignored in the code
+        if(e instanceof InterruptedException){
+        //Restore interrupted state...
+         Thread.currentThread().interrupt();}
```
```suggestion
        // InterruptedExceptions should never be ignored in the code
        if (e instanceof InterruptedException) {
          // Restore interrupted state
          Thread.currentThread().interrupt();
        }
```

PR #967, opened 2021-09-20
Rmb 850 json parser null condition check

SinghAdes:
At line 277 & 294 , parent may throw NullPointerException. so added a null check condition.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
@@ -274,7 +274,7 @@ public class JsonPathParser {
           parent.rootNode = o;
         }
       }
-      if(returnRoot){
+      if(returnRoot && parent != null){
         parent.requestedValue = ((JsonObject) o).getValue(path);
         return parent;
       }
@@ -291,7 +291,7 @@ public class JsonPathParser {
           parent.rootNode = o;
         }
       }
-      if(returnRoot){
+      if(returnRoot && parent != null){
         parent.requestedValue = o1;
         return parent;
       }
--- a/domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
@@ -274,7 +274,7 @@ public class JsonPathParser {
           parent.rootNode = o;
         }
       }
-      if(returnRoot && parent != null){
+      if (returnRoot && parent != null){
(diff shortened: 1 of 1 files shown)


domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -114,15 +114,15 @@ private void createFromSnippet() {
       from = new From();
       from.setSnippet(ARRAY_FROM_CLAUSE + "("
         + PostgresClient.DEFAULT_JSONB_FIELD_NAME + GET_JSON_FIELD + arrayField + ")");
-      from.setAsValue(field.get(0).replaceAll("^'|'$", "")); //remove ''
+      from.setAsValue(field.get(0).replace("'", "")); //remove ''
```
julianladisch, line 117: The old code removes leading and tailing single quotes: `'Rock'n'Roll'` results in `Rock'n'Roll`

The new code removes all single quotes: `'Rock'n'Roll'` results in `RocknRoll`

I would introduce a new method for the task:

```
  private static String unwrapFromSingleQuotes(String s) {
    if (s.length() >= 2 && s.startsWith("'") && s.endsWith("'")) {
      return s.substring(1, s.length() - 1);
    }
    return s;
  }
```

domain-models-runtime/src/main/java/org/folio/rest/persist/facets/FacetField.java
```diff
@@ -23,6 +23,7 @@ public FacetField(String path2facet, int amountOfValues2return){
     this(path2facet, amountOfValues2return, null);
   }
 
+  @SuppressWarnings("java:S5850")//Line:34 (in regex '$' has higher precedence than '|' , but it is a small regex so we can keep it)
```

```

    this.fieldPath = path2facet;
    }
    while (m.find()) {
      this.alias = m.group(1);
    }
  }
```

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -64,6 +64,11 @@ public void handle(AsyncResult<HttpResponse<Buffer>> res) {
         HttpModuleClient2.cache.put(endpoint, cf.get());
       } catch (Exception e) {
         log.error(e.getMessage(), e);
+
+        //InterruptedExceptions should never be ignored in the code
+        if(e instanceof InterruptedException){
+        //Restore interrupted state...
+         Thread.currentThread().interrupt();}
```
```suggestion
        // InterruptedExceptions should never be ignored in the code
        if (e instanceof InterruptedException) {
          // Restore interrupted state
          Thread.currentThread().interrupt();
        }
```

julianladisch (approved):
Looks good, thanks!

PR #968, opened 2021-09-20
Rmb 850 fixing security hotspots

SinghAdes:
added @SuppresWarning with comment .


domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -64,6 +64,11 @@ public void handle(AsyncResult<HttpResponse<Buffer>> res) {
         HttpModuleClient2.cache.put(endpoint, cf.get());
       } catch (Exception e) {
         log.error(e.getMessage(), e);
+
+        //InterruptedExceptions should never be ignored in the code
+        if(e instanceof InterruptedException){
```
adamdickmeiss, line 69:
```suggestion
        if (e instanceof InterruptedException){
```

domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
```diff
@@ -274,7 +274,7 @@ private Object getValueAt(String path, Object o, Pairs parent, boolean returnRoo
           parent.rootNode = o;
         }
       }
-      if(returnRoot){
+      if(returnRoot && parent != null){
```
adamdickmeiss, line 277:
```suggestion
      if (returnRoot && parent != null) {
```

domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
```diff
@@ -291,7 +291,7 @@ else if(o instanceof JsonArray){
...
```
adamdickmeiss, line 294:
```suggestion
      if (returnRoot && parent != null) {
```

domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
SinghAdes: @adamdickmeiss , committed suggested changes. thanks.

adamdickmeiss (approved):

domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java
julianladisch: Can you push the commit?

domain-models-runtime/src/main/java/org/folio/rest/tools/client/HTTPJsonResponseHandler.java
```diff
@@ -64,6 +64,11 @@ public void handle(AsyncResult<HttpResponse<Buffer>> res) {
         HttpModuleClient2.cache.put(endpoint, cf.get());
       } catch (Exception e) {
         log.error(e.getMessage(), e);
+
+        //InterruptedExceptions should never be ignored in the code
+        if (e instanceof InterruptedException){
+        //Restore interrupted state...
+         Thread.currentThread().interrupt();}
```
```suggestion
        // InterruptedExceptions should never be ignored in the code
        if (e instanceof InterruptedException) {
          //Restore interrupted state
          Thread.currentThread().interrupt();
        }
```

dbschema/src/main/java/org/folio/dbschema/Index.java
```diff
@@ -93,6 +93,7 @@ public void setSqlExpressionQuery(String sqlExpressionQuery) {
     this.sqlExpressionQuery = sqlExpressionQuery;
   }
 
+  @SuppressWarnings("java:S5852")//Line:105-Most of the regular expression engines use backtracking, but here regex is very small so we can keep it.
```
julianladisch, line 96: The split regexp tries to do two tasks in one step:
1. Split on comma
2. Trim leading and tailing spaces.

The implementation `split(" *, *")` has a bug because it doesn't trim leading spaces of the first name and tailing spaces of the last name.

How about this code that avoids suppressing the regexp warning, fixes the bug, and is easier to read for me:

```
    String [] splitIndex = this.getMultiFieldNames().split(",");
    Arrays.asList(splitIndex).replaceAll(s -> s.strip());
```

domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java
```diff
@@ -114,15 +114,15 @@ private void createFromSnippet() {
       from = new From();
       from.setSnippet(ARRAY_FROM_CLAUSE + "("
         + PostgresClient.DEFAULT_JSONB_FIELD_NAME + GET_JSON_FIELD + arrayField + ")");
-      from.setAsValue(field.get(0).replaceAll("^'|'$", "")); //remove ''
+      from.setAsValue(field.get(0).replace("'", "")); //remove ''
```
julianladisch, line 117: The old code removes leading and tailing single quotes: `'Rock'n'Roll'` results in `Rock'n'Roll`

The new code removes all single quotes: `'Rock'n'Roll'` results in `RocknRoll`

I would introduce a new method for the task:

```
  private static String unwrapFromSingleQuotes(String s) {
    if (s.length() >= 2 && s.startsWith("'") && s.endsWith("'")) {
      return s.substring(1, s.length() - 1);
    }
    return s;
  }
```

domain-models-runtime/src/main/java/org/folio/rest/persist/facets/FacetField.java
```diff
@@ -23,6 +23,7 @@ public FacetField(String path2facet, int amountOfValues2return){
     this(path2facet, amountOfValues2return, null);
   }
 
+  @SuppressWarnings({"java:S5850","java:S5852"})//Line:34 (in regex '$' has higher precedence than '|' , but it is a small regex so we can keep it)
```

```

    this.fieldPath = path2facet;
    }
    while (m.find()) {
      this.alias = m.group(1);
    }
  }
```

SinghAdes: Thanks , i will make changes and let you know.

domain-models-runtime/src/main/java/org/folio/rest/tools/utils/NetworkUtils.java
```diff
@@ -16,6 +16,7 @@
  */
 public class NetworkUtils {
 
+  @SuppressWarnings({"java:S2245","java:S2245"})//ThreadLocalRandom are not cryptographically secure, but performance is better in case of concurrent access.
```
julianladisch, line 19: This suppresses java:S2245 twice.

julianladisch, line 19: `ThreadLocalRandom are not cryptographically secure, but performance is better in case of concurrent access.`

I would replace it by

`Suppress "Weak Cryptography" warning, ThreadLocalRandom is good enough for finding a free port`

SinghAdes:
Issue has been resolved.

PR #974, opened 2021-09-21
Dropped Regex

SinghAdes:
Line:366- if(subPathsList[i+1].matches(".*\\[.*?\\].*"))  dropped. 
As per my understanding above regex matches string with in square brackets.


adamdickmeiss (approved):

julianladisch (approved):
Thanks, looks good.

PR #982, opened 2021-10-21
Improve main README readability

ncovercash:
The main point of this PR is to allow the README file to be read in dark mode.  With the current diagrams (and code snippets as images), there is no background, making it impossible to see the actual content:
image

This PR largely moves the content out of the images (making it more accessible in general, providing a better experience at different screen sizes, allowing find/copy/paste, and not to mention supporting light/dark mode).

The `build.png` and `generate.png` retained the diagram as text would not be enough to convey their ideas, however, the other images were replaced by actual code snippets in the markdown.   The removed images were either entirely code snippets, or code snippets with little annotation that could easily be replicated without being too verbose.

Finally, a few additional fixes were made, primarily just adding rich syntax highlighting to code blocks and moving inline code not in blocks into them.  Visiting the README in the fork allows you to view the new version.

No significant changes were made to the meaning of anything, only visual improvements,


dcrossleyau (approved):
Thanks again.

PR #984, opened 2021-11-01
RMB-664: One connection pool for all tenants

julianladisch:
DB_MAXSHAREDPOOLSIZE uses "SET ROLE ..." to switch to a tenant specific database role with restricted permission.

Vert.x prints a warning "Parameter is_superuser changed to off" or "Parameter is_superuser changed to on" since 4.1.0: https://github.com/eclipse-vertx/vertx-sql-client/commit/0bd1571d033ba35f57d188472c902b81b35dbd11
This should be changed in Vert.x: https://github.com/eclipse-vertx/vertx-sql-client/pull/1060

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -19,6 +19,7 @@ import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowIterator;
 import io.vertx.sqlclient.RowSet;
 import io.vertx.sqlclient.RowStream;
+import io.vertx.sqlclient.SqlConnection;
 import io.vertx.sqlclient.Transaction;
 import io.vertx.sqlclient.Tuple;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -104,9 +104,19 @@ private static void deployRestVerticle(TestContext context) {
    * Cleanup: Delete temporary file, close the vert.x instance.
    *
    * @param context  the test context
+   * @throws Throwable
    */
   @AfterClass
-  public static void tearDown(TestContext context) {
+  public static void tearDown(TestContext context) throws Throwable {
+    String sql = "select grantee, table_catalog, privilege_type, table_schema, table_name "
```
adamdickmeiss, line 111: grantee ?

julianladisch: debug code removed

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -116,6 +128,11 @@
 
   private static String          configPath               = null;
 
+  /**
+   * Used only if {@link #sharedPgPool} is true.
+   */
+  private static Map<Vertx,PgPool> pgPools = new HashMap<>();
```
adamdickmeiss, line 134: Could be final. (and so could connectionPool below)

adamdickmeiss (commented):
Looks good. I tried this branch and with mod-permissions and ran it with the shared pool and
is getting a new warning from mod-permissions log (RMB).

``
14:49:20 [] [] [] [] INFO  PermsAPI             Generating cql to request rows from table 'permissions' with query '(dummy == false)'
[Mon Nov 01 14:49:20 CET 2021] org.folio.cql2pgjson.CQL2PgJSON INFO loadDbSchema: Loaded templates/db_scripts/schema.json OK 
[Mon Nov 01 14:49:20 CET 2021] io.vertx.pgclient.impl.codec.PgCommandCodec WARNING Parameter is_superuser changed to off 
``

julianladisch:
> WARNING Parameter is_superuser changed to off

Yes, this should be fixed in Vert.x, see this PR description.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Thanks, fixed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -116,7 +128,12 @@
 
   private static String          configPath               = null;
 
-  private static MultiKeyMap<Object, PostgresClient> connectionPool = MultiKeyMap.multiKeyMap(new HashedMap<>());
+  /**
+   * Used only if {@link #sharedPgPool} is true.
+   */
+  private static final Map<Vertx,PgPool> pgPools = new HashMap<>();
+  /** map (Vertx, String tenantId) to PostgresClient */
```
adamdickmeiss, line 135: And now, SQ complains.. I wished it would only do that for public static final(s).

adamdickmeiss (approved):

julianladisch:
The misleading "WARNING Parameter is_superuser changed to off" is fixed in Vert.x 4.2.1: https://github.com/eclipse-vertx/vertx-sql-client/pull/1060

PR #985, opened 2021-11-01
RMB-718: Implement totalRecords=none disabling count_estimate() DB query

julianladisch:
https://issues.folio.org/browse/RMB-718 "Avoid calling count_estimate() if GET collection API always return 1 record"
https://issues.folio.org/browse/RMB-724 "totalRecords=exact/estimated/none/auto"

This pull request implements RMB-718 only, this is half of RMB-724: totalRecords=none/auto

There will be a separate PR for "Upgrade notes about language.raml trait removal and new totalRecords": https://issues.folio.org/browse/RMB-879

### DIFF
--- a/DB-schema-migration.md
+++ b/DB-schema-migration.md
@@ -6,6 +6,9 @@ This documentation includes information about possible strategies regarding DB s
 
 For RMB based modules a DB migration will be done automatically based on the content of the schema.json provided within the module. Indeed there are exceptions, but each such case must be considered separately.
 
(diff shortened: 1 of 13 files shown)


adamdickmeiss (commented):
Shouldn't there also be a raml PR for pageable.raml (which has limit and offset, so far). Should probably be merged before this one, so that this branch can be updated to use it.

adamdickmeiss (approved):

PR #987, opened 2021-11-04
MODSOURCE-392: No backward-compatible database migration

julianladisch:


adamdickmeiss (approved):
Confusing that this refer to a module (rather than RMB, FOLIO or similar).

javalkar-varun (approved):

PR #992, opened 2021-11-30
RMB-881: Reduce logging level of loadDbSchema

dltj:
Fixes CQLPG-105 = https://issues.folio.org/browse/RMB-881


julianladisch (approved):
Looks good, thanks!

adamdickmeiss (approved):

PR #995, opened 2021-12-02
RMB 817

MikhailFokanov:


domain-models-runtime/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.folio</groupId>
     <artifactId>raml-module-builder</artifactId>
-    <version>33.3.0-SNAPSHOT</version>
+    <version>33.4.0-SNAPSHOT</version>
```
julianladisch, line 9: This is a breaking change requiring a major version bump: 33 -> 34.

MikhailFokanov-ebsco: Done. Could you please review?

adamdickmeiss:
Merged without an approval ? ANyway.. I don't like that you update NEWS. it's not very helpful. The date is wrong. The RMB-817 in Jira has 34.0.0 fix version, that's all we need.

PR #997, opened 2021-12-10
RMB-887: deadlock in saveBatch, updateBatch, upsertBatch

julianladisch:
Add "SELECT ... FOR UPDATE" in saveBatch, updateBatch, upsertBatch in PostgresClient and Conn.

This waits until it can lock all affected records.
Then multiple transactions don't block each other (deadlock) when trying to update the same records.


adamdickmeiss (approved):
OK. Not that I fully understand why one wants to lock many records. Sounds like a recipe for a slow system.

julianladisch:
I've added performance numbers to https://issues.folio.org/browse/RMB-887

PR #1001, opened 2021-12-14
Fix no context parameters logged RMB-890

adamdickmeiss:
https://issues.folio.org/browse/RMB-890

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/RestRouting.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/RestRouting.java
@@ -433,7 +433,9 @@ public final class RestRouting {
   static void invoke(Method method, Object[] params, Object o, RoutingContext rc,
                      Map<String, String> headers, Handler<AsyncResult<Response>> resultHandler) {
 
-    headers.forEach(FolioLoggingContext::put);
+    FolioLoggingContext.put(FolioLoggingContext.TENANT_ID_LOGGING_VAR_NAME, headers.get(RestVerticle.OKAPI_HEADER_TENANT));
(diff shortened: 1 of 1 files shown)


domain-models-runtime/src/main/java/org/folio/rest/RestRouting.java
```diff
@@ -433,7 +433,9 @@ private static void parseParams(RoutingContext rc, Buffer body, Iterator<Map.Ent
   static void invoke(Method method, Object[] params, Object o, RoutingContext rc,
                      Map<String, String> headers, Handler<AsyncResult<Response>> resultHandler) {
 
-    headers.forEach(FolioLoggingContext::put);
+    FolioLoggingContext.put(FolioLoggingContext.TENANT_ID_LOGGING_VAR_NAME, headers.get(RestVerticle.OKAPI_HEADER_TENANT));
+    FolioLoggingContext.put(FolioLoggingContext.REQUEST_ID_LOGGING_VAR_NAME, headers.get(RestVerticle.OKAPI_REQUESTID_HEADER));
+    FolioLoggingContext.put(FolioLoggingContext.USER_ID_LOGGING_VAR_NAME, headers.get(RestVerticle.OKAPI_USERID_HEADER));
```
julianladisch, line 438: Can you also put MODULE_ID_LOGGING_VAR_NAME for https://github.com/folio-org/raml-module-builder/blob/v33.2.2/domain-models-runtime/src/main/resources/log4j2.properties#L15 ? org.folio.rest.tools.utils.ModuleName provides the name.

adamdickmeiss: ok

julianladisch (approved):
Looks good, thanks!

PR #1009, opened 2021-12-22
Remove deprecated get, streamGet methods RMB-885

adamdickmeiss:
https://issues.folio.org/browse/RMB-885

### DIFF
--- a/doc/upgrading.md
+++ b/doc/upgrading.md
@@ -4,6 +4,7 @@ These are notes to assist upgrading to newer versions.
 See the [NEWS](../NEWS.md) summary of changes for each version.
 
 <!-- ../../okapi/doc/md2toc -l 2 -h 3 upgrading.md -->
+* [Version 34.0](#version-340)
 * [Version 33.2](#version-332)
 * [Version 33.1](#version-331)
 * [Version 33.0](#version-330)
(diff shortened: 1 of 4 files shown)


doc/upgrading.md
```diff
@@ -21,6 +22,40 @@ See the [NEWS](../NEWS.md) summary of changes for each version.
...
+   Handler<AsyncResult<Results<T>>> replyHandler)
+
+<T> void get(String table, Class<T> clazz, String[] fields, String filter,
+   boolean returnCount, boolean setId,
+   Handler<AsyncResult<Results<T>>> replyHandler)
+
+<T> void get(String table, Class<T> clazz, String filter,
+   boolean returnCount, boolean setId,
+   Handler<AsyncResult<Results<T>>> replyHandler)
+```
```
julianladisch, line 57:
```suggestion
The `PostgresClient.streamGet` method without `PostgresClientStreamResult` parameter has been removed because it doesn't return totalCount and consumes too much memory. Use one of the other `PostgresClient.streamGet` methods with `PostgresClientStreamResult`.

`PostgresClient.get` methods with `String where` or `String filter` parameter have been removed to reduce SQL injection issues. Use `PostgresClient.get` methods with `CQLWrapper` or `Criterion` parameter instead.
```

Mentioning reasons and what to use instead.

The build will fail with a compile error showing where a removed method is called, no need to give the full details here.

julianladisch (approved):
Looks good, thanks!

PR #1015, opened 2022-01-26
Reword that the raml submodule upgrade for totalRecords is optional

julianladisch:


adamdickmeiss (approved):

marcjohnson-kint (approved):
Thanks @julianladisch for being willing to make it more prominent that this aspect is optional

PR #1018, opened 2022-02-03
Upgrade org.postgresql:postgresql

adamdickmeiss:
This is only used in testing.
See https://github.com/advisories/GHSA-v7wg-cpwc-24m4


julianladisch (approved):
Looks good, thanks.

PR #1021, opened 2022-02-07
Use XOkapiHeaders RMB-897

adamdickmeiss:
https://issues.folio.org/browse/RMB-897

Should this PR also change the generated clients that uses "X-Okapi-Token", "X-Okapi-Url" literals?


domain-models-runtime/src/main/java/org/folio/rest/RestRouting.java
```diff
@@ -504,18 +505,15 @@ private static void handleStream(Method method2Run, RoutingContext rc, Object in
     });
   }
 
-  static void getOkapiHeaders(RoutingContext rc, Map<String, String> headers, String[] tenantId) {
-    MultiMap mm = rc.request().headers();
+  static String getOkapiHeaders(MultiMap mm, Map<String, String> headers) {
```
julianladisch, line 508: static CaseInsensitiveMap<String, String> getOkapiHeaders(MultiMap mm) {

and

    Map<String, String> okapiHeaders = getOkapiHeaders(rc.request().headers());
    String tenantId = okapiHeaders.get(XOkapiHeaders.TENANT);

is easier to understand.

julianladisch (approved):
Looks good, thanks!

domain-models-runtime/src/main/java/org/folio/rest/RestRouting.java
adamdickmeiss: agreed.

PR #1024, opened 2022-02-10
Delete deprecated RollBackURL RMB-884

adamdickmeiss:


doc/upgrading.md
```diff
@@ -45,6 +45,11 @@ from `okapi-testing`. It has same API. Just import
 TenantLoading methods `addJsonIdContent` and `addJsonIdBasename`
 are removed.
 
+#### [RMB-884](https://issues.folio.org/browse/RMB-884)
+
+Class `RollBackURL' removed - including methods in interface
+`HttpClientInterface` using this class.
```
julianladisch, line 51: Can you remove this from the upgrading.md? This information can be put into the release notes, it is not needed for a developer that upgrades a module to this RMB version.

A GitHub search shows that RollBackURL hasn't been used, and we know that it is a skeleton without working implementation. The compiler will show any usage of RollBackURL. We don't want any developer running "grep -r -i RollBackURL" to find usage.

adamdickmeiss: OK. I think having both upgrading *and* news is misleading. Perhaps I should stop resisting updating news along the way.

julianladisch (approved):
Looks good, thanks!

PR #1025, opened 2022-02-10
RMB-547: PostgreSQL SSL CA Certificate configuration option

julianladisch:
Add an option to the PostgreSQL configuration that contains that ca .pem certificate.
If that option is used the module requires a SSL/TLS connection and a valid server certificate when connecting to PostgreSQL by setting sslmode=verify-full.

If the option is set:
* Connection is rejected if connection is unencrypted.
* Connection is rejected if connection uses SSLv2Hello, SSLv3, TLSv1, TLSv1.1 or TLSv1.2 (only TLSv1.3 is allowed)
* Connection is rejected if server certificate is valid but is not issued for the hostname configured in DB_HOST


adamdickmeiss (approved):
I wished the setting was called `DB_SERVER_PEM` as that's what it's called in in folio-vertx-lib. It's also called `postgres_server_pem` property in Okapi. Should you choose to merge with DB_SERVERPEM, then I'll make a PR on folio-vertx-lib to make it the same.

julianladisch:
> DB_SERVER_PEM

Good catch, thanks!
I've renamed it to be consistent with Okapi and folio-vertx-lib.

PR #1026, opened 2022-02-10
Make HttpServerOptions configurable RMB-856

adamdickmeiss:
https://issues.folio.org/browse/RMB-856

Only mod-login-saml needs this and it needs a high value. I don't think it needs to be configurable - even for mod-login-saml, but of course higher than the default.


domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -45,11 +45,11 @@
 
   public static final Map<String, String> MODULE_SPECIFIC_ARGS  = new HashMap<>(); //NOSONAR
 
-  private static final int FORM_ATTRIBUTE_SIZE_MAX = 32768;
   private static final String       HTTP_PORT_SETTING               = "http.port";
   private static final Logger       log                             = LogManager.getLogger(RestVerticle.class);
   private static String             deploymentId                     = "";
 
+  private static HttpServerOptions httpServerOptions = new HttpServerOptions();
```
julianladisch, line 52:
```suggestion
  // default options for server; may be changed in init hook
  private static HttpServerOptions httpServerOptions = new HttpServerOptions().setCompressionSupported(true);
```
A module might want to change it while processing config parameters before the init hook is called.

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
```diff
@@ -81,27 +81,23 @@ public void start(Promise<Void> startPromise) throws Exception {
...
     router.route("/apidocs/*").handler(StaticHandler.create("apidocs"));
-    // startup http server on port 8181 to serve documentation
 
-    //if client includes an Accept-Encoding header which includes
-    //the supported compressions - deflate or gzip.
-    HttpServerOptions serverOptions = new HttpServerOptions();
-    serverOptions.setCompressionSupported(true);
-    serverOptions.setMaxFormAttributeSize(FORM_ATTRIBUTE_SIZE_MAX);
+    // default options for server; may be changed in init hook
+    httpServerOptions.setCompressionSupported(true);
```
julianladisch, line 86:
```suggestion
```
Move to the variable declaration.

julianladisch (approved):
Looks good, nice solution, thanks!

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
adamdickmeiss: right. yes.

domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java
adamdickmeiss: yep

PR #1030, opened 2022-03-02
RMB-901: reconnectAttempts and reconnectInterval for PgConnectOptions

julianladisch:
https://issues.folio.org/browse/RMB-901

DB timeout issues happen most often when the system is not warmed up and
connection pool does not contain any open connections.

### DIFF
--- a/domain-models-runtime/src/test/java/org/folio/rest/tools/utils/EnvsTest.java
+++ b/domain-models-runtime/src/test/java/org/folio/rest/tools/utils/EnvsTest.java
@@ -27,6 +27,8 @@ public class EnvsTest {
     map.put("DB_QUERYTIMEOUT", "8");
     map.put("DB_MAXPOOLSIZE", "5");
     map.put("DB_CONNECTIONRELEASEDELAY", "12345");
+    map.put("DB_RECONNECTATTEMPTS", "3");
+    map.put("DB_RECONNECTINTERVAL", "2000");
     map.put("DB_EXPLAIN_QUERY_THRESHOLD", "100");
     // we dropped support for dot form. check that it is ignored
     map.put("db.username", "superwoman");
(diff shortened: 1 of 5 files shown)


adamdickmeiss (commented):
I'm not sure I like things like that to be configurable. It could be as sign of a abuse of PG usage in general. For example , parallel usage (where it shouln't be etc). Even if this is used, it will, most likely just postpone a bit (problem will still be there).

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/EnvsTest.java
```diff
@@ -27,6 +27,8 @@ public void setUp() {
     map.put("DB_QUERYTIMEOUT", "8");
     map.put("DB_MAXPOOLSIZE", "5");
     map.put("DB_CONNECTIONRELEASEDELAY", "12345");
+    map.put("DB_RECONNECTATTEMPTS", "3");
+    map.put("DB_RECONNECTINTERVAL", "2000");
     map.put("DB_EXPLAIN_QUERY_THRESHOLD", "100");
     // we dropped support for dot form. check that it is ignored
```
steveellis, line 33: Minor. Maybe revise comment to `We dropped support for dot form. check that it is ignored.`

steveellis (approved):

julianladisch:
> It could be as sign of a abuse of PG usage in general.
> most likely just postpone a bit (problem will still be there)

Folijet's requests indicates that the connection will succeed on second try: https://issues.folio.org/projects/RMB/issues/RMB-901

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/EnvsTest.java
julianladisch: Out of scope and not applicable for reconnection.

Create a separate pull request if you want this been changed.

adamdickmeiss:
Don't hold back on me on merge. The code is fine for this.. I just wished we'd use environment for real cases of configuration (like a host name!), and not for things that make modules work. So if some modules require this (or all), perhaps it's better -long term - to actually have this "hard-coded" or just a property..

PR #1033, opened 2022-03-05
RMB-905: NetworkUtilsTest

julianladisch:


adamdickmeiss (approved):
The algorithm is interesting.. you can call it a few times and hope that it will return a different port from last time.. And even if one port is returned, there's no guarantee somebody else hasn't grabbed it before you start the HTTP service (or other).

PR #1035, opened 2022-03-08
Rmb 901 reconnect options rmb 33.2.7

julianladisch:


adamdickmeiss (approved):
I'd call this a feature, but I understand that this may fix an issue for a module.

PR #1038, opened 2022-03-30
RMB-907: Move jackson-databind entry before jackson-bom

julianladisch:
The first entry takes precendence over later entries in pom.xml.

This test shows that it is correct now:

mvn dependency:tree -Dincludes=com.fasterxml.jackson.core:jackson-databind


adamdickmeiss (approved):
Voila!

PR #1043, opened 2022-04-22
RMB-907: b33.2 jackson-databind 2.13.2.1

julianladisch:
Move jackson entry before vertx entry like it is on master.
Otherwise jackson version provided by vertx overwrites the
patched version.


adamdickmeiss (approved):
ok

PR #1047, opened 2022-05-04
RMB-348: Add support for database read/write splitting to RAML Module Builder

mtraneis:


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -436,9 +455,9 @@ public static void closeAllClients() {
     PG_POOLS.clear();
   }
 
-  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig) {
+  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig, String desiredHost) {
     PgConnectOptions pgConnectOptions = new PgConnectOptions();
-    String host = sqlConfig.getString(HOST);
+    String host = sqlConfig.getString(desiredHost);
```
julianladisch, line 460: Can we be sure that the hostname is different and the port is the same?
Or is a different port on the same host a valid use case?

    pgproxy:8000
    pgproxy:8001

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -506,14 +525,16 @@ private void init() throws Exception {
     logPostgresConfig();
 
     if (sharedPgPool) {
-      client = PG_POOLS.computeIfAbsent(vertx, x -> createPgPool(vertx, postgreSQLClientConfig));
+      client = PG_POOLS.computeIfAbsent(vertx, x -> createPgPool(vertx, postgreSQLClientConfig, HOST));
+      clientReader = PG_POOLS.computeIfAbsent(vertx, x -> createPgPool(vertx, postgreSQLClientConfig, HOST_READER));
```
julianladisch, line 529: clientReader requires a PG_POOLS_READER

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -142,6 +143,8 @@ public class PostgresClient {
   /** map (Vertx, String tenantId) to PostgresClient */
   private static final MultiKeyMap<Object, PostgresClient> CONNECTION_POOL =
       MultiKeyMap.multiKeyMap(new HashedMap<>());
+  private static final MultiKeyMap<Object, PostgresClient> CONNECTION_POOL_READER =
+      MultiKeyMap.multiKeyMap(new HashedMap<>());
```
julianladisch, line 147: By default CONNECTION_POOL_READER should equal CONNECTION_POOL so that we have a single pool as before. Only when there's reader settings in the configuration the second pool should be created.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -350,6 +354,21 @@ static String createPassword(String password) throws Exception {
...
+   * This instance's PgPool for database connections.
+   *
+   * Take care to execute "SET ROLE ..." when {@link #sharedPgPool} is true; consider using
+   * one of the with... methods that automatically execute "SET ROLE ...".
+   *
+   * @see #withConnection(Function)
+   * @see #withConn(Function)
+   * @see #withTrans(Function)
+   * @see #withTransaction(Function)
+   */
```
julianladisch, line 367: The javadoc of getClient() and getReaderClient() should explain how they differ and refer to each other (`@link` or `@see`.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -350,6 +354,21 @@ static String createPassword(String password) throws Exception {
...
+  PgPool getReaderClient() {
+    return clientReader;
+  }
```
julianladisch, line 370: There should also be a setReaderClient()

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Or by default CONNECTION_POOL_READER should be null, and if null clientReader is set to client.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -156,6 +159,7 @@ public class PostgresClient {
   private final Vertx vertx;
   private JsonObject postgreSQLClientConfig = null;
   private PgPool client;
+  private PgPool clientReader;
```
julianladisch, line 162: Can you add javadoc to each of these variables explaining their differences?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -506,14 +525,16 @@ private void init() throws Exception {
...
-      client = createPgPool(vertx, postgreSQLClientConfig);
+      client = createPgPool(vertx, postgreSQLClientConfig, HOST);
+      clientReader = createPgPool(vertx, postgreSQLClientConfig, HOST_READER);
     }
   }
 
-  static PgPool createPgPool(Vertx vertx, JsonObject configuration) {
-    PgConnectOptions connectOptions = createPgConnectOptions(configuration);
+  static PgPool createPgPool(Vertx vertx, JsonObject configuration, String desiredHost) {
+    PgConnectOptions connectOptions = createPgConnectOptions(configuration, desiredHost);
```
julianladisch, line 537: closeClient() should also close the reader client.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3326,6 +3347,26 @@ public Future<PgConnection> getConnection() {
         .map((PgConnection) sqlConnection));
   }
 
+  /**
+   * Get vertx-pg-client connection
+   *
+   * @see #withConn(Function)
+   * @see #withConnection(Function)
+   * @see #withTrans(Function)
+   * @see #withTransaction(Function)
```
julianladisch, line 3356: There are many methods that lists all with* methods in their javadoc. These lists should be extended by the withRead* methods.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3496,6 +3609,21 @@ public <T> Future<T> withConn(Function<Conn, Future<T>> function) {
...
+   * <ul>
+   *   <li>The connection is automatically closed in all cases when the function exits.</li>
+   *   <li>The method returns the Future returned by the function, or a failed Future with the Throwable
+   *   thrown by the function.</li>
+   * </ul>
+   *
+   * @param function code to execute
+   */
+  public <T> Future<T> withReadConn(Function<Conn, Future<T>> function) {
+    return withConn(0, function);
```
julianladisch, line 3624: withReadConn

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3326,6 +3347,26 @@ public Future<PgConnection> getConnection() {
         .map((PgConnection) sqlConnection));
   }
 
+  /**
+   * Get vertx-pg-client connection
```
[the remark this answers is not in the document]
julianladisch: An a similar javadoc change for all other read methods.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3399,6 +3469,49 @@ void getSQLConnection(int queryTimeout, Handler<AsyncResult<SQLConnection>> hand
...
+                  queryTimeout));
+        } else {
+          log.warn("Failed to send cancelling request", ar.cause());
+        }
+      }));
+
+      SQLConnection sqlConnection = new SQLConnection(pgConnection, null, timerId);
+      handler.handle(Future.succeededFuture(sqlConnection));
+    });
+  }
```
julianladisch, line 3513: The duplicate code should be extracted into a method.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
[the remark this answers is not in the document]
julianladisch: Or readClient

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch, line 147: CONNECTION_POOL_READER is not needed because `PostgresClient` is never read-only, it always support both read-only and read+write.
Instead a PG_POOLS_READER is needed because we need read-only `PgPool`s.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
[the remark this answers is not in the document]
julianladisch: Could you also make similar javadoc changes for all other read methods?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
mtraneis: As far as I know, if using RDS then the "reader" port will be the same. But we could code for the case of the port being different. Is this a likely scenario?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
mtraneis: Yes, I will fix these comments. I was trying to copy & paste & make minor changes to get an idea of how it could work.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
mtraneis: done

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3512,6 +3640,22 @@ public <T> Future<T> withConn(int queryTimeout, Function<Conn, Future<T>> functi
...
+   *   thrown by the function.</li>
+   * </ul>
+   *
+   * @param queryTimeout in milliseconds, 0 for no timeout
+   * @param function code to execute
+   */
+  public <T> Future<T> withReadConn(int queryTimeout, Function<Conn, Future<T>> function) {
+    return withReadConnection(pgConnection -> withTimeout(pgConnection, queryTimeout, function));
+  }
+
```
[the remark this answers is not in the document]
mtraneis: Fixed

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -434,15 +490,22 @@ public static void closeAllClients() {
     }
     PG_POOLS.values().forEach(PgPool::close);
     PG_POOLS.clear();
+    PG_POOLS_READER.values().forEach(PgPool::close);
+    PG_POOLS_READER.clear();
   }
 
-  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig) {
+  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig, String desiredHost) {
```
julianladisch, line 497: Replacing `String desiredHost` by `boolean isReader` might be easier to understand and implement.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -374,15 +418,27 @@ void setClient(PgPool client) {
...
+   * Close the SQL client of this PostgresClient instance.
+   * This is idempotent: additional close invocations are always successful.
+   */
+  public Future<Void> closeClient(PgPool theClient) {
+    if (theClient == null) {
       return Future.succeededFuture();
     }
-    PgPool clientToClose = client;
-    client = null;
+    PgPool clientToClose = theClient;
```
julianladisch, line 441: This variable is not needed, use `theClient`.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: Example:
```
  void getSQLConnection(int queryTimeout, Handler<AsyncResult<SQLConnection>> handler) {
    getConnection(res -> getSQLConnection(res, queryTimeout, handler));
  }

  void getSQLReadConnection(int queryTimeout, Handler<AsyncResult<SQLConnection>> handler) {
    getReadConnection(res -> getSQLConnection(res, queryTimeout, handler));
  }

  void getSQLConnection(AsyncResult<PgConnection> res, int queryTimeout, Handler<AsyncResult<SQLConnection>> handler) {
    if (res.failed()) {
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3533,6 +3759,17 @@ public <T> Future<T> withConn(AsyncResult<SQLConnection> sqlConnection, Function
...
+  /**
+   * Take the connection from the {@link SQLConnection}, wrap it into a {@link Conn} and execute the function.
+   *
+   * @return the result from the function, the failure of sqlConnection or any thrown Throwable.
+   */
+  @SuppressWarnings("squid:S1181")  // suppress "Throwable and Error should not be caught"
+  // because a Future also handles Throwable, this is required for asynchronous reporting
+  public <T> Future<T> withReadConn(AsyncResult<SQLConnection> sqlConnection, Function<Conn, Future<T>> function) {
+    return withConn(sqlConnection, function);
+  }
```
julianladisch, line 3771: This method is not needed. `withConn` can directly be used.

mtraneis: It's needed because of this: 
https://github.com/folio-org/raml-module-builder/pull/1047#discussion_r866189399

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -392,20 +456,58 @@ public Future<Void> closeClient() {
...
+      return Future.succeededFuture();
+    }
+
+    PgPool clientToClose = client;
+    PgPool readClientToClose = client == readClient ? null : readClient;
+    client = null;
+    readClient = null;
+
+    closeClient(clientToClose);
+    return closeClient(readClientToClose);
```
adamdickmeiss, line 470:
```suggestion
    return closeClient(clientToClose).compose(x -> closeClient(readClientToClose));
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -392,20 +456,58 @@ public Future<Void> closeClient() {
...
+    /*
+    return CompositeFuture.all(closeClient(clientToClose), closeClient(readClientToClose))
+        .mapEmpty();
+    */
```
adamdickmeiss, line 474: Commented out? What's wrong with it? The compose suggested before is just sequental and here "concurrent".

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -392,20 +456,58 @@ public Future<Void> closeClient() {
...
    * Close the SQL client of this PostgresClient instance.
    * This is idempotent: additional close invocations are always successful.
    * @param whenDone invoked with the close result
    */
   public void closeClient(Handler<AsyncResult<Void>> whenDone) {
-    closeClient().onComplete(whenDone);
+    //closeClient().onComplete(whenDone);
+
+    closeReadClient().onComplete(whenDone);
+    closeWriteClient().onComplete(whenDone);
```
adamdickmeiss, line 502: whenDone called twice

adamdickmeiss (changes requested):

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss, line 502:
```suggestion
    closeReadClient().compose(x -> closeWriteClient()).onComplete(whenDone);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -376,10 +422,27 @@ void setClient(PgPool client) {
   }
 
   /**
-   * Close the SQL client of this PostgresClient instance.
+   * Close the SQL writer and reader clients of this PostgresClient instance in succession.
    * This is idempotent: additional close invocations are always successful.
    */
-  public Future<Void> closeClient() {
+  public Future<Void> closeReadClient() {
```
adamdickmeiss, line 428: closeReadClient is unused. We should not add useless code. So I suggest that to be removed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -376,10 +422,27 @@ void setClient(PgPool client) {
...
+      return Future.succeededFuture();
+    }
+    return clientToClose.close();
+  }
+
+  /**
+   * Close the SQL writer and reader clients of this PostgresClient instance in succession.
+   * This is idempotent: additional close invocations are always successful.
+   */
+  public Future<Void> closeWriteClient() {
```
adamdickmeiss, line 445: closeWriteClient is unsed. We should not add useless code. So I suggest that to be removed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Fixed now

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Removed

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
adamdickmeiss: Keeping the code as is as using compose fails.. Due to closeAllClients being used without Vert.x context in sync mode - at least in tests.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
mtraneis: Yes, it can be removed. I was trying various ways to get it to work and this may be a residue from doing it.

adamdickmeiss: OK. Will remove this and other one.

adamdickmeiss:
I tested this with mod-permissions MODPERMS-203 branch and did not see problems. Therefore removing draft state.

adamdickmeiss (approved):
OK, but biased now that I pushed some stuff.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: readClientToClose is null if no read client is configured. Therefore it should be swapped:

```
    closeClient(readClientToClose);
    return closeClient(clientToClose);
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -434,20 +492,40 @@ public static void closeAllClients() {
...
 
-  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig) {
+  static PgConnectOptions createPgConnectOptions(JsonObject sqlConfig, boolean isReader) {
     PgConnectOptions pgConnectOptions = new PgConnectOptions();
-    String host = sqlConfig.getString(HOST);
+    String hostToResolve = HOST;
+    String portToResolve = PORT;
+
+    if (isReader)
+    {
```
julianladisch, line 508: No line break before `{`

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -434,20 +492,40 @@ public static void closeAllClients() {
...
+       hostToResolve = HOST_READER;
+       portToResolve = PORT_READER ;
```
julianladisch, line 510: No space before `;`

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -434,20 +492,40 @@ public static void closeAllClients() {
...
-    Integer port = sqlConfig.getInteger(PORT);
+
+    Integer port;
+    port = sqlConfig.getInteger(portToResolve);
+
     if (port != null) {
       pgConnectOptions.setPort(port);
     }
+
+    if (isReader && (host == null || port == null)){
```
julianladisch, line 525: Space missing before `{`

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3354,9 +3482,28 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
    * @see #withTrans(Function)
    * @see #withTransaction(Function)
    */
+
   void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
     getSQLConnection(0, handler);
   }
+  /**
```
julianladisch, line 3489: Empty line missing

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3372,33 +3519,55 @@ void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
...
+  }
 
-      SQLConnection sqlConnection = new SQLConnection(pgConnection, null, timerId);
-      handler.handle(Future.succeededFuture(sqlConnection));
-    });
+  /**
+   * Get the SQL Write connection
+   *
+   * @see #withReadConn(Function)
+   * @see #withReadConnection(Function)
```
julianladisch, line 3553:
```
   * @see #withReadConn(Function)
   * @see #withReadConnection(Function)
```
should be replaced with
```
   * @see #withConn(Function)
   * @see #withConnection(Function)
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3372,33 +3519,55 @@ void getSQLConnection(Handler<AsyncResult<SQLConnection>> handler) {
...
+    getConnection(res -> getSQLConnection(res, queryTimeout, handler));
+  }
+
+  /**
+   * Get the SQL Read connection
+   *
+   * @see #withReadConn(Function)
+   * @see #withReadConnection(Function)
+   * @see #withTrans(Function)
+   * @see #withTransaction(Function)
```
julianladisch, line 3567:
```
   * @see #withTrans(Function)
   * @see #withTransaction(Function)
```
should be replaced with
```
   * @see #withReadTrans(Function)
   * @see #withReadTransaction(Function)
```

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: In line 2540 withConn should be called because a connection is passed as a parameter that cannot be changed from writer to reader.

This method `withReadConn(AsyncResult<SQLConnection> sqlConnection, Function<Conn, Future<T>> function)` is not needed and should be deleted.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3354,9 +3482,28 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
    * @see #withTrans(Function)
    * @see #withTransaction(Function)
    */
+
```
julianladisch, line 3485: Can you remove this empty line?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3354,9 +3482,28 @@ public void getConnection(Handler<AsyncResult<PgConnection>> replyHandler) {
...
+   * <p>Use closeAndHandleResult as replyHandler, for example:
+   *
+   * <pre>getSQLReadConnection(conn -> execute(conn, sql, params, closeAndHandleResult(conn, replyHandler)))</pre>
+   *
+   * <p>Or avoid this method and use the preferred {@link #withConn(Function)}.
+   *
+   * @see #withReadConn(Function)
+   * @see #withReadConnection(Function)
+   * @see #withTrans(Function)
+   * @see #withTransaction(Function)
```
julianladisch, line 3501: `withTrans` should be `withReadTrans`
`withTransaction` shoudl be `withReadTransaction`

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -2986,15 +3086,15 @@ public Future<RowSet<Row>> select(String sql) {
...
+   * Run a select query using the readonly connection.
    *
    * <p>To update see {@link #execute(String, Handler)}.
    *  @param sql - the sql query to run
    * @param queryTimeout query timeout in milliseconds, or 0 for no timeout
    * @param replyHandler the query result or the failure
    */
   public void select(String sql, int queryTimeout, Handler<AsyncResult<RowSet<Row>>> replyHandler) {
-    getSQLConnection(queryTimeout,
+    getSQLReadConnection(queryTimeout,
```
julianladisch, line 3097: sql can be a writing SQL: `UPDATE foo SET jsonb=$1 RETURNING jsonb`

Therefore this change cannot be made in this `select` method but requires a new `selectRead` method.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3097,15 +3197,15 @@ public void select(AsyncResult<SQLConnection> conn, String sql, Tuple params,
...
+   * Run a select query with the readonly connection and return the first record, or null if there is no result.
    *
    * <p>To update see {@link #execute(String, Handler)}.
    *
    * @param sql  The sql query to run.
    * @param replyHandler  The query result or the failure.
    */
   public void selectSingle(String sql, Handler<AsyncResult<Row>> replyHandler) {
-    getSQLConnection(conn -> selectSingle(conn, sql, closeAndHandleResult(conn, replyHandler)));
+    getSQLReadConnection(conn -> selectSingle(conn, sql, closeAndHandleResult(conn, replyHandler)));
```
julianladisch, line 3208: sql can be a writing SQL: `UPDATE foo SET jsonb='{}' where id='123' RETURNING jsonb`

Therefore this change cannot be made in this `selectSingle` method but requires a new `selectSingleRead` method.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3145,7 +3246,7 @@ public void selectSingle(AsyncResult<SQLConnection> conn, String sql, Handler<As
    * @param replyHandler  The query result or the failure.
    */
   public void selectSingle(String sql, Tuple params, Handler<AsyncResult<Row>> replyHandler) {
-    getSQLConnection(conn -> selectSingle(conn, sql, params, closeAndHandleResult(conn, replyHandler)));
+    getSQLReadConnection(conn -> selectSingle(conn, sql, params, closeAndHandleResult(conn, replyHandler)));
```
julianladisch, line 3249: sql can be a writing SQL: `UPDATE foo SET jsonb=$1 WHERE id='123' RETURNING jsonb`

Therefore this change cannot be made in this `selectSingle` method but requires a new `selectReadSingle` method.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -3221,7 +3322,7 @@ public void selectSingle(AsyncResult<SQLConnection> conn, String sql, Tuple para
    * @param chunkSize cursor fetch size
    */
   public Future<Void> selectStream(String sql, Tuple params, int chunkSize, Handler<RowStream<Row>> rowStreamHandler) {
-    return withTrans(trans -> trans.selectStream(sql, params, chunkSize, rowStreamHandler));
+    return withReadTrans(trans -> trans.selectStream(sql, params, chunkSize, rowStreamHandler));
```
julianladisch, line 3325: sql can be a writing SQL: `UPDATE foo SET jsonb=$1 RETURNING jsonb`

Therefore this change cannot be made in this `selectStream` method but requires a new `selectReadStream` method.

julianladisch, line 3325: The javadoc in Conn.java should be reviewed to show read methods where appropriate.

mtraneis: > The javadoc in Conn.java should be reviewed to show read methods where appropriate.
I don't think I understand this comment. No methods in Conn.java was updated.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
mtraneis: > Therefore this change cannot be made in this select method but requires a new selectRead method.

A new selectRead method to be created that has the same arguments and return type as this select method?

julianladisch: Yes.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
julianladisch: The javadoc in Conn.java should mention read-only transactions where appropriate.
```
   * Sample usage:
   *
   * <pre>
   * postgresClient.withTrans(conn -> step1()
   *     .compose(x -> conn.selectStream("SELECT i FROM numbers WHERE i > $1", Tuple.tuple(5),
   *         rowStream -> rowStream.handler(row -> task.process(row))))
   *     .compose(x -> ...
   * </pre>
   *
   * <p>Use withReadTrans if all SQL queries are read-only (no nextval(), no UPDATE, ...):
   *
   * <pre>
   * postgresClient.withReadTrans(conn ->
   *     conn.selectStream("SELECT i FROM numbers WHERE i > $1", Tuple.tuple(5),
   *         rowStream -> rowStream.handler(row -> task.process(row))));
   * </pre>
```

PR #1052, opened 2022-05-30
RMB-919: Delete by CQL rejects missing or empty CQL

julianladisch:
As a sysop or librarian using the DELETE by CQL API I might accidentally omit the query parameter or mistype it: quer=username==foo

How to reproduce:
Call PgUtil.delete by CQL with CQL query parameter that is null (= missing), empty or whitespace only.

Expected Results:
PgUtil.delete by CQL rejects the API call with a 400 HTTP status code and doesn't delete anything.

Actual Results:
PgUtil.delete deletes all records.

Additional Information:
This bug has already caused severe issues: https://issues.folio.org/browse/MODINVSTOR-901


adamdickmeiss (approved):
Makes sense to require query for deletion.

PR #1056, opened 2022-06-02
RMB-922: Case insensitive header name matching in TenantTool

julianladisch:
TenantTool.tenantId switched from x-okapi-tenant to X-Okapi-Tenant
for RMB 34.0.0, therefore we need to fix the matching.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantTool.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantTool.java
@@ -28,6 +28,13 @@ public final class TenantTool {
    * @return the tenantId for the headers, returns the default "folio_shared" if undefined
    */
   public static String tenantId(Map<String, String> headers) {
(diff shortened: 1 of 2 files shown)


julianladisch:
mod-inventory-storage has several unit tests that pass a Map.of() and fail when upgrading to RMB 34:
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/AsyncMigrationTest.java#L142
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/IterationJobRunnerTest.java#L134
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/NotificationSendingErrorRepositoryTest.java#L35
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/ReindexJobRunnerTest.java#L185

Do you prefer no code change and documentation and a migration note only? #1058

julianladisch:
Closing because the alternative solution #1058 is preferred.

PR #1058, opened 2022-06-02
RMB-922: Migrate to case insensitive x-okapi-tenant header

julianladisch:
mod-inventory-storage has several unit tests that pass a Map.of() and fail when upgrading to RMB 34:
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/AsyncMigrationTest.java#L142
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/IterationJobRunnerTest.java#L134
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/NotificationSendingErrorRepositoryTest.java#L35
https://github.com/folio-org/mod-inventory-storage/blob/v23.0.3/src/test/java/org/folio/rest/api/ReindexJobRunnerTest.java#L185

### DIFF
--- a/doc/upgrading.md
+++ b/doc/upgrading.md
@@ -93,6 +93,12 @@ New:
 
 This may cause some unit tests to fail, use java.time.Instant and compare Instants to be format agnostic.
 
+#### [RMB-944](https://issues.folio.org/browse/RMB-944)
+
+Unit tests that pass X-Okapi-Tenant header into TenantTool.tenantId must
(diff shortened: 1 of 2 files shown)


adamdickmeiss (approved):
I slightly prefer this over the other approach.

PR #1060, opened 2022-06-15
RMB-926: Minimum PostgreSQL version: make configurable, fix int comp

julianladisch:


README.md
```diff
@@ -855,7 +857,25 @@ create table <schema>.<table_name> (
...
+
+```
+public class TenantRefAPI extends TenantAPI {
+  @Validate
+  @Override
+  public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers,
+                         Handler<AsyncResult<Response>> handler, Context context)  {
+    // https://www.postgresql.org/docs/current/runtime-config-preset.html#GUC-SERVER-VERSION-NUM
+    context.putLocal("postgres_min_version_num", "100000");
+    context.putLocal("postgres_min_version", "10.0");  // human readable, for error message only
```
adamdickmeiss, line 872: Having only one setting - the string version - is more elegant and then calculate the numeric version from it.

Not a show-stopper, but would doing that calculation even be more code lines?

julianladisch: The latest 9.6 version is 9.6.24 with server_version_num = 90624

adamdickmeiss (approved):

PR #1064, opened 2022-06-17
http: Do not join response headers RMB-927

adamdickmeiss:


steveellis:
@adamdickmeiss - I pulled the snapshot build into mod-login and was able to see it work. If I do:
```
var response = Response.status(201)
  .header("Set-Cookie", "at=abc123; HttpOnly; Max-Age=123")
  .header("Set-Cookie", "rt=xyz321")
  .build();
```
I am able to test it with:
```
.statusCode(201)
  .cookie("at", RestAssuredMatchers.detailedCookie()
  .value("abc123")
  .maxAge(123)
  .httpOnly(true))
  .cookie("rt", "xyz321")
```
This doesn't account for the code that RMB generates (if it is RMB that is doing that). I guess the only downside of this approach is that we can't define the `Set-Cookie` as a header in the RAML. But it doesn't break the cookie code that gets generated should someone still want to use that feature of RAML. The code that gets generated when you have a `Set-Cookie` header has a `withSetCookie` setter, which still has a `Map` behind it since it looks like your change is on the implementation of `Response`.

I tried to find the place in RMB where this code generation happens but was unsuccessful.

Anyway working with the response object directly makes multiple `Set-Cookie` work for me and I think gets us where we want to go. Is there anything blocking merging this?

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -344,6 +345,38 @@ public void invalidCqlClosesConnection(TestContext context) {
...
+    d.setDatetime(new Datetime());
+    d.setLink("link");
+
+    Book b = new Book();
+    b.setData(d);
+    b.setStatus(0);
+    b.setSuccess(true);
+
+    ObjectMapper om = new ObjectMapper();
+    String book = om.writerWithDefaultPrettyPrinter().writeValueAsString(b);
```
julianladisch, line 370: The duplicate code here, in  getBookWithRoutingContext(TestContext) and test(TestContext) should be extracted into a helper method, for example `private static String book()`.

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
```diff
@@ -344,6 +345,38 @@ public void invalidCqlClosesConnection(TestContext context) {
...
+    given()
+        .spec(tenant).header("Content-Type", "application/json")
+        .body(book).post("/rmbtests/books")
+        .then()
+        .statusCode(201)
+        .cookie("a", RestAssuredMatchers.detailedCookie().value("1"))
+        .cookie("b", RestAssuredMatchers.detailedCookie().value("2"));
```
steveellis, line 377: Think you can just do `.cookie("a", "1")`. `detailedCooke` is for when you have other properties like `HttpOnly` etc.

The format for a detailed cookie is `a=1; HttpOnly; Max-Age=123`. I don't think we have to test for all the possible values. But just to be on the safe side, you could make one of the cookies be detailed and then test for them with the `detailedCooke`.

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
adamdickmeiss: ok. done

domain-models-runtime/src/test/java/org/folio/DemoRamlRestTest.java
adamdickmeiss: added more beefy cookie example

julianladisch (approved):
Looks good, thanks!

PR #1065, opened 2022-06-27
Update NEWS, set version to 34.2.0-SNAPSHOT

julianladisch:


cql2pgjson-cli/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.folio</groupId>
     <artifactId>raml-module-builder</artifactId>
-    <version>34.1.0-SNAPSHOT</version>
+    <version>34.2.0-SNAPSHOT</version>
```
adamdickmeiss, line 9: Why not RMB 35.0.0-SNAPSHOT? You argued that RMB-927 is a breaking change (I don't see how, but still).

adamdickmeiss (approved):

PR #1067, opened 2022-07-08
Update a SelectStream function to call withReadConn()

mtraneis:
Update a selectStream function to use withReadConn().  The selectStream() is called by mod-inventory-storage inventoryHierarchyItemsAndHoldings by RTAC.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
@@ -3386,7 +3386,7 @@ public class PostgresClient {
   void selectStream(AsyncResult<SQLConnection> sqlConnection, String sql, Tuple params, int chunkSize,
       Handler<AsyncResult<RowStream<Row>>> replyHandler) {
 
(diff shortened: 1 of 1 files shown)


julianladisch (changes requested):
sql might be a writing SQL: `UPDATE foo SET json=$1 RETURNING json`

A method withReadConn that takes an sqlConnection doesn't make sense, see code review of https://github.com/folio-org/raml-module-builder/pull/1047#discussion_r876010317

julianladisch:
https://github.com/folio-org/mod-inventory-storage/blob/v24.0.0/src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java#L97 should use `withReadTrans(conn -> conn.selectStream(`

mtraneis:
> https://github.com/folio-org/mod-inventory-storage/blob/v24.0.0/src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java#L97 should use `withReadTrans(conn -> conn.selectStream(`

Thanks, @julianladisch . Looks like this change isn't needed then if there is already a function that mod-inventory-storage can use.

PR #1070, opened 2022-07-13
Updates for tests OpenJDK 17

adamdickmeiss:


domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1405,15 +1404,7 @@ private UserdataCollection searchForDataUnoptimizedNoClass(String cql, int offse
...
         vertx.getOrCreateContext(), ResponseImpl.class, testContext.asyncAssertSuccess(response -> {
-          if (response.getStatus() != 500) {
-            testContext.fail("Expected status 500, got "
-                + response.getStatus() + " " + response.getStatusInfo().getReasonPhrase());
-            async.complete();
-            return;
-          }
-          UserdataCollection c = (UserdataCollection) response.getEntity();
-          userdataCollection.setTotalRecords(c.getTotalRecords());
-          userdataCollection.setUsers(c.getUsers());
```
julianladisch, line 1416: If userdataCollection is not filled with the results the method should return void.

domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java
```diff
@@ -1425,15 +1416,6 @@ private UserdataCollection searchForDataUnoptimizedNo500(String cql, int offset,
...
         vertx.getOrCreateContext(), ResponseWithout500.class, testContext.asyncAssertSuccess(response -> {
-          if (response.getStatus() != 400) {
-            testContext.fail("Expected status 400, got "
-                + response.getStatus() + " " + response.getStatusInfo().getReasonPhrase());
-            async.complete();
-            return;
-          }
-          UserdataCollection c = (UserdataCollection) response.getEntity();
-          userdataCollection.setTotalRecords(c.getTotalRecords());
-          userdataCollection.setUsers(c.getUsers());
```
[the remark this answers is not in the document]
adamdickmeiss: right

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/OutStreamTest.java
```diff
@@ -9,12 +9,24 @@
...
+
+  @BeforeClass
+  public static void beforeClass() {
+    String version = System.getProperty("java.version");
+    int dot = version.indexOf('.');
+    major = Integer.parseInt(version.substring(0, dot));
+  }
+
   private void setDefaultCharset(Charset charset) {
+    Assume.assumeTrue(major < 17);
```
julianladisch, line 29: Disabling tests should be avoided.
Instead of testing that the default charset is not used we can test that a given charset is used.

```
  private void testCharset(Charset charset) {
    try {
      OutStream outStream = new OutStream(charset);
      outStream.setData("ä");
      ByteArrayOutputStream result = new ByteArrayOutputStream();
      outStream.write(result);
      assertArrayEquals("ä".getBytes(charset), result.toByteArray());
    } catch (IOException e) {
      throw new UncheckedIOException(e);
    }
  }

  @Test
  public void utf8() {
    testCharset(StandardCharsets.UTF_8);
  }

  @Test
  public void usAscii() {
    testCharset(StandardCharsets.US_ASCII);
  }

  @Test
  public void iso_8859_1() {
    testCharset(StandardCharsets.ISO_8859_1);
  }
```

```
public class OutStream implements javax.ws.rs.core.StreamingOutput {

  private final Charset charset;
  private Object data;

  /**
   * Default constructor for using UTF8 when encoding data Object.
   */
  public OutStream() {
    this(StandardCharsets.UTF_8);
  }

  /**
   * Constructor with a user defined charset for encoding data Object.
   */
  public OutStream(Charset charset) {
    this.charset = charset;
  }

  /**
   * Constructor for using UTF8 when encoding data Object and initializing
   * the data Object with a JSON representation of the RowSet result parameter.
   */
  public OutStream(RowSet<Row> result) {
    this(StandardCharsets.UTF_8);
    JsonArray ar = new JsonArray();
    RowIterator<Row> it = result.iterator();
    while (it.hasNext()) {
      Row row = it.next();
      JsonObject o = new JsonObject();
      for (int i = 0; i < row.size(); i++) {
        o.put(row.getColumnName(i), row.getValue(i));
      }
      ar.add(o);
    }
    data = ar.encode();
  }

  @Override
  public void write(OutputStream output) throws IOException {
    output.write(data.toString().getBytes(charset));
    output.flush();
  }
```

adamdickmeiss: I still don't understand the whole point of this test. The point of the default charset was to see that it was just not in use.

So I'd prefer to just remove the non-working JDK 17 tests.

adamdickmeiss: While your code looks nice it doesn't really prove that the non-default is used and UTF-8. One would have to read the constructor code in OutStream to prove. And now OutStream has more useless code (like setting the charset, which is exactly what we want to avoid). .We want to ensure UTF-8 is always used.

I don't think there's a way to do this without mocking. I've removed the tests that are IMHO useless and we read the code and see that it does things right.

julianladisch: The OutStream(Charset charset) constructor is used to mock. By providing a different Charset we see whether the provided charset is used (good) or Java's default charset (bad).

julianladisch (approved):

domain-models-runtime/src/test/java/org/folio/rest/tools/utils/OutStreamTest.java
julianladisch:
```
  @Test
  public void utf8() {
    testCharset(StandardCharsets.UTF_8);
  }
```
should use
```
  @Test
  public void utf8() {
    testCharset();
  }
```

julianladisch: To make use of the non-argument constructor:
```
  private void testCharset(OutStream outStream, Charset charset) {
```
```
  @Test
  public void utf8() {
    testCharset(new OutStream(), StandardCharsets.UTF_8);
  }
```

PR #1073, opened 2022-08-08
RMB-933 exclude self reference elements from URI encoding

grabsefx:
https://issues.folio.org/browse/RMB-933
*  exclude self reference elements from URI encoding


SerhiiNosko (approved):

domain-models-maven-plugin/src/main/java/org/folio/rest/tools/utils/SchemaDereferencer.java
```diff
@@ -106,4 +106,8 @@ static String toFileUri(Path basePathFile, String file) {
     Path absolutePath = basePathFile.resolveSibling(file).normalize();
     return absolutePath.toUri().toString();
   }
+
+  private static boolean hasSelfReference(String refValue) {
+    return refValue.contains("#");
+  }
```
adamdickmeiss, line 112: Either check for first character to make this conform to the method name ; or rename the method to say hasUriFragment in which the method does what it says. In both cases, things should work. The hasUriScheme would also return true if fragment in a URL.

adamdickmeiss (changes requested):

domain-models-maven-plugin/src/main/java/org/folio/rest/tools/utils/SchemaDereferencer.java
julianladisch: I prefer the former.

julianladisch (approved):
Looks good, thanks!

adamdickmeiss (approved):

grabsefx:
@adamdickmeiss could you merge this PR? It seems I dont have permissisions to do that

adamdickmeiss:
Thx for the contribution

PR #1075, opened 2022-08-11
RMB-934: Vert.x 4.3.3 fixing disabled SSL in 4.3.0/4.3.1

julianladisch:
Vert.x 4.3.0 and 4.3.1 have this regression: They disable SSL in WebClient.

Vert.x >= 4.3.2 has a fix.

https://github.com/vert-x3/vertx-web/issues/2196 "SSL not set for secure Request"

https://github.com/vert-x3/vertx-web/pull/2226 "WebClient ignores the SSL flag when a request is created from request options with the SSL flag true"

https://github.com/vert-x3/vertx-web/issues/2225 "HttpRequestImpl's SSL Isn't Set By RequestOptions"

### DIFF
--- a/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTransactionsIT.java
+++ b/domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTransactionsIT.java
@@ -422,7 +422,7 @@ public class PostgresClientTransactionsIT extends PostgresClientITBase {
     }
   }
 
-  class MonitorPgConnection implements PgConnection {
+  class MonitorPgConnection extends PgConnectionMock {
     final PgConnection conn;
     final AtomicInteger open;
     final AtomicInteger active;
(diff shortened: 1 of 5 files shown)


domain-models-runtime/src/test/java/org/folio/rest/persist/PostgresClientTransactionsIT.java
```diff
@@ -521,11 +516,6 @@ public PreparedQuery<RowSet<Row>> preparedQuery(String sql, PrepareOptions optio
       return conn.preparedQuery(sql, options);
     }
 
-    @Override
-    public void close(Handler<AsyncResult<Void>> handler) {
-      close().onComplete(handler);
-    }
-
```
adamdickmeiss, line 528: Really annoying that Vert.x changes interfaces in what appears as bug fix release. Not following semantic versioning policy at all.

adamdickmeiss (approved):
Thx

PR #1082, opened 2022-09-17
Fix previous version used for migration scripts RMB-937

adamdickmeiss:
Use the stored moduleVersion as previous version rather than module_from from tenant init call. The module_from does not always reflect the true state of module data.


julianladisch (approved):
Looks good, thanks!

PR #1092, opened 2022-10-15
RMB-946: Retry GET request in TenantLoading (reference/sample data)

julianladisch:
Trying to build the platform complete snapshot reference environment repeatedly failed with error message from https://github.com/folio-org/raml-module-builder/blob/v35.0.1/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java#L326-L341 :

```
GET http://10.36.1.44:9134/identifier-types/1795ea23-6856-48a5-a772-f356e16a8a6c returned status 500: Timeout
POST http://10.36.1.44:9134/identifier-types returned status 400: duplicate key value violates unique constraint
```

Retrying the GET request after some sleep time might resolve the issue.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java
@@ -242,7 +242,7 @@ public class TenantLoading {
     }
   }
 
-  private static Future<Integer> loadURL(Map<String, String> headers, URL url,
+  private Future<Integer> loadURL(Map<String, String> headers, URL url,
                                       WebClient httpClient, LoadingEntry loadingEntry, String endPointUrl) {
 
     return getContent(url, loadingEntry)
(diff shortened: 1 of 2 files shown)


adamdickmeiss (changes requested):
I don't think we should resort to random waits without understanding why timeout errors happen.

julianladisch:
While it's true that we should also find the underlying cause of the timeout I don't see a reason why we shouldn't implement this well known retry pattern:
https://learn.microsoft.com/en-us/azure/architecture/patterns/retry
https://dev.to/azure/learn-about-the-retry-pattern-in-5-minutes-fjo
https://www.abhinavpandey.dev/blog/retry-pattern
https://blog.codecentric.de/resilience-design-patterns-retry-fallback-timeout-circuit-breaker
https://ieeexplore.ieee.org/abstract/document/9101301

jakub-id:
I don't think this should be merged.

PR #1093, opened 2022-10-17
RMB-947: LocalRowSet for RowDesc from both 4.3.3 and 4.3.4

julianladisch:
Vert.x 4.3.4 has a breaking change in the constructor of RowDesc.

Add reflection to LocalRowSet so that is can use RowDesc from both Vert.x 4.3.3 and 4.3.4.

This allows to downgrade Vert.x, 4.3.1 less frequently (32%) fails the mod-inventory-storage unit tests than 4.3.4 (94%).


jakub-id:
@adamdickmeiss  @julianladisch  Can we reject and close this?

julianladisch:
No, why?

adamdickmeiss (approved):
LGTM. Difficult to unit test.

PR #1098, opened 2022-11-08
RMB-952: Deprecate RowDesc in LocalRowDesc

julianladisch:
org.folio.rest.persist.helpers.LocalRowDesc extends io.vertx.sqlclient.impl.RowDesc.

Classes in the impl directories are Vert.x internals and must not be used.

Using RowDesc broke RMB because Vert.x made a breaking change: https://issues.folio.org/browse/RMB-945
https://issues.folio.org/browse/RMB-947

Therefore deprecate RowDesc usage so that it can be removed in RMB 36.0.0.


adamdickmeiss (approved):
Remember also upgrading notes

adamdickmeiss (approved):

PR #1102, opened 2022-11-21
RMB-944: Migration "ERROR: tuple concurrently updated" on DROP TRIGGER

julianladisch:
Migrating mod-inventory-storage from Lotus to Kiwi (mod-inventory-storage 24.1.0) fails with this error:

```
Tenant operation failed for module mod-inventory-storage-24.1.0:
SQL error
DROP TRIGGER IF EXISTS set_id_in_jsonb ON tenantid_mod_inventory_storage.item CASCADE;
ERROR: tuple concurrently updated (XX000)
```

The same error is sometimes thrown when executing "DROP table if exists".

To mitigate this PR adds pg_advisory_xact_lock and retries.


adamdickmeiss (approved):
Postgresql is a mess.

PR #1103, opened 2022-11-23
Release 35.0.4

julianladisch:


adamdickmeiss (commented):
If there's a security fix in Vert.x 4.3.5, perhaps include that too (so you save a release). EDGCOMMON-56

adamdickmeiss (approved):

PR #1105, opened 2022-11-30
RMB-957: "duplicate key" error on parallel CREATE EXTENSION

julianladisch:
How to reproduce: Enable two tenants for mod-source-record-manager in parallel.

One succeeds, the other fails:

```
CREATE EXTENSION IF NOT EXISTS unaccent WITH SCHEMA public
ERROR: duplicate key value violates unique constraint "pg_extension_name_index" (23505)
```

Solution:

Wrap CREATE EXTENSION into a pg_advisory_xact_lock.


adamdickmeiss (commented):
Does this have to part of public schema? Why not the module+tenant schema?

adamdickmeiss (approved):
Too bad with tenants not being independent on the DB side.

PR #1108, opened 2022-12-20
UXPROD-3833. Logging adjustment (NFR) - Orchid work

SerhiiNosko:
In folio-spring-base we use param appender.console.layout.includeTimeMillis as true [Link](https://github.com/folio-org/folio-spring-base/blob/master/src/main/resources/log4j2-json.properties).
This PR has been created to make logging configuration consistent between vertx and spring modules.
Here is article that describes desired logging configuration: https://wiki.folio.org/display/DD/Folio+logging+solution


adamdickmeiss:
Please file issue on RMB for this.

PR #1109, opened 2022-12-20
Fix NPE in PomReader url test

julianladisch:
Fixes this NPE that I cannot reproduce locally but happens on Jenkins:
```
java.lang.NullPointerException
	at org.folio.rest.tools.PomReader.getModelFromJar(PomReader.java:101)
	at org.folio.rest.tools.PomReader.readIt(PomReader.java:77)
	at org.folio.rest.tools.PomReaderTest.lambda$readFromJarNoResource$1(PomReaderTest.java:67)
```
https://jenkins-aws.indexdata.com/job/folio-org/job/raml-module-builder/job/PomReaderTest-NPE/7/execution/node/53/log/


adamdickmeiss (approved):
Maybe on Jenkins it's using Java 11 and you are running Java 17?

PR #1114, opened 2023-02-13
RMB-962: Report missing snippetPath file in schema.json

julianladisch:
schema.json can have snippetPath entries that point to an SQL script file.

If the file doesn't exist the installation or upgrade migration ignores it and succeeds without warning or error.

Fix: Fail the installation or upgrade on missing snippetPath file.


adamdickmeiss (approved):
This is a breaking change

PR #1118, opened 2023-03-26
RMB-965: runSqlFile with failed Future on SQL failure

julianladisch:
The existing PostgresClient#runSQLFile methods return a succeeded Future when an SQL statement fails. The database failure messages are returned as a List of String.

This doesn't comply with the Future contract.

Add a PostgresClient#runSqlFile(String sqlFile) method that returns a failed Future when an SQL statement fails.

This simplifies code including RMB production and test code.

### DIFF
--- a/.github/workflows/api-doc.yml
+++ b/.github/workflows/api-doc.yml
@@ -1,0 +1,91 @@
+name: api-doc
+
+# https://dev.folio.org/guides/api-doc/
+
+# API_TYPES: string: The space-separated list of types to consider.
+# One or more of 'RAML OAS'.
+# e.g. 'OAS'
+#
+# API_DIRECTORIES: string: The space-separated list of directories to search
+# for API description files.
+# e.g. 'src/main/resources/openapi'
+# NOTE: -- Also add each separate path to each of the "on: paths:" sections.
+# e.g. 'src/main/resources/openapi/**'
+#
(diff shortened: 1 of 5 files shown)


domain-models-runtime/src/main/java/org/folio/rest/impl/AdminAPI.java
```diff
@@ -208,25 +207,19 @@ public void postAdminImportSQL(InputStream entity, Map<String, String> okapiHead
...
+      })
+      .onSuccess(x -> asyncResultHandler.handle(Future.succeededFuture(
+          PostAdminImportSQLResponse.respond200(PostAdminImportSQLResponse.headersFor200())))
+      );
     } catch (Exception e) {
-      log.error(e.getMessage(), e);
-      asyncResultHandler.handle(io.vertx.core.Future.failedFuture(e.getMessage()));
+      log.error("{}", e.getMessage(), e);
+      asyncResultHandler.handle(Future.succeededFuture(
+          PostAdminImportSQLResponse.respond500WithTextPlain(e.getMessage())));
```
steveellis, line 222: I agree what you have here uses futures the right way. Also, it looks like this isn't a breaking change to modules that rely on this method. Do I have that right?

steveellis (approved):

PR #1122, opened 2023-04-20
Suppress false positive ResourceInfo equals Sonar warning

julianladisch:


domain-models-api-aspects/src/main/java/org/folio/rest/tools/utils/ClassPath.java
```diff
@@ -178,6 +178,8 @@ public int hashCode() {
       return resourceName.hashCode();
     }
 
+    @SuppressWarnings("java:S2162")  // false positive:
+    // Compare to "this.getClass()" instead. "equals" methods should be symmetric and work for subclasses
```
steveellis, line 182: This comment doesn't make it clear to me exactly why we want to suppress this warning in this override in this class. The sonar warning is valid in general, but I'm assuming that in this specific case it is not valid. Can the comment say why?

steveellis (approved):

PR #1124, opened 2023-05-31
RMB-970: Add PgExceptionFacade.isVersionConflict()

julianladisch:
Add PgExceptionFacade.isVersionConflict(). It was forgotten when we've added PgExceptionUtil.isVersionConflict(Throwable).

https://github.com/folio-org/raml-module-builder/blob/v35.0.6/domain-models-runtime/src/main/java/org/folio/rest/persist/PgExceptionUtil.java#L73-L83

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 
 buildMvn {
   mvnDeploy = 'yes'
-  buildNode =  'jenkins-agent-java11'
+  buildNode =  'jenkins-agent-java17'
 }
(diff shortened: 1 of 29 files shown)


domain-models-runtime/src/test/java/org/folio/rest/persist/PgExceptionFacadeTest.java
```diff
@@ -24,10 +24,16 @@ void getter() {
...
     PgExceptionFacade f = new PgExceptionFacade(new PgException(null, null, "22P02", null));
     assertThat(f.isInvalidTextRepresentation(), is(true));
-    PgExceptionFacade f2 = new PgExceptionFacade(new PgException(null, null, "22P03", null));
-    assertThat(f2.isInvalidTextRepresentation(), is(false));
+    assertThat(f.isVersionConflict(), is(false));
+  }
+
+  @Test
+  void versionConflict() {
+    PgExceptionFacade f = new PgExceptionFacade(new PgException(null, null, "23F09", null));
```
steveellis, line 35: Might make it easier to follow if you used the constant `PgExceptionUtil.VERSION_CONFLICT` to set the 3rd parma here and elsewhere just to show that these strings are not just random but have meaning.

steveellis (approved):

PR #1130, opened 2023-09-20
RMB-978: TESTCONTAINERS_POSTGRES

julianladisch:
TC requests an evaluation how bumping the PostgreSQL version from 12 to 13, 14 or 15 affects the modules: https://wiki.folio.org/display/TC/2023-09-18+Meeting+notes

Modules use a hardcoded postgres version number when starting a postgres container with testcontainers for their unit tests (mvn test) and integration tests (mvn verify).

Allow to change the postgres container image with the environment variable TESTCONTAINERS_POSTGRES.

This is a build time configuration, it doesn't affect the runtime.

### DIFF
--- a/postgres-testing/src/test/java/org/folio/postgres/testing/PostgresTesterContainerTest.java
+++ b/postgres-testing/src/test/java/org/folio/postgres/testing/PostgresTesterContainerTest.java
@@ -25,6 +25,11 @@ public class PostgresTesterContainerTest {
 
   @Test
   public void imageName() {
+    assertEquals(PostgresTesterContainer.DEFAULT_IMAGE_NAME, PostgresTesterContainer.getImageName());
+  }
+
+  @Test
+  public void imageNameEnvEmpty() {
     assertEquals(PostgresTesterContainer.DEFAULT_IMAGE_NAME, PostgresTesterContainer.getImageName(Map.of()));
   }
 


postgres-testing/src/main/java/org/folio/postgres/testing/PostgresTesterContainer.java
```diff
@@ -36,6 +38,8 @@ public class PostgresTesterContainer implements PostgresTester {
    */
   public static final int SIMULATED_ASYNC_REPLICATION_LAG_MILLISECONDS = 300;
 
+  private static final String IMAGE_NAME = getImageName(System.getenv());
```
steveellis, line 41: Would it maybe be more convenient to pass this as a system parameter rather than as an environment variable, or perhaps support both? If it is a system parameter then it is more easily passed to `-D` in the same command as `mvn test`.

Or is it that this will only be used in a "multiple module" integration style test which perhaps would argue for just using an env var.

steveellis (approved):

postgres-testing/src/main/java/org/folio/postgres/testing/PostgresTesterContainer.java
steveellis: Seems fine thanks.

PR #1131, opened 2023-09-23
RMB-979: Poppy dependencies: Vert.x 4.4.5, log4j 2.20.0, ...

julianladisch:
For the RMB Poppy release upgrade all dependencies to current stable versions.

Add `<project.build.outputTimestamp>` towards reproducible-builds:
https://maven.apache.org/guides/mini/guide-reproducible-builds.html

### DIFF
--- a/domain-models-runtime/src/test/java/org/folio/rest/persist/PgConnectionMock.java
+++ b/domain-models-runtime/src/test/java/org/folio/rest/persist/PgConnectionMock.java
@@ -35,6 +35,11 @@ public class PgConnectionMock implements PgConnection {
   }
 
   @Override
+  public Future<Void> cancelRequest() {
+    return Future.succeededFuture();
+  }
+
+  @Override
   public int processId() {
     return 0;
   }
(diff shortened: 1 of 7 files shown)


domain-models-runtime/src/test/java/org/folio/rest/persist/PgConnectionMock.java
```diff
@@ -91,6 +96,11 @@ public Future<Transaction> begin() {
     throw new PgConnectionMockException();
   }
 
+  @Override
+  public Transaction transaction() {
+    return null;
+  }
+
```
steveellis, line 103: It looks like this PR is mostly about upgrading dependencies. Yet there are these two new method overrides. Is it that one of the dependencies that got upgraded now require these overrides? Seems fine. Just thought I'd ask.

steveellis: I'm not really sure what problem that would solve. I suppose you _could_ do it if you feel it would be an improvement. I see mockito is used elsewhere in RMB.

steveellis (approved):

domain-models-runtime/src/test/java/org/folio/rest/persist/PgConnectionMock.java
steveellis: Maybe do a separate PR if you feel like making the change but I think what you have takes care of upgrading the deps.

PR #1142, opened 2024-01-02
RMB-985: Avoid SET ROLE/SET SCHEMA in Shared Pool

steveellis:
https://folio-org.atlassian.net/browse/RMB-985

Avoid `SET ROLE`/`SET SCHEMA` when a connection is reused for the same tenant.


domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public void clearCache() {
+    currentTenant = "";
+    currentIndex = 0;
+    connectionCache.clear();
+
+    LOG.debug("Cleared connection manager");
+  }
+
+  public Future<PgConnection> getConnection(Pool pool, String schemaName, String tenantId) {
```
okolawole-ebsco, line 35: can we change the datatype of the pool parameter from `Pool` to `PgPool`?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -164,6 +163,8 @@ public class PostgresClient {
   /** analyze threshold value in milliseconds */
   private static long explainQueryThreshold = EXPLAIN_QUERY_THRESHOLD_DEFAULT;
 
+  private static final PostgresConnectionManager POSTGRES_CONNECTION_MANAGER = new PostgresConnectionManager();
```
okolawole-ebsco, line 166: I would expect an instance of `PostgresConnectionManager` to have an instance per `PostgresClient`. With the current implementation, I believe there is going to be a churn on connections cache in the `PostgresConnectionManager` when running multi-tenant production load.

A `PostgresClient` can only be assigned to one tenant so it will not reused for other tenants. So i don't see a scenario where one `PostgresClient` is used by multiple tenants causes the `PostgresConnetionManager` to switch tenants.

okolawole-ebsco: It might be easier to have a `PostgresClient` be able to retrieve for a pool specific to a tenant. Such pool can be aggressive to have its min connection count to be zero. Some logic could ensure that we have tenant pools only for tenant ids needed.

steveellis: One tenant per pool with variable sized pools, could help with the contention, and also would allow for potentially longer sessions. The only tricky bit there is coming up with the max connections and the fractional  connections per tenant values. I think there’s still would need to be some coordination across clients to determine when to nuke the pool for one tenant in favor of another.

Something in between your idea and what I currently have could be still a static shared data structure but with connections in indexed by tenant. This could be implemented as an LRU cache. This would allow for connections to be longer lived for a given tenant and could reduce contention because of that. The only catch with this is making sure that when we get a new connection from the pool, it’s not already being used in the external cache for another tenant. Seems like what we really need is our Is our own pool implementation or a wrapper that stores the tenant info along with the connection.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private PgConnection getCachedConnection() {
+    if (connectionCache.isEmpty()) {
+      throw new IllegalStateException("No connections available");
+    }
+
+    LOG.debug("Returning cache item for tenant {} {}", currentTenant, currentIndex);
+
+    PgConnection nextConnection = connectionCache.get(currentIndex);
```
hjiebsco, line 91: There is no implementation to guarantee the connection is in usable state. It could be in use or even closed.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,96 @@
...
+            connectionCache.add((PgConnection) sqlConnection);
+            return (PgConnection) sqlConnection;
+          });
+    });
+  }
+
+  private PgConnection getCachedConnection() {
+    if (connectionCache.isEmpty()) {
+      throw new IllegalStateException("No connections available");
+    }
```
hjiebsco, line 87: Does this ever happen?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+    LOG.debug("Returning cache item for tenant {} {}", currentTenant, currentIndex);
+
+    PgConnection nextConnection = connectionCache.get(currentIndex);
+    currentIndex = (currentIndex + 1) % connectionCache.size();
```
hjiebsco, line 92: Any concurrency concern here?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,96 @@
...
+    if (!PostgresClient.sharedPgPool) {
+      LOG.debug("Not in shared pool mode");
+      return pool.getConnection().map(PgConnection.class::cast);
+    }
+
+    return getCachedConnection(pool, schemaName, tenantId);
+  }
+
+  private Future<PgConnection> getCachedConnection(Pool pool, String schemaName, String tenantId) {
+    if (tenantId.equals(currentTenant)) {
```
hjiebsco, line 45: Not sure how this will play out when multiple tenants use the system actively and concurrently.

hjiebsco (commented):
Managing connections created by `PgPool` directly might not be trivial. Alternatively, probably should consider tweaking the existing non-shared mode (one `PgPool` per tenant) to make it more efficient using connections.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
okolawole-ebsco: Another alternative could be,  having some data structure/service component that maintains counts of connections by tenant would help with setting max connections and fractional connections. This service component could act as a gate when retrieving a connection. if the max count for a tenant is already, it could error or wait. I am weary about our new logic maintaining a reference to connections itself. Overall, I think you understand my concern in my initial comment.

Also trying to not implement our own pool implementation if we can help it :)

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
steveellis: I considered that but am making the assumption that since its vertx (single thread event loop) and static we're not having to worry. The shared pool is static in the same way and shared between all clients if I'm remembering correctly. All callers should have the same view of the current tenant id and the cache. At least that's what I want to be true in order for this design to work at all.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
steveellis: The key to the idea working is that all callers have the same view of the cache and the current tenant id. If the tenant id happens to change, the cache is destroyed and we start building a new one. This of course it will perhaps multiple times a second, but sometimes not for longer periods. The hope is that in these longer periods, when requests flow in from a single tenant, we can reduce round trips during these batches.

Maybe the implementation needs work, but that is the design I'm going for. Everything static and shared with unified view of the current tenant and the cache that causes no conflicts and is fast.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
steveellis: I think this is THE question. I'm thinking the connection being actually closed isn't an issue, since the pool is open connections. Closing returns the conn to the pool, but leaves the underlying conn open. But what _could_ be an issue is that another caller grabs a connection that is in the cache and has its session pegged to whatever was set which is no longer right. But I don't think this will happen (or at least I want a mechanism to prevent it).

I think it really hinges on the lifecycle of the connections. Let's say I start two sets of SQL commands, each set that flows over their own connection, and each set having different tenant ids set. One finishes before the other. What is stopping me from giving out the connection that hasn't finished to another tenant when it is still not finished?

domain-models-runtime/src/main/resources/log4j2.properties
```diff
@@ -5,7 +5,7 @@ packages = org.folio.okapi.common.logging
 filters = threshold
 
 filter.threshold.type = ThresholdFilter
-filter.threshold.level = info
+filter.threshold.level = debug
```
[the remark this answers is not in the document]
steveellis: Just for testing. Will be set back to info before merging.

domain-models-runtime/src/test/java/org/folio/PostgresConnectionManagerTest.java
```diff
@@ -0,0 +1,33 @@
...
+        availableConn.setAvailable();
+        manager.tryClose(availableConn);
+      } else {
+        manager.tryClose(new CachedPgConnection("tenant2", null, null));
+      }
+    });
+    assertEquals(PostgresClient.DEFAULT_MAX_POOL_SIZE - 1, manager.getCacheSize());
+  }
+}
+
```
steveellis, line 33: More tests are coming!

domain-models-runtime/src/main/java/org/folio/rest/persist/CachedPgConnection.java
```diff
@@ -0,0 +1,194 @@
...
+    this.manager.tryClose(this);
+  }
+
+  @Override
+  public PgConnection closeHandler(Handler<Void> handler) {
+    LOG.debug("Calling close: closeHandler Handler<Void>");
+    this.available = true;
+    this.manager.tryClose(this);
+    return this;
+  }
```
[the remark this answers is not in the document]
steveellis: Not sure I'm understanding this one. Isn't what I have equivalent to this: https://github.com/eclipse-vertx/vertx-sql-client/blob/5cc06bf5251e96b9005d703b66504fc0260d4c01/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/SqlConnectionBase.java#L164

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,128 @@
...
+        .ifPresent(connection -> {
+          LOG.debug("Removing and closing oldest available connection: {} {}",
+              connection.getTenantId(), connection.getSessionId());
+          if (connection.getWrappedConnection() != null) {
+            activeConnectionCount--;
+            connection.getWrappedConnection().close();
+          }
+          connectionCache.remove(connection);
+        });
+  }
```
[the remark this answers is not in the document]
steveellis: Ok agree we need to handle timeout for idle connections since even though the connections will cycle through fairly quickly under load there could be connections that hang around during idle times.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,128 @@
...
+        LOG.debug("{} {} {} {} {} {} {} {} {}", cacheHits, cacheMisses, connectionCache.size(),
+            activeConnectionCount, poolSize, c.getSessionId(), c.getTenantId(), c.isAvailable(),
+            c.getConnectionHash()));
+    }
+  }
+
+  private Future<PgConnection> getOrCreateCachedConnection(Pool pool, String schemaName, String tenantId) {
+    Optional<CachedPgConnection> connectionOptional =
+        connectionCache.stream().filter(item ->
+            item.getTenantId().equals(tenantId) && item.isAvailable()).findFirst();
```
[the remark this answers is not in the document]
steveellis: Actually this makes me realize that I'm not updating the timestamp of the connection when they are used but rather only setting it once when the connection is created. Perhaps it should be updated when they are used to properly do the LRU pattern.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
steveellis: I think the new implementation should resolve both of these concerns.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
steveellis: Changing it to `Pool` was done because of a sonar warning

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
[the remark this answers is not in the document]
steveellis: I took a stab at implementing a timer tied to the the `Vertx` of `PostgresClient`. Since there can be multiple clients with different configs, on consideration was to let the lowest connection release delay be the one controlling eviction based on the timeout.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,239 @@
...
+    if (this.lowestReleaseDelayReceived == 0 || connectionReleaseDelay < lowestReleaseDelayReceived) {
+      // Since there can be multiple clients, we let whichever one has provided the lowest release delay win.
+      this.lowestReleaseDelayReceived = connectionReleaseDelay;
+    }
+
+    LOG.debug("Setting idle connection timeout observer with interval: {}", observerInterval);
+    vertx.setPeriodic(observerInterval, id -> {
+      LOG.debug("Observer firing:: connectionReleaseDelay: {}, observerInterval: {}, observer id {}",
+          connectionReleaseDelay, observerInterval, id);
+      logCache();
```
okolawole-ebsco, line 77: I think we should consider showing the metrics via the MetricsUtil of RMB and having them show up in JMX/Mbeans. This can be done in another iteration.

domain-models-runtime/src/test/java/org/folio/PostgresConnectionManagerTest.java
okolawole-ebsco, line 1: Any reason this class is not in a similar package as PostgresConnectionManager?

steveellis: This is just me being sloppy. The test belongs in `org.folio.rest.persist`.

domain-models-runtime/src/test/java/org/folio/rest/impl/PostgresConnectionManagerLoadTest.java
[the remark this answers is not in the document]
steveellis: Think it's because I based this class off `TenantAPIIT.java` which is in `org.folio.rest.impl`. It shares some code through a helper class I created called `TenantHelper` for doing the sharing.

domain-models-runtime/src/main/java/org/folio/rest/persist/CachedPgConnection.java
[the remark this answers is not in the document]
steveellis: I believe I have resolved this. Please take a look.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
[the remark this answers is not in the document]
steveellis: I think I am not yet configuring the idleTimeout of PgPool to be 0. Let me do that.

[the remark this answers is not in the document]
steveellis: I think I have decided to go ahead with the idea of making the timeout a feature of the overridden connection class. This will be a lot cleaner. Also I think it could be that the timeouts in the PTF tests are the result of idleTimeout not being set to 0 (fully overriding the internal timeout). If we fully override we fully control it.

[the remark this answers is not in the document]
steveellis: Please see the latest for a new implementation of the connection release delay (the [ReleaseDelayObserver](https://github.com/folio-org/raml-module-builder/pull/1142/files#diff-df1fdffb5f8379246f2a6659389f2a203b589a51cabbf2e757a12d1f3b056884)). This allows each connection to manage its own release delay timer which is reset each time a connection becomes available.

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresConnectionManager.java
```diff
@@ -0,0 +1,128 @@
...
+      logCache(ctx, pool.size());
+      connection.setUnavailable();
+      return Future.succeededFuture(connection);
+    }
+
+    var ctx = String.format("cache miss %s", tenantId);
+    cacheMisses++;
+    logCache(ctx, pool.size());
+
+    return createConnectionSession(pool, schemaName, tenantId);
```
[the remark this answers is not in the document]
steveellis: My preference is not to implement our own wait queue since the current code has been tested and is known to work, adding this additional feature will add time (another round of testing) and risk (potential unforeseen problems) to this implementation. Perhaps we can tackle this in a phase 2 if performance of the shared pool is thought to need further enhancements.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedConnectionManager.java
```diff
@@ -0,0 +1,111 @@
...
+    this.tryRemoveOldestAvailableConnectionAndClose();
+  }
+
+  public Future<PgConnection> getConnection(Vertx vertx, Pool pool, String schemaName, String tenantId) {
+    if (! PostgresClient.isSharedPool()) {
+      LOG.debug("Not in shared pool mode");
+      return pool.getConnection().map(PgConnection.class::cast);
+    }
+
+    if (this.connectionReleaseDelaySeconds == -1) {
```
okolawole-ebsco, line 49: is there a way to ensure this is set before calling `getConnection`?

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedConnectionManager.java
```diff
@@ -0,0 +1,111 @@
...
+
+    LOG.debug("In shared pool mode");
+    return getOrCreateCachedConnection(vertx, pool, schemaName, tenantId);
+  }
+
+  public void setConnectionReleaseDelay(int seconds) {
+    this.connectionReleaseDelaySeconds = seconds;
+  }
+
+  private static int getMaxSharedPoolSize() {
```
okolawole-ebsco, line 61: It looks like this method is only used in the constructor of this class and it is static because the object has not been created yet. Consider moving this into a utilities class or similar. There should be a function in RMB somewhere to derive this configuration.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedConnectionManager.java
```diff
@@ -0,0 +1,111 @@
...
+    return Envs.getEnv(Envs.DB_MAXSHAREDPOOLSIZE) == null ?
+        PostgresClient.DEFAULT_MAX_POOL_SIZE :
+        Integer.parseInt(Envs.getEnv(Envs.DB_MAXSHAREDPOOLSIZE));
+  }
+
+  private void tryRemoveOldestAvailableConnectionAndClose() {
+    // The cache must not grow larger than the max pool size of the underlying pool.
+    var cacheExhausted = this.connectionCache.size() >= maxPoolSize;
+    if (! cacheExhausted) {
```
okolawole-ebsco, line 70: nitpick: there is a space between `!` and `cacheExhausted`

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  public int size() {
+    synchronized (cache) {
+      return cache.size();
+    }
+  }
+
+  public void log(String event) {
```
okolawole-ebsco, line 83: log methods on a cache object does not seem intuitive. Can you add some comments here?

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    String toString(String msg) {
+      return msg + String.format(":: %s (hits) %s (misses) %s (size) %s (active) %s (pool)",
+          hits, misses, cache.size(), active, poolSize);
+    }
+
+    String toStringDebug() {
+      var items = "\nCONNECTION MANAGER CACHE ITEMS (DEBUG):\n" ;
+      synchronized (cache) {
```
okolawole-ebsco, line 144: nitpick: locking the cache to print debug information seems excessive. I would expect non-blocking reads at least for diagnostics. Maybe you had a concurrent modification exception at some point...

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ReleaseDelayObserver.java
```diff
@@ -0,0 +1,38 @@
+package org.folio.rest.persist.cache;
+
+import io.vertx.core.Vertx;
+
+public class ReleaseDelayObserver {
```
okolawole-ebsco, line 5: Please add comments describing the function of this class.

domain-models-runtime/src/main/resources/log4j2.properties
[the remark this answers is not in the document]
okolawole-ebsco: I think we should revert this now.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedConnectionManager.java
steveellis: It will be set before since it is set in `PostgresClient.createPgPool`[here](https://github.com/folio-org/raml-module-builder/pull/1142/files#diff-0de14d572773e0367110bc28030981f9105376ca37b0b92716f5389f819cf8c6R627) but I think you are right it is a code smell.

I have 2 ways of dealing with configuration/env values. The way shown here, and the way I get the `maxSharedPoolSize` in the constructor you rightly take issue with below. There shouldn't be 2 ways.

I think I can statically define both of these values in the class where they either read from env or the default if not present.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedConnectionManager.java
steveellis: Gonna statically define this.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ReleaseDelayObserver.java
steveellis: Adding comments to all of the classes that are part of this new package to document how they work together.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
steveellis: I think iteration needs to be synchronized too because if one thread tries to modify it while it is iterating it could lead to a `ConcurrentModificationException`. Anyway it shouldn't matter in prod since the debug logging will be off.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
```diff
@@ -0,0 +1,156 @@
...
+          });
+    }
+  }
+
+  public Optional<CachedPgConnection> getConnection(String tenantId) {
+    Optional<CachedPgConnection> connectionOptional;
+    synchronized (cache) {
+      connectionOptional =
+          cache.stream().filter(item ->
+              item.getTenantId().equals(tenantId) && item.isAvailable()).findFirst();
```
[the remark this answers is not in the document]
steveellis: Agree this is a good idea. I think this can be handled by adding a call to `tryRemoveOldestAvailableAndClose` (if cache.size is == pool) size before this code. Or would you prefer a new method that was something like `getOldestAvailable` which returned the connection, which we then call `SET` on?

domain-models-runtime/src/main/java/org/folio/rest/persist/PostgresClient.java
```diff
@@ -74,6 +74,9 @@ public class PostgresClient {
 
   public static final String     DEFAULT_SCHEMA           = "public";
   public static final String     DEFAULT_JSONB_FIELD_NAME = "jsonb";
+  public static final int        DEFAULT_MAX_POOL_SIZE = 10;
```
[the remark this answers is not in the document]
steveellis: It was easier to debug and see it working with 10. I can change it back to 4.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedPgConnection.java
```diff
@@ -0,0 +1,225 @@
...
+                            Vertx vertx,
+                            int releaseDelaySeconds) {
+    if (tenantId == null || tenantId.isEmpty() || connection == null || manager == null) {
+      throw new IllegalArgumentException();
+    }
+
+    this.tenantId = tenantId;
+    this.connection = connection;
+    this.manager = manager;
+    this.sessionId = UUID.randomUUID();
```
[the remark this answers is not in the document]
steveellis: I tried this but noticed that the id wasn't unique for each session which I think is what we want here.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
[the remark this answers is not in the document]
steveellis: See what you think of the implementation. Instead of letting callers check the tenant id, I created a new property called `isRecycled` which signals to callers that it is a connection that belonged to another tenant and is being "recycled" for a new one. The tenant id is changed in `conn.setRecycled(tenantId)`.  Then the manager takes care of setting a new role and schema for the connection when `isRecycled` is true.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ReleaseDelayObserver.java
```diff
@@ -0,0 +1,60 @@
...
+ * connection. See {@link CachedPgConnection#setUnavailable()} and {@link CachedPgConnection#setAvailable()} for how
+ * connection availability is managed. When a connection becomes available the timer starts.
+ * @see ConnectionCache
+ * @see CachedConnectionManager
+ * @see CachedPgConnection
+ */
+public class ReleaseDelayObserver {
+  private final int releaseDelaySeconds;
+  private final Vertx vertx;
+  private Long timerId;
```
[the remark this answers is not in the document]
steveellis: If I switch to the value type then I loose the convenience of the null check.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/CachedConnectionManager.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+    try {
+      return Integer.parseInt(envValue);
+    } catch (NumberFormatException e) {
+      LOG.error("Environment variable has wrong format:: variable: {}, value: {}", envKey.name(), envValue);
+      return defaultValue;
+    }
+  }
+
+  private void tryRemoveOldestAvailableConnectionAndClose() {
```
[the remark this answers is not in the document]
steveellis: Changed my mind. Agree your name is better.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ReleaseDelayObserver.java
```diff
@@ -0,0 +1,61 @@
...
+    if (releaseDelaySeconds == 0) {
+      return;
+    }
+
+    if (timerId != null) {
+      vertx.cancelTimer(this.timerId);
+    }
+
+    vertx.setTimer(toMilliseconds(releaseDelaySeconds), id -> {
+      timerId = id;
```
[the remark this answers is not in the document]
steveellis: Good catch

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
```diff
@@ -0,0 +1,189 @@
...
+      return;
+    }
+
+    var msg = metrics.toString(LOGGER_LABEL + ": " + context);
+    if (level == Level.DEBUG) {
+      var debugMsg = msg + metrics.toStringDebug();
+      LOG.debug(debugMsg);
+    }
+    LOG.info(msg);
+  }
```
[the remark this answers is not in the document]
okolawole-ebsco: Checking log levels can be curbed by using the log methods that accept lambdas 

`void info(String message, Supplier<?>... paramSuppliers);`

This means the lambda is only evaluated when the log needs to be written for that particular log level.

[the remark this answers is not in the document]
steveellis: Incorporated both of your ideas. Think it is more readable now and works better no matter what the logging level is.

domain-models-runtime/src/main/java/org/folio/rest/persist/cache/ConnectionCache.java
```diff
@@ -0,0 +1,190 @@
...
+   * is hit or missed.
+   * @param context Any details that help contextualize the event.
+   */
+  public void log(String context) {
+    Supplier<String> msgSupplier = () -> metrics.toString(LOGGER_LABEL + ": " + context);
+    Supplier<String> msgDebugSupplier = metrics::toStringDebug;
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("{} {}", msgSupplier.get(), msgDebugSupplier.get());
+    }
```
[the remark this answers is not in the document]
steveellis: Please take a look at the latest.

pom.xml
```diff
@@ -34,6 +34,8 @@
     <maven.version>3.9.6</maven.version>
     <vertx.version>4.5.5</vertx.version>
     <micrometer.version>1.12.3</micrometer.version> <!-- https://github.com/vert-x3/vertx-micrometer-metrics/blob/master/pom.xml#L41 -->
+    <junit-jupiter-version>5.10.0</junit-jupiter-version>
```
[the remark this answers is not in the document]
steveellis: Yes it can. Not sure how this crept in.

domain-models-runtime/src/test/java/org/folio/rest/persist/cache/CachedConnectionManagerTest.java
```diff
@@ -0,0 +1,89 @@
...
+    connection.close();
+    assertTrue(handled.get());
+  }
+
+  @Test
+  public void closeWithHandlerAsyncResultTest() {
+    var manager = new CachedConnectionManager();
+    var connection = new CachedPgConnection("tenant1", new PgConnectionMock(), manager, Vertx.vertx(), 1);
+    assertFalse(connection.isAvailable());
+    connection.close(event -> {
```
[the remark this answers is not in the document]
steveellis: I do know this.  Good catch.

pom.xml
```diff
@@ -34,6 +34,8 @@
     <maven.version>3.9.6</maven.version>
     <vertx.version>4.5.5</vertx.version>
     <micrometer.version>1.12.3</micrometer.version> <!-- https://github.com/vert-x3/vertx-micrometer-metrics/blob/master/pom.xml#L41 -->
+<!--    <junit-jupiter-version>5.10.0</junit-jupiter-version>-->
```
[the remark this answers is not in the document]
steveellis: Missed this, thanks. It's removed now.

okolawole-ebsco (approved):

PR #1151, opened 2024-03-30
RMB-943: Some dates are not deserialized properly (year 0)

julianladisch:
https://folio-org.atlassian.net/browse/RMB-943

Jackson Databind serialization of a Date prepends a plua (+) in front of year 0. This fails in Jackson's Date deserialization.

Fix: Add a custom Date serializer and custom Date deserializer that remove the plus.

### DIFF
--- a/dbschema/src/main/java/org/folio/dbschema/ObjectMapperTool.java
+++ b/dbschema/src/main/java/org/folio/dbschema/ObjectMapperTool.java
@@ -1,6 +1,5 @@
 package org.folio.dbschema;
 
-import com.fasterxml.jackson.core.JacksonException;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;


dbschema/src/test/java/org/folio/dbschema/ObjectMapperToolTest.java
```diff
@@ -62,4 +73,43 @@ void canReadSchema() throws Throwable {
       is(true));
     assertThat(dbSchema.getExactCount(), is(10000));
   }
+
+  @ParameterizedTest
+  @CsvSource({
+    " 2024-12-31T13:14:15.789+00:00, 2024-12-31T13:14:15.789+00:00",
+    " 0000-01-01T00:00:00.000+00:00, 0000-01-01T00:00:00.000+00:00",
+    "+0000-01-01T00:00:00.000+00:00, 0000-01-01T00:00:00.000+00:00",
+    "+0000-12-31T23:59:59.999+00:00, 0000-12-31T23:59:59.999+00:00",
```
steveellis, line 82: Will all dates either have plus in front of them or a space? Or could some have no space and no plus?

steveellis: > CsvSource trims whitespace at the beginning and the end of the value: https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-CsvSource
> 
> Do you want me to reformat using textBlock or single quotes?

No it's fine. I can see from the link that it trims by default. There's an option if you don't want to trim but I don't see a reason for that here.

steveellis (approved):
Agree that it looks like `+0000-01-01T00:00:00.000+00:00` isn't a valid ISO 8601 date from what I can tell and that its a bug in jackson. Think your comment in the code is enough to explain what we're doing here.

PR #1152, opened 2024-04-04
RMB-992: Increment optimistic locking _version for jsonb only

julianladisch:
Some code like
https://folio-org.atlassian.net/browse/MODINVSTOR-1105 = https://github.com/folio-org/mod-inventory-storage/pull/926/files 
maintains an additional derived column like complete_updated_date.

Only updates of the jsonb column should trigger an _version increment, not updates of other columns like complete_updated_date.

Other columns are derived values and therefore don’t need optimistic locking.

If developers manually add a column that requires optimistic locking they can manually adjust the optimistic locking trigger accordingly. Columns not maintained by RMB are out of scope of RMB’s optimistic locking support.

### DIFF
--- a/dbschema/src/main/java/org/folio/dbschema/ObjectMapperTool.java
+++ b/dbschema/src/main/java/org/folio/dbschema/ObjectMapperTool.java
@@ -1,26 +1,27 @@
 package org.folio.dbschema;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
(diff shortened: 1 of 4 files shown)


domain-models-runtime/src/main/resources/templates/db_scripts/optimistic_locking.ftl
```diff
@@ -39,7 +39,7 @@
   DROP TRIGGER IF EXISTS set_${table.tableName}_ol_version_trigger
     ON ${myuniversity}_${mymodule}.${table.tableName} CASCADE;
     
-  CREATE TRIGGER set_${table.tableName}_ol_version_trigger BEFORE INSERT OR UPDATE
+  CREATE TRIGGER set_${table.tableName}_ol_version_trigger BEFORE INSERT OR UPDATE OF jsonb
```
steveellis, line 42: This seems like a pretty significant change if this disables OL for any field that isn't a jsonb field. What am I not understanding? What does this mean for modules that are relying on OL for non-jsonb fields?

steveellis: Yes thanks. Just in case others come back to this PR and had similar questions: the existing API only contains derived columns (columns that are derived from other columns) and jsonb. Reducing the scope of _version to jsonb doesn’t affect existing code since these columns didn’t need OL in the first place, and it helps with changes that need to be made to mod-inventory.

steveellis (approved):

TarasSpashchenko (approved):

PR #1158, opened 2024-04-23
RMB-994: Cache loading of schema.json in CQL2PgJSON

julianladisch:
https://folio-org.atlassian.net/browse/RMB-994

RMB should cache the loading and deserialization of schema.json as some modules and RMB’s PgUtil reload it on every request.


cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -65,6 +66,8 @@ public class CQL2PgJSON {
   private static Logger logger = LogManager.getLogger(CQL2PgJSON.class);
 
   private static final String JSONB_COLUMN_NAME = "jsonb";
+  /** map schemaPath to Schema */
+  private static Map<String, Schema> dbSchemaCache = new HashMap<>();
```
okolawole-ebsco, line 70: We should consider using a `ConcurrentHashMap` here. At the minimum, multiple eventloops may try to modify contents of the schema. Also this class does not communicate that it should be used in a vertx only context. 
Most of the modifications will happen at the beginning of the app so I don't think it will be a big issue.

cql2pgjson/src/main/java/org/folio/cql2pgjson/CQL2PgJSON.java
```diff
@@ -168,15 +171,19 @@ public void setDbSchemaPath(String dbSchemaPath) {
...
-        schemaPath = "templates/db_scripts/schema.json";
-      }
       String dbJson = ResourceUtil.asString(schemaPath, CQL2PgJSON.class);
       logger.info("loadDbSchema: Loaded {} OK", schemaPath);
-      dbSchema = ObjectMapperTool.getMapper().readValue(dbJson, Schema.class);
-    } catch (IOException ex) {
-      logger.error("No schema.json found", ex);
+      return ObjectMapperTool.getMapper().readValue(dbJson, Schema.class);
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
```
okolawole-ebsco, line 186: Since there is no logging here, i think the exception will get swallowed in the `dbSchemaCache.computeIfAbsent` call above since the call is effectively in a lambda. Do you agree?

okolawole-ebsco: Yep, I saw that earlier. Looks like setDbSchemaPath calls the method. All is well!

okolawole-ebsco (approved):

PR #1159, opened 2024-04-29
RMB-997: Avoid left(...,600) for sqlExpression/sqlExpressionQuery

julianladisch:
https://folio-org.atlassian.net/browse/RMB-997


damien-git (approved):
This looks good.
I tried to see if it would have any impact on existing code. I found one usage of `sqlExpression` in mod-circulation-storage, but the [schema](https://github.com/folio-org/mod-circulation-storage/blob/8e65a7ad75bf2be6288747320d7a48a5a951df65/src/main/resources/templates/db_scripts/schema.json#L120C25-L120C51) does not seem to match the [index creation](https://github.com/folio-org/mod-circulation-storage/blob/8e65a7ad75bf2be6288747320d7a48a5a951df65/src/main/resources/templates/db_scripts/index_dateLostItemShouldBeBilled.sql#L11) or [usage](https://github.com/folio-org/mod-circulation/blob/6e24f8567d26984fb0ca372e17aa505b86d590d7/src/main/java/org/folio/circulation/services/agedtolost/ChargeLostFeesWhenAgedToLostService.java#L271), so I guess the change would not affect the result.

SerhiiNosko (approved):

PR #1163, opened 2024-06-13
RMB-1000: Provide ObjectMapperTool.valueAsString

julianladisch:
Provide ObjectMapperTool.valueAsString to parallel existing ObjectMapperTool.readValue to support hand-written SQL.

Add warnings to README.md and upgrading.md notes about incompatibility of JsonObject.mapFrom and the new valueAsString method.


steveellis (approved):
Seems like a nice improvement

PR #1165, opened 2024-08-24
RMB-1005: Unset DB_ env fixing Failed to resolve 'myhost'

julianladisch:
https://folio-org.atlassian.net/browse/RMB-1005


steveellis (approved):
Thanks for the fix Julian.

PR #1171, opened 2024-10-11
RMB-360: db init without superuser/create rule/create schema privileges

julianladisch:
https://folio-org.atlassian.net/browse/RMB-360

Some sysops want to give the module's database user no superuser, no "CREATE ROLE" and no "CREATE SCHEMA" privileges and create the role and the schema beforehand, before the module gets enabled for the tenant.

Change RMB to gracefully handle this case.

This code has be successfully tested with mod-permissions.

### DIFF
--- a/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
+++ b/domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
@@ -86,7 +86,7 @@ public class TenantAPI implements Tenant {
     return requirePostgres(context, Integer.parseInt(minNum), min);
   }
 
-  Future<Boolean> tenantExists(Context context, String tenantId){
+  Future<Boolean> tenantExists(Context context, String tenantId) {
     /* Connect as super user so that all commands will be available.
      * Return true if at least one table exists (rmb_internal).
      * Reason: Some installations create role (with restricted permissions) and


domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -83,12 +84,17 @@ Future<Void> requirePostgresVersion(Context context) {
   }
 
   Future<Boolean> tenantExists(Context context, String tenantId){
```
steveellis, line 86: Need space between `)` and `{`

domain-models-runtime/src/main/java/org/folio/rest/impl/TenantAPI.java
```diff
@@ -83,12 +84,17 @@ Future<Void> requirePostgresVersion(Context context) {
...
+     * Return true if at least one table exists (rmb_internal).
+     * Reason: Some installations create role (with restricted permissions) and
+     * empty schema before enabling a tenant therefore we cannot simply check
+     * whether schema exists.
+     */
+    var sql = "SELECT 1 FROM information_schema.tables WHERE table_schema=$1";
+    return postgresClient(context)
+        .selectSingle(sql, Tuple.of(PostgresClient.convertToPsqlStandard(tenantId)))
+        .map(Objects::nonNull)
+        .onFailure(e -> log.error("{}: {}", sql, e.getMessage(), e));
```
steveellis, line 97: Can a string be appended to the beginning of the error message (i.e before `{}: {}`) so that errors caused by this can be easily traced back to this point in the code? For example `Select failure in tenantExists: {}: {}`.

steveellis: I suppose the stacktrace will include the method name provided.

steveellis (approved):
Just a small formatting nitpick. I don't need to re-review.

PR #1180, opened 2024-12-17
Fix chown syntax; execInContainer error reporting

julianladisch:
Fix chown syntax, the old syntax is no longer available in Alpine 3.21. postgres:16.6-alpine and postgres:12.22-alpine no longer come with Alpine 3.20 but with Alpine 3.21.

When execInContainer fails throw an exception with complete stderr and stdout.


barbaraloehle (approved):
The syntax fix of chown works in the method execInContainer. It could be help to test the error reporting with a further test like
@Test
  static void testExecRuntimeException() {
    assertThrows(RuntimeException.class, () -> {
      exec("chown", "postgres..postgres", KEY_PATH, CRT_PATH);});
  }

steveellis (approved):

barbaraloehle (approved):
Thank you. The test method exeTest cReportsFailure() is very evident.

PR #1241, opened 2026-08-17
RMB-1058 Fix generated resource interfaces using wrong import package when schema defines javaType

psmagin:
https://folio-org.atlassian.net/browse/RMB-1058

## Purpose
Fix generated resource interfaces using wrong import package when schema defines javaType

## Approach
- Added GenerateRunner.fixModelPackageImports(), called after code generation. It scans all generated POJO files to build a simpleClassName → actualPackage map, then rewrites any stale import org.folio.rest.jaxrs.model.X; in the generated resource interfaces to use the correct package. The approach mirrors the existing migrateToJakarta() post-processing pattern. 
- Added test fixture (widget.json schema with "javaType": "org.folio.acme.Widget" + widgets.raml) and a regression test fixModelPackageImports_usesJavaTypePackageInResourceInterface that verifies the generated Widgets resource interface imports org.folio.acme.Widget and does not import org.folio.rest.jaxrs.model.Widget.

### DIFF
--- a/domain-models-maven-plugin/src/main/java/org/folio/rest/tools/GenerateRunner.java
+++ b/domain-models-maven-plugin/src/main/java/org/folio/rest/tools/GenerateRunner.java
@@ -240,7 +240,7 @@ public class GenerateRunner {
       var fixed = DEFAULT_IMPORT_PATTERN.matcher(content).replaceAll(mr -> {
         var simpleName = mr.group(1);
         var actualPkg = classPackageMap.get(simpleName);
-        return (actualPkg != null && !actualPkg.equals(MODEL_PACKAGE_DEFAULT))
+        return (actualPkg != null)
             ? "import " + actualPkg + "." + simpleName + ";"
             : mr.group(0);
       });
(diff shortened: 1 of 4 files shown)


domain-models-maven-plugin/src/main/java/org/folio/rest/tools/GenerateRunner.java
```diff
@@ -185,9 +189,69 @@ public void generate(String inputDirectory) throws IOException {
...
+    return map;
+  }
+
+  private void fixImports(Path file, Map<String, String> classPackageMap) {
+    try {
+      var content = Files.readString(file, StandardCharsets.UTF_8);
+      var fixed = DEFAULT_IMPORT_PATTERN.matcher(content).replaceAll(mr -> {
+        var simpleName = mr.group(1);
+        var actualPkg = classPackageMap.get(simpleName);
+        return (actualPkg != null && !actualPkg.equals(MODEL_PACKAGE_DEFAULT))
```
julianladisch:
```suggestion
        return (actualPkg != null)
```

julianladisch (approved):
Looks good, thanks!

